PR #1, opened 2016-11-04
Adds a few example pack files and component files.

mrdewitt:
Also defines a ".attributes" file for those files that should be
included with either specific headers or from subpackages.


examples/example1/example1.pack
```diff
@@ -0,0 +1,29 @@
...
+Content-Location: /otherPage.html
+Content-Type: text/html
+
+<body>
+  Hello World! <img src="images/world.png">
+</body>
+--j38n02qryf9n0eqny8cq0
+Content-Location: /images/bank.png
+Content-Type: image/png
+Content-Transfer-Encoding: base64
```
dimich-g: Would be nice to avoid base64 from the start... would guthub barf if it was just a binary image?

PR #19, opened 2016-12-05
Adds an initial version of a packager script.

mrdewitt:
This script can be updated alongside the explainer to give a concrete
answer to many format-related questions, and to uncover and resolve
ambiguities in the proposed spec.

Many parts of the spec are unimplemented as yet.


dimich-g:
Great! This looks like a good start, looking forward for the rest of it.

PR #22, opened 2016-12-28
No need for algorithm header, fixes #14

### LINKED ISSUE #14: No need to specify cert algo as that is defined by the cert itself
As asked [here](https://github.com/dimich-g/webpackage#use-case-signed-package-one-origin):
> Should there be '-with-ECDSA' since certificate specifies the key's algorithm?

No, not necessary since the certificate specifies the algorithm.

bmeck:
fixes #14


README.md
```diff
@@ -185,13 +185,14 @@ The example contains an HTML page and an image. The package is signed by the exa
...
-2. The Content Index contains hashes of all parts of the package, so it is enough to validate the index to trust its hashes, then compute the hash of the each part upon using it to validate each part. Hashes have hash algorithm specified in front.
-3. The inclusion of certificate makes it possible to validate the package offline (certificate revocation aside, this can be done out-of-band when device is actually online).
-4. Certificate is included as one of standard the DER-encoded resource (with proper Content-type).
+2. The **algorithm** must be encoded within the certificate that signed the package.
+3. The Content Index contains hashes of all parts of the package, so it is enough to validate the index to trust its hashes, then compute the hash of the each part upon using it to validate each part. Hashes have hash algorithm specified in front.
+4. Content Index Entry `part-location` and `part-size` must not refer to locations outside of the package which contains the entry or locations within nested packages. They may refer to the boundaries of a nested package.
```
mrdewitt: For clarity, could you change "locations within nested packages" to "to locations within nested packages"

dimich-g: @mrdewitt I'm separately merging your suggestion since it was a part of another pr, thanks!
@bmeck There should at least be an algorithm used for the hash, right? While the encryption algorithm is specified in the certificate, the specific form of hash used here can vary (not a part of cert) so needs to be mentioned.

bmeck: @dimich-g ah yes, that is a good point the I was just looking at how Content Index Entry has its own algorithm and missed that we still need the hash-algorithm in the Package-Signature. I put the hash algorithm back in Package-Signature, but we can still drop the cert one.

bmeck: if we only have X.509 I think it will always contain a Signature Algorithm, but I might want to investigate other cert types to see if they also always carry that.

lrosenthol:
Yes, X.509 certs always have the algorithm.  What are types of certs are
you thinking - (Open)PGP?

bmeck:
@lrosenthol PGP in real world scenarios I could see being a candidate. I once heard of npm modules potentially being signed for publish using developer SSH keys, but I don't want that.

bmeck:
@lrosenthol PGP Version3 appears to mandate the hash algorithm as well, but only in the signature packet https://tools.ietf.org/html/rfc4880#section-5.2.2, not in the key packet https://tools.ietf.org/html/rfc4880#section-5.2.2 . To remove the algorithm in `Package-Signature` we would also need to mandate a signature packet which somewhat defeats the purpose here.

lrosenthol:
Or we decide that signing of web packages via PGP key isn't a goal.  And
while I love many things about PGP (having worked with Phil and the
original team back in the 90's) - it unfortunately never reached wide
adoption.  I think we would better served with an X.509-only based solution
that is simpler.

PR #23, opened 2016-12-29
[do not merge] MIME boundary => chunked encoding

### LINKED ISSUE #6: Streaming not supported
One of the reasons that the [packaging on the web](https://w3ctag.github.io/packaging-on-the-web/) spec chose to avoid ZIP in favor of something new & different was due to the (perceived) lack of streaming support.

However, this proposal suffers from the same problem.  You cannot create it entirely in stream, due to (a) the way that offsets are used and (b) the index file needing to list all other files.

If streaming is not a requirement for this format - that's fine.  But then, that should also be called out in the spec.

bmeck:
This allows for:

* streaming without worrying about collision and without the need to pre-parse the file
* signatures to be preserved even if chunk sizes differ due to streaming

fixes #6

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ We are proposing to improve on the spec, in particular by introducing 3 major ad
 
 We also propose to remove the [fragment-based URL schema](https://w3ctag.github.io/packaging-on-the-web/#fragment-identifiers) from the spec as it is not clear what would be the use case supporting it.
 
-We also propose to change the encoding of content within the package format to use [chunked encoding](https://tools.ietf.org/html/rfc7230#section-4.1). The encoding will be bound to the start of a content's headers and end at the completion of the content's body. Multiple chunks for a single content resource may be used when sending data. The data for chunked encoding chunk lengths will be ignored during signing. Withing packages there must be no use of extension for chunked encoding.
+We also propose to change the encoding of content bodies within the package format to use [chunked encoding](https://tools.ietf.org/html/rfc7230#section-4.1). The data for chunked encoding chunk lengths will be ignored during signing. Within packages there must be no use of extension for chunked encoding.
 
 Following are some example usages that correspond to these additions:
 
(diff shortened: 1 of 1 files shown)


bmeck:
marked [do not merge]

need to discuss the outer boundaries more in depth so that package level header and trailer are clear.

bmeck:
I am thinking you mean "part" to be body of a resource within the package.

This does allow encoding resources into multiple chunks. Individual chunk lengths are not related to total size of a resource.

bmeck:
After a while of thinking I cannot easily preserve nesting packages while maintaining streamking via normal chunked encoding. A variant seems possible, but the problem comes from:

```
... outer webpackage ...
<- no delimiter for declaring entering nested package
<- would need add extra wrapping header + chunked encoding content
inner package <- all chunked (even headers :( )
<- extra `0\r\n` is a bit odd
... outer webpackage ...
```

I will investigate some other framing mechanism which can support streaming nested packages without this problem.

jyasskin:
Could you update this to use `\n` line endings? Right now git thinks every line is changed.

Do we need to be able to stream the nested packages? It'd be a little weird for the main server to be reflecting a stream served from another server, since it couldn't then take over and serve part of its own content if the nested server pauses. That ... makes me think we should actually be using something from HTTP2 directly for the streaming case and only use a packaging format for serializing the content.

bmeck:
@jyasskin for streaming onto a disk (from any source) from the web we don't have the guarantees of a HTTP/2 style multiplexing stream.

jyasskin:
I think that mixes two use cases.
1. Server sends content as it's generated, rather than waiting for a whole file to be ready at once. For this, you have HTTP2.
2. Serialize content to disk and share it to peers. At this point, you've downloaded the whole thing and can use an encoding that needs lengths up front. You might have to do a second pass over the data.

I could be missing something.

lrosenthol:
@jyasskin 

1) Content can be served via means other than HTTP, and HTTP2 is still very far from having even a small penetration.  We need something that works over any protocol.

2) But if you modify the stream/data - you break the signatures.  That's not good.

jyasskin:
I don't think the packaging format needs to solve problems that new transports have already solved. People who don't upgrade to the new transports just have to deal with the problems of the transports they're using.

I do think there's a problem with including offsets in the signed data, but I need to double-check with security folks that it's safe to avoid it.

lrosenthol:
I agree that we don't need to solve problems already solved - and I don't think we are doing that here.  Instead, we are designing a protocol and storage system neutral solution.

Actually the opposite - if the offsets aren't signed, you have a big hole in your signature model.

bmeck:
closing due to newer PR and the move to CBOR

PR #25, opened 2017-01-09
avoid colliding MIME for Content Index

bmeck:
fixes https://github.com/dimich-g/webpackage/issues/9 by using a sub tree of `application/package.index`


bmeck:
I am neutral on this topic overall given how UA generally treat MIME type as opaque strings.

dimich-g:
Thanks a lot, this is a good suggestion!

PR #28, opened 2017-01-24
Fix a bunch of typos and missing links.

jyasskin:
I'm not sure I've got the right fix for each of these; feel free to say anything should be different.

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,7 +120,7 @@ Content-Type" application/javascript
 The package in this case contains a lot of pages with resources ("Encyclopedia in a file") or multiple sites (in subpackages). The proposed structure facilitates efficient access, assuming the whole package is available locally. Several important notes:
 
 1. The Link: header with **rel=index** declares a specified part to be a Content Index of the package. The **offset=12014** attribute specifies the octet offset/size from the beginning of the Package Header of the package to the Content Index part. That can be used in file-seek APIs to quickly read the part without a need to parse the potentially huge package itself.
-2. Content Index part is typically generated during package creation, it doesn't have a natural URL. We propose to use cid: generated URLs (UUID-based, 128-bit) for such generated parts. The visibility scope of those URLs is limited similar to package boundaries, and is for the current package only.
+2. Content Index part is typically generated during package creation, it doesn't have a natural URL. We propose to use [`urn:uuid:`](https://tools.ietf.org/html/rfc4122) generated URLs (UUID-based, 128-bit) for such generated parts. The visibility scope of those URLs is limited similar to package boundaries, and is for the current package only.
 3. Content-type of the Content Index is application/package.index.
 4. The Content Index consists of the Content Index Entries (see below for the discussion of what they are).
 5. Content Index part may be compressed (as specified by Transfer-Encoding header).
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -121,8 +120,8 @@ Content-Type" application/javascript
...
 
 1. The Link: header with **rel=index** declares a specified part to be a Content Index of the package. The **offset=12014** attribute specifies the octet offset/size from the beginning of the Package Header of the package to the Content Index part. That can be used in file-seek APIs to quickly read the part without a need to parse the potentially huge package itself.
-2. Content Index part is typically generated during package creation, it doesn't have a natural URL. We propose to use cid: generated URLs (UUID-based, 128-bit) for such generated parts. The visibility scope of those URLs is limited similar to package boundaries, and is for the current package only.
-3. Content-type of the Content Index is application/index.
+2. Content Index part is typically generated during package creation, it doesn't have a natural URL. We propose to use [`cid:`](https://tools.ietf.org/html/rfc2392) generated URLs (UUID-based, 128-bit) for such generated parts. The visibility scope of those URLs is limited similar to package boundaries, and is for the current package only.
```
jyasskin: It seems from the FAQ that this link is wrong. I'd strongly suggest using a scheme that doesn't collide with anything that already has a definition.

README.md
```diff
@@ -121,8 +120,8 @@ Content-Type" application/javascript
...
+3. Content-type of the Content Index is [application/index](https://www.iana.org/assignments/media-types/application/index).
```
jyasskin: #25 handles this better.

README.md
```diff
@@ -186,10 +185,10 @@ Important notes:
...
 5. The inclusion of certificate makes it possible to validate the package offline (certificate revocation aside, this can be done out-of-band when device is actually online).
-6. Certificate is included as one of standard the DER-encoded resource (with proper Content-type) corresponding to a [X.509 certificate](https://tools.ietf.org/html/rfc5280). The policies for verifying the validity of the certificate is left up to the host environment. There are no mandates for any fields in the certificate such as [Key Usage](https://tools.ietf.org/html/rfc5280#section-4.2.1.3) required, this also is left up to the host environment.
+6. Certificate is included as a standard DER-encoded resource (with proper Content-type) corresponding to a [X.509 certificate](https://tools.ietf.org/html/rfc5280). The policies for verifying the validity of the certificate are left up to the host environment. There are no requirements on any fields in the certificate such as [Key Usage](https://tools.ietf.org/html/rfc5280#section-4.2.1.3); this also is left up to the host environment.
```
jyasskin: Please double-check this paragraph.

README.md
dimich-g: Right. The cid: schema is not really the right one here. Chrome uses URLs like cid:[uuid] in MHTML generation and this is apparently not standard, but then again MHTML is not well formed standard either... I've picked the cid: schema after Chrome's MHTML implementation, but perhaps urn:uuid: would work better, and it is standard:  http://www.ietf.org/rfc/rfc4122.txt
If you think it is better here, please feel free to change to it or let me know and I'll change after merging this PR.

README.md
dimich-g: Indeed. I've merged #25.

README.md
jyasskin: I'm happy to use `urn:uuid:` for UUIDs. I've made that a separate commit since it touches several parts of the document.

dimich-g:
Thanks!

Issue #16, opened 2016-11-17
What type of certificates can be used to sign? 


lrosenthol:
What is the expectation of what types of certificates can be used to sign these packages? 

SSL/TLS?
Code signing?
Standard document/message signing?
Other?

bmeck:
@lrosenthol do we need to constrain the types of certs or just the certificate container format: x509 vs PGP vs ...?

lrosenthol:
@bmeck Yes, we would need both.  First the certificate format (eg. X509 vs. PGP) and then second the type of cert - because you can't necessarily intermix SSL/TLS certs with code or document signing certs.  (which is, of course, by design)

lrosenthol:
@bmeck

It's pretty difficult to get a code signing cert, due to the normal usage
cases for it.   And timestamping even more so - since that's really only
for use in a timestamp server.

But can't hurt to ask...

dimich-g:
Was id-kp-codeSigning extension introduced to be able to issue those separately from SSL (and even using separate roots?) to reduce associated risks? 

Perhaps the limits on key usage (especially key usage extensions) should be a detail of implementation? For example, a browser would use SSL/TLS keys since the content it will validate is web content and the purpose of validation is to afford it the same privileges that delivering it via SSL provides. It is also likely that the procedure to sign the web content will involve the same certs the web servers use already. 

The code distribution tools for JS runtimes may have different limits and require different type of keys (if this is practical).

lrosenthol:
Given that we have identified at least three different use cases for this
work (browsers loading scripts/etc, application packagers and
publications), each of which would logically want/require a different type
of certificate - perhaps we shouldn't put any restrictions here and leave
it up to the use case in question to create a "profile" for their own
needs??

bmeck:
@lrosenthol @dimich-g if we want to leave it up to env that seems fine.

PR #31, opened 2017-01-26
Add some discussion of the security issues.

jyasskin:
* Outdated certificates and packages.
* That there may not be a single certificate chain.

@estark37

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,8 +32,25 @@ If the original origin signs a package, the contents can get full access to
 browser state and network requests for that origin. This lets people share full
 PWAs peer-to-peer.
 
+Local sharing tends to have persistent constraints on the amount of data
+transferred: each byte costs money. This means the client may be stuck with an
+outdated version of a package for a significant amount of time, including that
+package's vulnerabilities. It may be feasible to periodically check for:
+1. OCSP notification that a package's certificate has been revoked, and
+2. A yet-to-be-designed notification that the package is critically vulnerable
+   and needs to be disabled until it can be updated.
+
 ### Physical Web
 [Beacons](https://google.github.io/physical-web/) and other physical web devices often want to 'broadcast' various content locally. Today, they broadcast a URL and make the user's device go to a web site. This delivers the trusted content to the user's browser (user can observe the address bar to verify) and allow web apps to talk back to their services. It can be useful to be able to broadcast a package containing several pages or even a simple web app, even without a need to immediately have a Web connection - for example, via Bluetooth. If combined with signature from the publisher, the loaded pages may be treated as if they were loaded via TLS connection with a valid certificate, in terms of the [origin-based security model](https://tools.ietf.org/html/rfc6454). For example, they can use [`fetch()`](https://fetch.spec.whatwg.org/#fetch-api) against its service or use "Add To Homescreen" for the convenience of the user.
+
+Physical web beacons may be located in a place that has no connectivity, and
+their owner may only visit to update them as often as their battery needs
+replacement: annually or less often. This leaves a large window during which a
+certificate could be compromised or a package could get out of date. We think
+there's no way to prevent a client from trusting its initial download of a
+package signed by a compromised certificate. When the client gets back to a
+public network, it should attempt to validate both the certificate and the
+package, likely by fetching the package's `Content-Location`.
 
 ### Content Distribution Networks and Web caches.
 The CDNs can provide service of hosting web content that should be delivered at scale. This includes both hosting subresources (JS libraries, images) as well as entire content ([Google AMP](https://developers.google.com/amp/cache/overview)) on network of servers, often provided as a service by 3rd party. Unfortunately, origin-based security model of the Web limits the ways a 3rd-party caches/servers can be used. Indeed, for example in case of hosting JS subresources, the original document must explicitly trust the CDN origin to serve the trusted script. The user agent must use protocol-based means to verify the subresource is coming from the trusted CDN. Another example is a CDN that caches the whole content. Because the origin of CDN is different from the origin of the site, the browser normally can't afford the origin treatment of the site to the loaded content. Look at how an article from USA Today is represented:
@@ -43,6 +60,10 @@ The CDNs can provide service of hosting web content that should be delivered at 
 Note the address bar indicating google.com. Also, since the content of USA Today is hosted in an iframe, it can't use all the functionality typically afforded to a top-level document:
 - Can't request permissions
 - Can't be added to homescreen
+
+Packages served to CDNs can staple an OCSP response and have a short expiration
+time, avoiding the above problems with outdated packages.
+
 
 ## Goals and non-goals
 
--- a/README.md
+++ b/README.md
@@ -575,7 +575,12 @@ The example contains an HTML page and an image. The package is signed by the exa
 
 Important notes:
 
-1. The very first header in Package Header section of the package is **Package-Signature**, a new header that contains a signed hash of the Package Header section (not including Package-Signature header) and Content Index. It also contains a reference (via urn:uuid: UUID-based URL) to the part that contains the public key certificate (or if needed, a chain of certificates to the root CA).
+1. The very first header in Package Header section of the package is
+   **Package-Signature**, a new header that contains a signed hash of the
+   Package Header section (not including Package-Signature header) and Content
+   Index. It also contains a reference (via urn:uuid: UUID-based URL) to the
+   part that contains the public key certificate (or if needed, a group of
+   certificates that chain to one or more root CAs).
 2. The **certificate algorithm** must be encoded within the certificate that signed the package. The **algorithm** in **Package-Signature** is the hash algorithm used to sign the **Content Index** and produce the Package-Signature.
 3. The Content Index contains hashes of all parts of the package, so it is enough to validate the index to trust its hashes, then compute the hash of each part upon using it to validate each part. Hashes have the hash algorithm specified in front.
 4. Content Index Entry `part-location` and `part-size` must not refer to locations outside of the package which contains the entry or to locations within nested packages. They may refer to the boundaries of a nested package.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -291,43 +316,50 @@ Content-Type: application/pkcs7-mime
...
+### Does the package supply a full chain of certificates to a known CA root?
 
-Not necessarily. To quote one of the Chrome security engineers, "because different devices have different sets of roots in their trust stores, it's not always the case that there is a single "correct" set of certificates to send that will work best for all clients. Instead, for compatibility, sites will sometimes send a set of certificates and rely on clients to dynamically fetch additional intermediates when needed". We suspect that to be an interesting issue since offline validation would need a full chain. Browsers (at least Chrome) are implementing automatic fetching of intermediate certificates but for this to work they have to be online. This will probably become a matter of best practices when creating the packaging format, possibly with validation by the packaging tool.
+Not necessarily. Different devices have different sets of roots in their trust
+stores, so there is not a single "correct" set of certificates to send that will
+work best for all clients. Instead, for compatibility, packages may need to
+include a set of certificates from which chains can be built to multiple roots,
```
jyasskin: @sleevi, @estark37 said I should ask you if packagers are going to cause problems by including the certificates needed to chain to the roots of multiple different clients. She referenced https://groups.google.com/a/chromium.org/d/msg/net-dev/H-ysp5UM_rk/TqoyjV1eDAAJ as an example of servers choosing to make the client download some intermediates, which won't work as well for offline verification.

sleevi: @jyasskin I'll read up on the full proposal, but it sounds like you're suggesting using CA-rooted certificates, possibly even TLS certificates, for the packaging. That's not ideal for a number of reasons - this included - so I'll take a scan through and see what alternatives might be worth suggesting.

jyasskin: Thanks. The main constraint that I think leads us to use CA-rooted certificates is that we need the packages to be trusted as coming from particular web origins. I'm happy with anything else that proves an origin to the displaying browser.

sleevi: Right, unfortunately, this introduces subtle-but-significant cross-protocol attacks when you re-use a TLS certificate to sign this data, unfortunately, so this is going to be quite difficult to get what you want.

dimich-g: @sleevi is this true even if the publisher of the package obtains a separate TLS certificate for signing packages? Sort of similar to how code-signing certificates are separate form TLS ones.

lrosenthol:
This conversation has also taken place as part of #16. where its been pointed out that TLS-based certs won't work (as @sleevi says here).  Other types of certs - that aren't tied to domains, but instead of orgs or individuals - will need to be used.

But regardless, the set of valid and/or trusted roots needs to come from somewhere that may (or may not) be able to be mandated as part of the standardization.  And as part of that, you need to consider other standards in this area such as the [EUTL](https://ec.europa.eu/digital-single-market/en/eu-trusted-lists-trust-service-providers).

sleevi:
@lrosenthol That's not an accurate statement - that is, that other types of certs _will_ need to be used, or more generally, that certificates not bound to domains should be used. And the EUTL should absolutely be ignored in this space - not even bothered with.

@dimich-g I think the question was worded incorrectly; that is, you don't want to get a "second" TLS certificate that you use for signing, because as far as design goes, you're setting people up to fail if they have two certificates, interchangable, but one should be used for TLS and one not. To reduce the risk of cross-protocol attacks, you want to ensure that Certificate A (which can/should/is bound to a domain) cannot be used in Protocol B, and that Certificate B (which can/should/is bound to a domain) should not be used in Protocol A.

I don't anticipate being able to provide a more thorough review of the cryptography in the next few days, but there are definitely red flags with what's currently specified - ranging from the discussion of Key Usage to these issues of cross-protocol confusion. The signature design also doesn't account for the ability to perform counter-signatures, which is generally seen as a necessary scheme in document/code signing solutions (in Authenticode, it's used for timestampping; in Chrome extensions, it's used to vet that it's a Google WebStore delivered extension; in Android APKs, it's used to comply with local laws).

lrosenthol:
@sleevi 
>And the EUTL should absolutely be ignored in this space - not even bothered with.
Then you and I are going to have to agree to disagree.  The EUTL is a requirement for adoption of signatures within the URL for various business processes.  In the context of the use of this technology for publications, it would be necessary to be allow compatibility with it.  I am **not** suggesting mandating, just compatibility.

>doesn't account for the ability to perform counter-signatures
I don't believe that the current design prevents counter sigs, but I agree that it's not called out - and you are right, it should be as it is an important technology.  It's also going to be important for publications where the author may sign the original publication with the publisher adding theirs at the final release stage.

jyasskin:
@sleevi At the moment, we don't have a specification, just a sketch. Would you prefer that we non-cryptographers come up with a draft specification that you can review, or would you rather we write down a list of requirements and wait for a real cryptographer to design the right system?

sleevi:
@jyasskin Either way, knowing your goals clearly will help make sure whatever solution meets those goals - whether it's starting from a point in the draft where you're trying to accomplish those goals, or whether it's starting from scratch to build something to achieve those goals. Knowing, for example, whether you're exclusively targeting UAs or whether, as it seems suggested both here and the draft, you're imagining things like JS packaging or document delivery, is itself a useful aspect.

lrosenthol:
@sleevi - there are at least three different groups that are looking to
potentially leverage this work
- Browsers (which I assume you meant when you said UA, but is just one
class of UA for the OWP)
- Application environments (eg. Node+Electron)
- Documents & Publications (eg. Web Publications)

This to me means a common model and file format.  However, I see no reason
why the types of certificates allowed/used couldn't be different in each
case.

bmeck:
To clarify, I would expect Node and Electron to use Code Signing based x.509 certs rather than regular TLS certs. In particular, the origin of applications might not be tied to a specific WWW domain but an app store submission (like Android APK).

README.md
```diff
@@ -8,9 +8,26 @@ Some new use cases for Web technology have motivated thinking about a multi-reso
 
 Local sharing is quite popular, especially in Emerging Markets countries, due to cost and limitations on cellular data and relatively spotty WiFi availability. It is typically done over local Bluetooth/WiFi, by either built-in OS features like [Android Beam](https://en.wikipedia.org/wiki/Android_Beam) or with popular 3-rd party apps, such as [ShareIt](https://play.google.com/store/apps/details?id=com.lenovo.anyshare.gps) or [Xender](https://play.google.com/store/apps/details?id=cn.xender)). Typically, the locally stored media files and apps (APK files for Android for example) are shared this way. Extending sharing to bundles of content and web apps (Progressive Web Apps in particular) opens up new possibilities, especially if combined with a form of signing the content. Cryptographic signing could make it possible to afford the shared content the treatment normally reserved for the origin that content is claiming to be in - by verifying that the content indeed was produced by the holder of the corresponding certificate.
 
+Local sharing tends to have persistent constraints on the amount of data
```
mrdewitt: I'm confused - local sharing doesn't typically have per-byte costs.  Storage and downloading do.

README.md
```diff
@@ -8,9 +8,26 @@ Some new use cases for Web technology have motivated thinking about a multi-reso
...
+transferred: each byte costs money. This means the client may be stuck with an
+outdated version of a package for a significant amount of time, including that
+package's vulnerabilities. It may be feasible to periodically check for:
```
mrdewitt: Seems to me that rewording this without the numbered list would be clearer.  "It may be feasible to periodically check for OCSP notification that a package's certificate has been revoked.  Other channels for notifying about vulnerable packages have not yet been discussed/designed."

README.md
```diff
@@ -8,9 +8,26 @@ Some new use cases for Web technology have motivated thinking about a multi-reso
...
 
+Physical web beacons may be located in a place that has no connectivity, and
+their owner may only visit to update them as often as their battery needs
+replacement: annually or less often. This leaves a large window during which a
+certificate could be compromised or a package could get out of date. We think
+there's no way to prevent a client from trusting its initial download of a
+package signed by a compromised certificate. When the client gets back to a
+public network, it should attempt to validate both the certificate and the
+package, likely by fetching the package's `Content-Location`.
```
mrdewitt: This makes me think that analogue to the HEAD request where the client could fetch only the manifest/index and revalidate stale portions would be useful.

README.md
```diff
@@ -183,7 +203,12 @@ The example contains an HTML page and an image. The package is signed by the exa
 
 Important notes:
 
-1. The very first header in Package Header section of the package is **Package-Signature**, a new header that contains a signed hash of the Package Header section (not including Package-Signature header) and Content Index. It also contains a reference (via urn:uuid: UUID-based URL) to the part that contains the public key certificate (or if needed, a chain of certificates to the root CA).
+1. The very first header in Package Header section of the package is
```
mrdewitt: s/is/must be/

README.md
jyasskin: Indeed, that was worded badly. I meant that the use case happens where global connectivity is expensive, and have now updated the PR wording.

README.md
jyasskin: Done.

README.md
jyasskin: Yeah. We should define an optimized update mechanism, although validation might be simpler.

README.md
jyasskin: I'm going to undo the change to this section until we can update all the use cases.

README.md
jyasskin: #44 takes agl's advice to use the same signing format as TLS (and I used a different context string to distinguish package signatures from handshake signatures). I'd like to postpone the question of exactly what origin-trusted certificates will convince browsers to load a package until a later PR.

The current format even looks like it'll represent packages that have only signatures that aren't origin-trusted. Although web browsers would refuse to load those packages, they're useful for Node and Publishers.

dimich-g:
Points by @mrdewitt addressed, merged.

Issue #33, opened 2017-01-26
Document Node's use cases


jyasskin:
@bmeck, could you describe the use cases you're interested in for the packaging format, either in this issue or as a PR against the [`README.md`](https://github.com/dimich-g/webpackage/blob/master/README.md)?

bmeck:
From an older call: https://docs.google.com/document/d/1iPHPLMUVimh8cZgB_vkO-XtmDrnanOMI9v5kfnwiv4U/view

bmeck:
Mandatory:

* Signing
* Random Access
* Self Extracting capable

Important but not absolute requirements (would need some path to mitigate):

* Inter-package linking
* Nested packages
* Deduplication

Nice to have:

* Streaming
* Expiration

bmeck:
mandatory requirements have been reached, feel free to mark closed

Issue #29, opened 2017-01-26
Subpackage should have index in main package


jianli-chromium:
Without index, we will have to scan the whole package to find it out.

jyasskin:
I think #36 fixes this: https://github.com/dimich-g/webpackage#index

bmeck:
@jianli-chromium can this be closed?

Issue #12, opened 2016-11-17
Details of exactly what is hashed (both for individual entries and the package itself) are missing


lrosenthol:
There is nothing specified for exactly which bytes are hashed - either for an individual file/resource for the header or the index.  

Do you include the boundary string?  Do you include any closing line endings?

bmeck:
see note about each resource having its HEADER+BODY signed in : https://github.com/dimich-g/webpackage/blame/master/README.md#L308

My PR rephrased in a few places to make more explicit in https://github.com/dimich-g/webpackage/pull/23

It might be prudent however to change the phrasing of https://github.com/dimich-g/webpackage/blame/master/README.md#L312 to state that nested packages as signed as opaque entities as it sounds currently like nested packages are not something that the root package can verify as being correct currently.

bmeck:
Just noticed, the certificates are not signed in the Content-Index examples; unsure if this is intentional, but seems like just an accident.

bmeck:
@lrosenthol if certificates are signed is the rest covered?

lrosenthol:
You can't sign a certificate with itself - but you could sign the
certificates from a nested package with the certs from the parent - that
would be perfectly fine.

jyasskin:
Arguably, these details can go in the draft specification rather than the explainer, but it's probably time to start that specification and add these details, so we can iterate on them.

jyasskin:
#36 now specifies exactly which bytes are hashed and signed, so I believe this is fixed.

Issue #11, opened 2016-11-17
Index offset may not survive some processes


lrosenthol:
However, because the header of the is document is text - most sniffers will consider this a text file and treat it accordingly (eg. screwing up line endings).  If this happens, then all byte offsets will be incorrect and the file will be unable to load.

Consider finding a way to fool the sniffer or move to a binary format.

bmeck:
These offsets can also suffer if the package:

1. is re-ordered (perhaps due to priorities)

    Signatures are Headers+Body of a content resource currently which means it can be re-ordered fairly simply and maintain the signatures still. This was talked about on a phone call as being possible, particularly since the boundaries allow the Content Index Entities to keep the same hash.

2. streamed in such a way as chunked encoded where encoding boundaries/headers could change the exact offsets.

   In order for a Content Index to have these offsets for a streaming resource it would need to be at the end of file and not allow removing the encoding boundaries/headers. This seems a bit wasteful since a stream would need to track these offsets if the size of streamed chunks is not known ahead of time.

In any case, both of these have a need to recalculate the Package-Signature if the offsets change. Doing so while streaming would mean that Package-Signature also would need to occur at the end of file, presumably as a trailer.

jyasskin:
#36 takes the index structure out of the signed content, so it can be re-generated by any process that invalidates its offsets. I think this is fixed.

PR #36, opened 2017-02-07
Sketch an update to the package format and its signatures.

jyasskin:
I haven't updated the Use Cases yet, but here are some thoughts on how we should update the format to deal with #33 (@bmeck), Ryan's security concerns in #31, #29 (@jianli-chromium), and #10, #11, and #12 (@lrosenthal). I've tried to mark all the places that still need work with "TODO", but I may have missed some.


README.md
```diff
@@ -28,14 +28,222 @@ In addition, that format would include optional **signing** of the resources, wh
...
+Note that this is just an explainer, **not a specification**. We'll add more
+precision when we translate it to a spec.
+
+### Overall format
+
+The package is structured as follows. All "offsets" are 8-byte little-endian
+integers representing a byte offset from the start of the initial magic number.
+
+1. A magic number: `F0 9F 8C 90 F0 9F 93 A6` (🌐📦 in UTF-8).
+2. The length of the package (all bytes from the first of the initial magic
```
dimich-g: What makes the overall length of the package needed at the beginning?

README.md
```diff
@@ -28,14 +28,222 @@ In addition, that format would include optional **signing** of the resources, wh
...
+   number to the last of the final magic number) as a little-endian 8-byte
+   integer.
+3. A 1-byte count of metadata sections followed by that many of the following
```
dimich-g: As far as binary format goes, we might as well consider at least 32-bit alignment, since it may make memory-mapped files easier/faster to interpret.

README.md
```diff
@@ -28,14 +28,222 @@ In addition, that format would include optional **signing** of the resources, wh
...
+6. A sequence of HTTP responses constituting the main content of the package.
+   Mismatches with the manifest usually cause the package to fail to parse, as
+   described in the [manifest](#manifest) section, but mismatches with the index
+   might not.
+7. The length of the package, repeated from earlier. This allows tools to
+   operate on the package even if it's been prepended with a self-extracting
+   executable.
+8. The same magic number: `F0 9F 8C 90 F0 9F 93 A6` (🌐📦 in UTF-8).
+
+The MIME type of this format is `application/package`. TODO: Is that right?
```
dimich-g: I'd remove TODO. If it's not right, it'll be obvious on review/registration.
The package does need a mime type as it requires a different parser/handler typically, so the only issue may be actual names.

README.md
jyasskin: It's not essential. I included it as a way to validate that the length at the end is actually referring to a package, but the other constraints here probably do that just fine. I can remove it.

README.md
jyasskin: I suspect that "package" is too general to suggest, and that we should suggest something more specific like `webpackage` or `httppackage` (since it holds HTTP responses).

README.md
jyasskin: Right, I'm torn between allowing alignment for the binary parts, vs compressing the integers better. The amount of data here is pretty small, so I decided that decoding speed wasn't critical enough to worry about alignment. 'course, it's also not enough to really worry about the size.

README.md
```diff
@@ -28,14 +28,222 @@ In addition, that format would include optional **signing** of the resources, wh
...
+and [Extended Key Usage](https://tools.ietf.org/html/rfc5280#section-4.2.1.12)
+are TBD. It may or may not be important to prevent TLS serving certificates from
+being used to sign packages, in order to prevent cross-protocol attacks.
+
+### Manifest
+
+A package's manifest lists all of the resources included in that package, and
+any sub-packages the package depends on. It starts with at least the
+`Content-Type`, `Content-Location`, `Content-Length`, and `Date` headers, and
+must include one or more [signatures](#signatures).
```
dimich-g: Signatures are also optional. Not-signed packages may have a lot of utility since they can be produced locally and shared easily. The limits on contents are obvious, but for many use cases (packaging a movie that has its own DRM schema embedded, or a static snapshot of a page) the signing is not needed.

README.md
```diff
@@ -28,14 +28,222 @@ In addition, that format would include optional **signing** of the resources, wh
...
+```
+
+### Certificates
+
+The certificates block contains
+an
+[application/pkcs7-mime; smime-type=certs-only](https://tools.ietf.org/html/rfc5751) resource
+holding enough [X.509](https://tools.ietf.org/html/rfc5280) certificates that
+chains can be built from the certificates that sign the top-level
+and [sub-package](#sub-packages) manifests to root certificates recognized by
```
dimich-g: This makes it not clear what certificates and for what signature are included. The original idea went with a packages being authonomous units (since if signed, they most likely will be signed by separate domain owners, using separate certs etc). It's unclear why including certs needed to verify a subpackage is needed in a containing package, if that is what you meant.

README.md
jyasskin: My thought is that unsigned packages would omit the manifest. Can you think of a reason to include a manifest without signing it? Note that the index is always unsigned.

README.md
```diff
@@ -28,14 +28,222 @@ In addition, that format would include optional **signing** of the resources, wh
...
+must include one or more [signatures](#signatures).
+
+At least one of a manifest's signatures must be from a certificate that has a
+valid chain to a known root, and that is trusted to sign content from the
+[origin](https://html.spec.whatwg.org/multipage/browsers.html#concept-origin) of
+the manifest's `Content-Location`.
+
+If no signature gives the manifest authority to speak for its origin, or if any
+resource or sub-package doesn't match the description in the manifest, the whole
+package must fail to parse. TODO: Should there be a way to express that
```
dimich-g: Again, not signed resources have their use and may be used even wider than signed, at least initially. Likely need to preserve optional nature of signing.

README.md
```diff
@@ -28,14 +28,222 @@ In addition, that format would include optional **signing** of the resources, wh
...
 
-We also propose to remove the [fragment-based URL schema](https://w3ctag.github.io/packaging-on-the-web/#fragment-identifiers) from the spec as it is not clear what would be the use case supporting it.
+```
+URL offset length
+```
+
+Where offset and length are 64-bit integers, and the URLs are absolute. Offsets
+are interpreted relative to the first byte of the package. All resources in the
+package should be included in the index, even if they come from a different
+origin.
```
dimich-g: This needs to be clarified wrt subpackages. If a subpackage has its own index, it should not be duplicated. Realistically, the process of serving resources from the packages should include memory cache of indices of subpackages, so there is no need to flatten the index. If I understand the sentence correctly, of course.

README.md
jyasskin: At the moment, I have a single block of certificates, which are used to validate all the signatures anywhere. If we want to send the sub-package's certificates with that sub-package, we'd need to include extra certificate blocks. That said, `application/pkcs7-mime` can actually wrap both certificates and signed data, so we could store the manifests in that format and avoid having any separate certificate block.

README.md
jyasskin: We could give the top-level index a way to refer to subpackage indices, but I think that would constrain subpackages to be contiguous, which might break the ability to deduplicate shared dependencies. It seems more straightforward to ask the top-level package to regenerate the sub-package's indices.

README.md
jyasskin: We met and decided to use 8-byte numbers everywhere here, and to ignore alignment everywhere else.

README.md
```diff
@@ -28,14 +28,245 @@ In addition, that format would include optional **signing** of the resources, wh
...
+
+```
+https://package/manifest/url hashalgorithm-base64digest hashalgorithm2-base64digest2 ...
+```
+
+Note that because the sub-package may include sub-sub-packages by date, the top
+package may need to explicitly list those sub-sub-packages' hashes in order to
+be completely constrained.
+
+It should be possible to additionally constrain packages using semver, for the
```
bmeck: I don't think this is needed as we are not requesting updates via this format.

README.md
```diff
@@ -28,14 +28,245 @@ In addition, that format would include optional **signing** of the resources, wh
...
+benefit of NPM and other packaging systems, but I haven't explored that
+extension here.
+
+
+#### Signatures
+
+A [manifest](#manifest) is signed by one or more `Signature` headers. The
+message to sign consists of the length, followed by the concatenation of all
+headers except any `Signature` headers, followed by the body of the manifest.
```
bmeck: This needs to be reworded, `The message to sign` is confusing, are we talking about the whole package, the manifest, or an individual resource. I assume we are only talking about the manifest.

The wording here makes it sound like after generating a manifest we put the signatures on the resources.

README.md
```diff
@@ -28,14 +28,245 @@ In addition, that format would include optional **signing** of the resources, wh
...
+own [signature](#signatures), so if the main package trusts the sub-package's
+server, it could avoid specifying a version of the sub-package at all. However,
+this opens the main package up to downgrade attacks, where the sub-package is
+replaced by an older, vulnerable version, so we don't allow this option.
+
+If the main package wants to load either the sub-package it was built with or
+any upgrade, it can specify the date of the original sub-package:
+
+```
+https://package/manifest/url Date: Tue, 07 Feb 2017 01:05:54 GMT
```
bmeck: need to bikeshed the format here, HTTP Header on same line as other data seems an odd choice and hard to deal with.

jyasskin: Yep, the new CBOR-based format avoids this problem.

README.md
jyasskin: Works for me; removed.

README.md
jyasskin: I was talking about the manifest, and this wording got replaced when I switched to the CBOR format. See if the new wording in https://github.com/jyasskin/webpackage/tree/sketch-new-signing#manifest makes sense?

bmeck: +1

README.md
bmeck: +1

dimich-g:
This is a bit raw still, we'll need to clarify some topics, something that jumps at me:
- subpackages. seems w/o signed manifest it's impossible to have them
- subpackages - can they be external (by URL)
- index and manifest seem to have duplicated list of resource uris
- the COSI seems interesting but is it the right one?
- obviously, the examples should be updated, etc.

PR #37, opened 2017-02-08
Add a list of goals and non-goals.

jyasskin:
It's a fairly long list. Did I miss anything?

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,8 @@ Note the address bar indicating google.com. Also, since the content of USA Today
 * **external-dependencies**: Sub-packages can be "external" to the main package,
   meaning the UA will need to either fetch them separately or already have them.
   ([#35, App Installer Story](https://github.com/dimich-g/webpackage/issues/35))
-* **self-extracting**: Wrap packages in self-extracting executables.
+* **self-extracting**: Wrap packages in self-extracting executables. This is
+  required for Node to use the format.
 * **optional-components**: A package can identify resources that don't need to
   be authenticated before the UA starts loading the package.
 * **unsigned-streamed-loading**: Unsigned packages can be loaded as they
--- a/README.md
+++ b/README.md
@@ -20,6 +20,62 @@ Note the address bar indicating google.com. Also, since the content of USA Today
 - Can't request permissions
 - Can't be added to homescreen
 
+## Goals and non-goals
+
+### Goals
+
+* **authenticated-sharing**: Alice can download a package containing
+  a [PWA](https://developers.google.com/web/progressive-web-apps/checklist) from
+  origin `O`, and transmit it to their peer Bob, and then Bob can install
+  the PWA with a proof that it came from `O`.
+* **streamed-loading**: When a package is downloaded over TLS from its original
+  origin, the UA can load the package as it downloads.
+* **random-access**: When a package is stored on disk, the UA can access
+  arbitrary resources without a linear scan.
+* **sub-packages**: A package from origin `A` can contain resources from origin
+  `B` that can make same-origin requests back to `B`.
+* **crypto-agility**: Obsolete cryptographic algorithms can be replaced.
+* **cross-signatures**: Third-parties can vouch for packages by signing them.
+* **unsigned-content**: Alice can create their own package without a CA-signed
+  certificate, and Bob can view the content of the package.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -20,6 +20,62 @@ Note the address bar indicating google.com. Also, since the content of USA Today
...
+
+* **authenticated-sharing**: Alice can download a package containing
+  a [PWA](https://developers.google.com/web/progressive-web-apps/checklist) from
+  origin `O`, and transmit it to their peer Bob, and then Bob can install
+  the PWA with a proof that it came from `O`.
+* **streamed-loading**: When a package is downloaded over TLS from its original
+  origin, the UA can load the package as it downloads.
+* **random-access**: When a package is stored on disk, the UA can access
+  arbitrary resources without a linear scan.
+* **sub-packages**: A package from origin `A` can contain resources from origin
```
mrdewitt: The goal's name "sub-packages" refers to our current solution to the real goal which is "multiple origins can make same-origin requests"

README.md
```diff
@@ -20,6 +20,62 @@ Note the address bar indicating google.com. Also, since the content of USA Today
...
+* **confidential**: Store confidential data.
+* **streamed-generation**: Generate packages on the fly.
+  ([#6](https://github.com/dimich-g/webpackage/issues/6#issuecomment-275746125))
+* **non-origin-identity**: A package can be primarily identified as coming from
+  something other than a
+  [Web Origin](https://html.spec.whatwg.org/multipage/browsers.html#concept-origin).
+* **peer-streamed-loading**: When transferred over an unauthenticated
+  connection, the UA can load a signed package as it downloads. This is ruled
+  out by attacks involving dropped resources.
+
```
jyasskin: TODO: Add DRM support as a non-goal.

README.md
jyasskin: Changed to "multiple-origins".

README.md
jyasskin: Done.

README.md
```diff
@@ -20,6 +20,62 @@ Note the address bar indicating google.com. Also, since the content of USA Today
...
 
+## Goals and non-goals
+
+### Goals
+
+* **authenticated-sharing**: Alice can download a package containing
+  a [PWA](https://developers.google.com/web/progressive-web-apps/checklist) from
+  origin `O`, and transmit it to their peer Bob, and then Bob can install
+  the PWA with a proof that it came from `O`.
+* **streamed-loading**: When a package is downloaded over TLS from its original
```
dimich-g: I think it can be simplified by dropping 'TLS' requirement since unsigned content also makes sense. The TLS can be used in lieu of signature verification (but even that is more complicated, with countersigning) - so lets just keep streamed-loading as a goal regardless trust issues.

jyasskin: Done, especially given the fact that TLS doesn't guarantee resources aren't dropped.

bmeck:
I lean towards "self-extracting" as required for Node. It really is just adding a `length` field at the end of the file as long as we can check file signature somehow, don't really need anything else.

lrosenthol:
Two things

1 - To me, self-extracting implies that there is code in the package that
can extract its own contents.  I would actually find that to be an
anti-goal.  In fact, it almost screams security problem.

2 - For publication use cases, "non-origin-identity" is a
requirement/goal.  Do you want me to submit the PR on that?

bmeck:
@lrosenthol we can rename it to w/e else but that is the common name for archive formats that allow the behavior we have described. It is up to the runtime env to provide proper mechanics to load things safely. I just need the format to allow the archive to be concatenated on the end of a binary and find the start of the archive safely.

lrosenthol:
Ah - it's not that the package itself is self-extracting - but that it can
be used in a self-extracting context.  Then maybe just add "compatible"
(ie. self-extracting compatible)?

jyasskin:
I've marked that Node requires the packages to be compatible with self-extraction, and I'm almost certain we can accommodate it, but I've left it in the Nice-to-have section because Node support *at all* is just nice-to-have.

jyasskin:
Publication use cases are also nice-to-have, but packages with non-origin identities are hard enough to fit into web browsers that I think this PR is right to list it as a non-goal. We could explore it further in a separate issue/PR if you really want.

PR #46, opened 2017-04-12
Consolidate resource-keys into the index.

### LINKED ISSUE #41: Optimize the storage of the resource URLs in the manifest, index, and main content
From @dimich-g's https://github.com/dimich-g/webpackage/pull/36#issuecomment-290316360:

The `resource-key` structure (which holds a URL and possibly a set of request headers) is used as the key for the index, manifest, and main content. URLs can be fairly long, so we may want to optimize or compress that redundancy.

One idea for avoiding this duplication: Add a section of abbreviations that assigns each resource-key an index, and use those indices as the keys in the other three sections.

1. We could re-use the index as that section of abbreviations.
2. We probably don't want to re-use the manifest as that section, so that aggregating packages can build their own abbreviations.
3. The abbreviations need to be expanded before hashing or checking signatures.
4. We could shorten the abbreviation section even more by:
   1. Resolving each URL/resource-key relative to the previous one.
   1. Using the array index of the URL as its abbreviation instead of making the section name an abbreviation.

Another possibility would be to use a shared brotli dictionary to compress all three sections. The main downside here is that shared brotli dictionaries don't exist yet, and when they do, we may want to use a URL to identify which shared dictionary to use, which we'd again want to compress.

jyasskin:
This requires the index where it used to be optional.

Fixes #41.


README.md
```diff
@@ -150,15 +148,18 @@ Only information in the manifest and below can be trusted.
...
 
 ```cddl
-$section-name /= "content"
-$$section //= ("content" =>
-  {* resource-key => [response-headers: http-headers, body: bstr]}
-)
+$section-name /= "indexed-content"
+$$section //= ("indexed-content" => [
+  index: [* [resource-key, offset, ? length: uint] ],
+  responses: [* [response-headers: http-headers, body: bstr]],
```
jyasskin: I don't strictly need the nested array here: it could just be a list of alternating headers and bodies.

README.md
```diff
@@ -150,15 +148,18 @@ Only information in the manifest and below can be trusted.
...
+package is streamed.
 
 ```cddl
-$section-name /= "content"
-$$section //= ("content" =>
-  {* resource-key => [response-headers: http-headers, body: bstr]}
-)
+$section-name /= "indexed-content"
+$$section //= ("indexed-content" => [
+  index: [* [resource-key, offset, ? length: uint] ],
```
jyasskin: I don't strictly need this array either, but it's a little more difficult to tease out where the next index entry starts with the optional length. I could require lengths, but that wastes some space when a package isn't expected to be downloaded using http range requests.

README.md
```diff
@@ -167,54 +168,48 @@ resource-key = uri / http-headers
 http-headers = bstr
 ```
 
+A `uri` `resource-key` is equivalent to an `http-headers` block with ":method"
+set to "GET" and with ":scheme", ":authority", and ":path" headers set from the
```
mrdewitt: Later the resource key is described with a particular order, should that be reflected here (or pulled out and referenced in both places)

README.md
dimich-g: I think it makes sense and will make parser simpler.

README.md
dimich-g: Same here.

README.md
jyasskin: Having the array makes sense, or not-having it?

README.md
jyasskin: Done, thanks.

README.md
dimich-g: Having the array.

README.md
```diff
@@ -319,15 +312,19 @@ being used to sign packages, in order to prevent cross-protocol attacks.
 
 #### Validating resources
 
-For a resource to be valid, its hash needs to match all of the hashes provided
-in the manifest, although, like
+For a resource to be valid, its hash needs to appear in the list for each
```
dimich-g: Could you add a comment explaining how the hashes are mapped to resources? Is it based on the order in the array?

jyasskin: The idea is that we make a set of the allowed `hash-value`s, hash each resource, and then the resource is allowed if its hash appears in the set. I'll try to describe that more clearly.

jyasskin: How's this?

README.md
```diff
@@ -228,7 +227,7 @@ signed-manifest = {
 
 manifest = {
   metadata: manifest-metadata,
-  resources: [* [resource: resource-key, hashes: hashes]],
+  resource-hashes: {* hash-algorithm => [hash-value]},
```
dimich-g: Is the mapping resource->hash done here by using the same order of the hash-values as resources in indexed content? It's not explicitly said anywhere I think...

jyasskin: No, the list of hash values here is just a set. If the resource matches any of them, it's valid. I think https://github.com/dimich-g/webpackage/pull/46/files#diff-04c6e90faac2675aa89e2176d2eec7d8R316 says this.

PR #60, opened 2017-06-02
Store URL and status in HTTP2 pseudoheaders.

jyasskin:


go/webpack/cbor_format_test.go
```diff
@@ -31,10 +31,14 @@ func TestWriteCBOR(t *testing.T) {
 	pack := Package{
 		parts: []*PackPart{
 			&PackPart{
-				url:            staticUrl("https://example.com/index.html?query"),
-				requestHeaders: HTTPHeaders{},
-				status:         200,
+				requestHeaders: HTTPHeaders{
+					httpHeader(":method", "GET"),
```
dimich-g: I feel I should have asked before, but what is potential purpose to keep 'method' here? It seems to add to confusion because it's unclear how would that be processed by UA and what the method would exactly mean.

go/webpack/cbor_format_test.go
```diff
@@ -31,10 +31,14 @@ func TestWriteCBOR(t *testing.T) {
...
+					httpHeader(":scheme", "https"),
+					httpHeader(":authority", "example.com"),
+					httpHeader(":path", "/index.html?query"),
+				},
 				responseHeaders: HTTPHeaders{
+					httpHeader(":status", "200"),
```
dimich-g: Same question about 200 - do we need it as part of the format? If not, it's probably better to remove and add later if/when the need arises.

go/webpack/cbor_format_test.go
jyasskin: I'm not sure we need `:method`: it's here because it's in HTTP2, but I don't know of any methods besides GET that we'd want to support. I think this is something the folks at the IETF might be able to give us useful feedback about. I've filed #70 to track this.

go/webpack/cbor_format_test.go
jyasskin: I think we *do* have uses for other `:status`es: specifically, we should be able to store 301 and 302 redirections in the package. We could make 200 the default if the `:status` is omitted, but in HPACK it's a 1-byte overhead, so I didn't think omitting it was worthwhile.

go/webpack/text_format.go
```diff
@@ -123,7 +128,7 @@ func checkRequestHeadersInVary(requestHeaders, responseHeaders HTTPHeaders) erro
 		}
 	}
 
-	for _, header := range requestHeaders {
+	for _, header := range requestHeaders[4:] {
```
jyasskin: Done.

go/webpack/cbor_format_test.go
jyasskin: Is it all right if I submit this PR and send a separate one to remove `:method`? I think that'll do better to focus discussion on whether it's a good idea to diverge from HTTP2 here.

mrdewitt:
I think that the requested changes have been factored out, so this should be good to go.

PR #61, opened 2017-06-02
Rearrange writing the resource body so it doesn't have to materialize in memory.

jyasskin:
Using a Writer makes it easier to hash the resource at the same time it's being written to the CBOR serialization.

### DIFF
--- a/go/webpack/cbor/cbor_writer.go
+++ b/go/webpack/cbor/cbor_writer.go
@@ -48,16 +48,19 @@ func (cw *countingWriter) Flush() error {
 	return cw.w.Flush()
 }
 
-type compoundItem struct {
+type item struct {
 	*countingWriter
 	// nil for the root.
 	parent *compoundItem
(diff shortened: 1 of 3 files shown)


go/webpack/cbor/cbor_writer.go
```diff
@@ -100,96 +103,140 @@ func encodedSize(i uint64) int {
...
+}
+
+type BytesWriter struct {
+	item
+	remainingSize int64
+}
+
+func (bw *BytesWriter) Write(p []byte) (int, error) {
+	n, err := bw.item.Write(p)
+	bw.remainingSize -= int64(n)
```
mrdewitt: if bw.item.Write fails, is n negative?

jyasskin: According to https://golang.org/pkg/bufio/#Writer.Write and https://golang.org/pkg/io/#Writer, no: Go uses the err return instead of negative lengths to convey errors.

PR #55, opened 2017-05-19
Parse manifests

jyasskin:
There's no support yet for serializing the text format including manifest data, or serializing and signing the CBOR format.

PR #62, opened 2017-06-03
Write signed manifests to the CBOR format.

jyasskin:
This sits on top of #55, #60, #61, #63, #64, #65, and #66. Sorry that it's so long again.

### DIFF
--- a/go/webpack/cbor_format.go
+++ b/go/webpack/cbor_format.go
@@ -2,10 +2,12 @@ package webpack
 
 import (
 	"bytes"
+	"crypto"
 	"fmt"
 	"io"
 	"io/ioutil"
 	"os"
+	"sort"
 
 	"github.com/WICG/webpackage/go/webpack/cbor"
 )
@@ -32,7 +34,7 @@ func parseIndexedContent(reader *bytes.Reader, parts []*PackPart) error {
 
 func WriteCBOR(p *Package, to io.Writer) error {
 	// Write the indexed-content/responses array first in order to compute
-	// the offsets of each response within it.
+	// the offsets and hashes of each response within it.
 	tempResponsesFile, err := ioutil.TempFile("", "webpack-responses")
 	if err != nil {
 		return err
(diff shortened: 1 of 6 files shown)


go/webpack/cbor_format.go
```diff
@@ -108,29 +156,172 @@ func WriteCBOR(p *Package, to io.Writer) error {
...
+		metadataKeys = append(metadataKeys, key)
+	}
+	sort.Slice(metadataKeys, func(i, j int) bool {
+		return cbor.CanonicalLessStrings(metadataKeys[i], metadataKeys[j])
+	})
+	metadata := manifest.AppendMap(uint64(len(metadataKeys)))
+	for _, key := range metadataKeys {
+		metadata.AppendUTF8S(key)
+		switch key {
+		case "origin":
```
mrdewitt: micro nit: process "date" first to match the order in the spec

go/webpack/cbor_format_test.go
```diff
@@ -27,6 +42,17 @@ func hpackByteArray(headersAndValues ...string) []byte {
 	return append(cbor.Encoded(cbor.TypeBytes, len(result)), result...)
 }
 
+func hpackDecode(t *testing.T, encoded []byte) HTTPHeaders {
```
mrdewitt: is it more canonical to name the return value than to declare one in the first line of the function?

go/webpack/cbor_format_test.go
```diff
@@ -88,3 +114,209 @@ func TestWriteCBOR(t *testing.T) {
...
+	assert.EqualValues("certificates", d.read(d.decodeType(cbor.TypeText)))
+	numCertificates := d.decodeType(cbor.TypeArray)
+	certs := make([]*x509.Certificate, numCertificates)
+	for i := range certs {
+		certBytes := d.read(d.decodeType(cbor.TypeBytes))
+		certs[i], err = x509.ParseCertificate(certBytes)
+		require.NoError(err)
+	}
+	signingCert := certs[keyIndex]
+	signingCert.Verify(x509.VerifyOptions{
```
mrdewitt: Does this actually verify that signingCert corresponds to a DNSName of "example.com"?  Does it make sense to add a test for attempts to sign a package with no signatures corresponding to the origin?

go/webpack/cbor_format.go
jyasskin: Done.

go/webpack/cbor_format_test.go
jyasskin: I don't know, but I can do that instead.

go/webpack/cbor_format_test.go
jyasskin: Yes, and I've added an extra test with the wrong DNSName to check that.

And I've added a check that at least one signature covers the package's origin.

PR #63, opened 2017-06-07
Refine the algorithm for hashing a resource.

jyasskin:
Headers are decompressed before being hashed.


jyasskin:
I'm going to get these merged, and then send another PR to remove `:method`.

PR #64, opened 2017-06-07
Define how to pick a signing algorithm based on a certificate's public key.

jyasskin:
And add some Go code to sign a message into a byte string.


README.md
```diff
@@ -304,13 +305,15 @@ certificates are used to sign manifests.
...
-`rsa_pkcs1*` and the ones marked "SHOULD NOT be offered".
+generate it.
+The [TLS 1.3 signing algorithm](https://tlswg.github.io/tls13-spec/#rfc.section.4.2.3)
+is determined from the certificate's public key type:
+
+* RSA <= 3072 bits: rsa_pss_sha256
+* RSA, 3073-7680 bits: rsa_pss_sha384
+* secp256r1: ecdsa_secp256r1_sha256
+* secp384r1: ecdsa_secp384r1_sha384
+* secp521r1: ecdsa_secp521r1_sha512
```
jyasskin: Moved @mrdewitt's comment from https://github.com/WICG/webpackage/pull/55#discussion_r121008410:

> Can this list be expanded?

go/webpack/signutil.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+	return nil, fmt.Errorf("unknown public key type: %T", pk)
+}
+
+type rsaPSSSigner struct {
+	privKey *rsa.PrivateKey
+	hash    crypto.Hash
+}
+
+func (s rsaPSSSigner) sign(message []byte) ([]byte, error) {
```
jyasskin: Moved @mrdewitt's comment from https://github.com/WICG/webpackage/pull/55#discussion_r121012844:

> Any reason you didn't just use crypto.Signer? https://golang.org/src/crypto/crypto.go?s=3217:4212#L97

README.md
jyasskin: Yes, but the expansion can't be implemented in Go yet. e.g. ed25519 is an obvious extension, but it's missing from their crypto library. For now, it doesn't matter because there are no TLS certificates with public key types other than the ones here (except perhaps for really long RSA keys).

go/webpack/signutil.go
jyasskin: crypto.Signer requires that the caller hash the message, while this file incorporates the hash into the Sign() function. I could replace the rsa.SignPSS call with s.privKey.Sign(..., & PSSOptions{Hash:s.Hash}), but that doesn't seem any clearer, and the [documentation](https://golang.org/pkg/crypto/rsa/#PrivateKey.Sign) says "Common uses should use the Sign* functions in this package."

go/webpack/signutil.go
```diff
@@ -0,0 +1,160 @@
...
+		case bits <= 3072:
+			return rsaPSSSigner{pk, crypto.SHA256}, nil
+		case bits <= 7680:
+			return rsaPSSSigner{pk, crypto.SHA384}, nil
+		default:
+			return nil, fmt.Errorf("RSA key too big: %v bits", bits)
+		}
+	case *ecdsa.PrivateKey:
+		switch name := pk.Curve.Params().Name; name {
+		case "P-256":
```
mrdewitt: I wish these were pre-defined constants somewhere. But I didn't see that in a brief overview of crypto/elliptic

README.md
```diff
@@ -276,6 +276,7 @@ signature = {
   ; the public key, not the certificate, so we identify certificates by their
   ; index in the certificates array instead.
   keyIndex: uint,
+  ; Encoded as described in TLS 1.3.
```
mrdewitt: nit: link?

README.md
jyasskin: Interesting. Sounds good, to reduce the amount of flexibility in the protocol.

go/webpack/signutil.go
```diff
@@ -0,0 +1,160 @@
...
+
+type messageSigner interface {
+	sign(message []byte) ([]byte, error)
+}
+
+func signerForPrivateKey(pk crypto.PrivateKey) (messageSigner, error) {
+	switch pk := pk.(type) {
+	case *rsa.PrivateKey:
+		switch bits := pk.N.BitLen(); {
+		case bits <= 3072:
```
jyasskin: Right now, it's actually inequality, but I think you're right that we should just list explicit lengths.

README.md
jyasskin: Added.

go/webpack/signutil.go
jyasskin: I could use elliptic.P256().Params().Name.

go/webpack/signutil_test.go
```diff
@@ -0,0 +1,87 @@
...
+import (
+	"crypto"
+	"crypto/ecdsa"
+	"crypto/elliptic"
+	"crypto/rand"
+	"crypto/rsa"
+	"math/big"
+	"testing"
+
+	"github.com/dimich-g/webpackage/go/webpack"
```
jyasskin: Thanks, done in b8ae7b5797550d28f94edcbf461e6d96ac9071eb.

jyasskin:
I checked with @sleevi, who thinks this set is ok. We may need to add RSA-4096 in the future, but because intermediates and roots tend to be 2048 bits, those don't actually add security.

PR #66, opened 2017-06-07
Add a simple CBOR parser to make it easier to test the CBOR writer.

jyasskin:


go/webpack/cbor/cbor_reader.go
```diff
@@ -0,0 +1,71 @@
...
+	"encoding/binary"
+	"errors"
+	"fmt"
+	"io"
+)
+
+// A Decoder helps callers iterate through a CBOR buffer, extracting item
+// headers and string bodies as they go.
+type Decoder struct {
+	full []byte
```
mrdewitt: nit: not clear what "full" is at first glance.

jyasskin: Yeah, it's the full buffer that's being decoded. Does this just need a comment, or would you rather some other name?

mrdewitt: probably prefer something like cborBytes or cborBuffer but that might just be my verbose C++ history. I leave it up to you.

jyasskin: Sure, done.

PR #68, opened 2017-06-07
Update the signed one-origin example.

jyasskin:
This update was generated using #67, but we can commit it without all that code.


dimich-g:
This looks good. Merging...

PR #77, opened 2017-06-30
Describe the TAG's draft as out-of-date.

jyasskin:
I've heard from at least two people who started reading the TAG's document instead of this one because of how prominently we highlight it.


igrigorik:
Can we also update the TAG document to point to this one? Their version is one of the top links when you search for "web packaging". 

/cc @triblondon

triblondon:
That sounds like a sensible idea.  I'm happy to push that change.

triblondon:
Oh, actually since the TAG version is also published in W3C TR space I think I will have to pass the buck to @ylafon

jyasskin:
Thanks @triblondon and @ylafon. And I've taken @KenjiBaheux's suggestion to strike out the link, although I'm not sure that's necessary.

ylafon:
Yes, I already amended the README, and will fix both the old edcopy and in /TR/

jyasskin:
Thanks @ylafon!

PR #81, opened 2017-08-23
Start an Internet Draft for signed single responses.

jyasskin:
Preview at https://jyasskin.github.io/webpackage/draft-yasskin-origin-signed-responses/draft-yasskin-http-origin-signed-responses.html

This doesn't yet have a design, just some use cases and constraints on the design. I also haven't yet worked out how this would fit into a bundling proposal. It's possible that it doesn't.

@nyaxt @KenjiBaheux @mnot @mikewest

### DIFF
--- a/draft-yasskin-http-origin-signed-responses.md
+++ b/draft-yasskin-http-origin-signed-responses.md
@@ -106,9 +106,17 @@ The W3C WebAppSec group is investigating
 {{SRI}}. They need a way to transmit the signature with the
 response, which this proposal could provide.
 
-However, with the `integrity="ed25519-[public-key]"` attribute, they don't need
-the key that signs the signature (which is conveyed in that attribute) to be
-also trusted to sign arbitrary content for an origin.
+However, their needs also differ in some significant ways:
+
+1. The `integrity="ed25519-[public-key]"` attribute and CSP-based ways of
+   expressing a public key don't need the signing key to be also trusted to sign
+   arbitrary content for an origin.
+2. Some uses of SRI want to constrain subresources to be vouched for by a
+   third-party, rather than just being signed by the subresource's author.
+
+While we can design this system to cover both origin-trusted and simple-key
+signatures, we should check that this is better than having two separate systems
+for the two kinds of signatures.
 
 ## Offline websites {#uc-offline-websites}
 
--- a/draft-yasskin-http-origin-signed-responses.md
+++ b/draft-yasskin-http-origin-signed-responses.md
@@ -1,0 +1,308 @@
+---
+coding: utf-8
+
+title: Origin-signed HTTP Responses
+docname: draft-yasskin-http-origin-signed-responses
+category: std
+
+ipr: trust200902
+area: gen
+workgroup: http
+keyword: Internet-Draft
+
+stand_alone: yes
+pi: [comments, sortrefs, strict, symrefs, toc]
+
+author:
+ -
+    name: Jeffrey Yasskin
+    organization: Google
+    email: jyasskin@chromium.org
+
+informative:
+  SRI: W3C.REC-SRI-20160623
+
+--- abstract
+
+This document explores how a server can send particular responses that are
+authoritative for an origin, when the server itself is not authoritative for
+that origin. For now, it focuses on the constraints covering any such mechanism.
+
+--- note_Note_to_Readers
+
+Discussion of this draft takes place on the HTTP working group mailing list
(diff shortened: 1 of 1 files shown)


draft-yasskin-http-origin-signed-responses.md
```diff
@@ -0,0 +1,316 @@
...
+
+Intermediate
+: An entity that fetches signed HTTP responses from an author or another
+  intermediate and forwards them to another intermediate or a client.
+
+Client
+: An entity that uses a signed HTTP response and needs to be able to prove that
+  the author vouched for it as coming from its claimed origin.
+
+# Use cases
```
jyasskin: @igrigorik, could you check that these performance-related use cases for signed **single** responses make sense? (https://jyasskin.github.io/webpackage/draft-yasskin-origin-signed-responses/draft-yasskin-http-origin-signed-responses.html#use-cases)

I added some explicit text to the [packaging use cases](https://wicg.github.io/webpackage/draft-yasskin-webpackage-use-cases.html) to discourage folks from thinking of them as a performance tool, but with fewer use cases overall for non-bundled signed responses, I'm left with the performance ones here.

draft-yasskin-http-origin-signed-responses.md
```diff
@@ -0,0 +1,316 @@
...
+subresources from an intermediate CDN that's not authoritative, but require that
+those resources be signed by `O.com` so that the CDN couldn't modify the
+resources.
+
+~~~html
+<img logicalsrc="https://O.com/img.png"
+     physicalsrc="https://cdn.com/O.com/img.png">
+~~~
+
+This could be used for some of the same purposes as SRI ({{uc-sri}}).
```
igrigorik: The overall use case makes sense, but enumerating CDNs in markup is painful. Could this be elevated to a document-level opt-in (e.g. via response header or meta directive) that notifies UA that content on this origin can also be fetched from cdn.com? Specifically, say I detect that user is coming from Geo-region X, or mobile network Y, and I have edge caches optimized for each of those... I'd like to emit a header that notifies UA that it can fetch from X or Y; rewriting HTML is super painful.

draft-yasskin-http-origin-signed-responses.md
```diff
@@ -0,0 +1,316 @@
...
+
+## Subresource Integrity {#uc-sri}
```
igrigorik: This may be out of scope, but.. is content-addressable caching something to consider? You note that we want to sign headers+body, but what about cases where I only care about body -- e.g. I don't particularly care which CDN you fetch jquery.v.1.2.3.js from, as long as its content signature matches.

jyasskin: There are two separable features here:

1. Content-addressed caching where the resource original author is willing to vouch for the resource's actual URL. That is, if you have content with the same logical URL and a good signature, you can use it without contacting the CDN. (But this isn't actually content-addressing; it's URL-addressing again.) I think this is safe from a same-origin perspective, but I'm not sufficiently current with the SRI folks' discussion of content-addressing to know if it solves their privacy worries. @mikewest?
2. Cases where you only care about the content, and not its URL. I think this doesn't make sense for signed content: the point of a public keypair is that you can sign more than one thing with it, but then which data signed with the provided public key are you actually trying to use?

So ... I'm happy to consider content-addressing, but I think I don't understand what you mean by it in this context.

draft-yasskin-http-origin-signed-responses.md
igrigorik: FWIW, I'm skeptical that's a scalable way to do this, or even required one for this use case.. What I'm describing is another iteration on base tag.

draft-yasskin-http-origin-signed-responses.md
igrigorik: My question is mostly as an exercise to identify whether this is in or out of scope of this proposal. Specifically, you call out signing headers+body, but signing body only may have merit in context of sharing responses -- that said, you're right in this case you cant be signing with your key, so perhaps that's out of scope.

jyasskin: I've added a link to https://github.com/mikewest/signature-based-sri/issues/5 arguing that it's unsafe to sign just-a-body, with any key. :smiling_imp:

jyasskin:
I'm going to commit this so I can start discussion on the httpbis mailing list. Of course feel free to suggest more changes either here or there.

PR #82, opened 2017-08-28
Add notes to requirements saying when they conflict with re-using MHTML or ZIP.

jyasskin:
Preview in https://jyasskin.github.io/webpackage/alex-changes/draft-yasskin-webpackage-use-cases.html.

Did I miss anything obvious or say anything incorrect?

This also adds an explicit "Implementations are hard to get wrong" requirement.


jyasskin:
Hm, I omitted a date because I can't find a publication date in https://docs.oracle.com/javase/7/docs/technotes/guides/jar/jar.html. It's latest copyright is 2014, so maybe that's the right thing to use? Done.

PR #87, opened 2017-11-17
Add a security consideration to revoke keys exposed by a signing oracle.

jyasskin:
How's this, @agl, @sleevi?


agl:
Makes sense. Coincidentally, [here's](https://support.f5.com/csp/article/K21905460) a vulnerability from F5, published yesterday, that lets attackers get signatures from an online TLS key.

jyasskin:
Thanks. I would not have guessed that an "Adaptive Chosen Ciphertext attack" would lead to a signing oracle, which makes me worry that folks won't know to revoke their key after this sort of vulnerability.

PR #88, opened 2017-11-21
A rough proposal for signing HTTP exchanges.

jyasskin:
There's a preview at https://jyasskin.github.io/webpackage/design-signatures/draft-yasskin-http-origin-signed-responses.html#proposal

What's clearly broken about this? Is the complexity of allowing either plain responses or encapsulated request headers worth it?

I updated the validity algorithm to handle raw public keys and to return exactly which headers are trustworthy, but I did it quickly, so there could be mistakes there.

### DIFF
--- a/draft-yasskin-http-origin-signed-responses.md
+++ b/draft-yasskin-http-origin-signed-responses.md
@@ -217,20 +217,13 @@ sharing the exchange among themselves without going back to the intermeidate?
 
 The significant parts of an exchange are:
 
-* The request pseudo-header fields whose names are listed in that exchange's
-  `Signed-Headers` header field ({{signed-headers}}), in the order they appear
-  in that header field.  If a request pseudo-header field name from
-  `Signed-Headers` does not appear in the exchange's request pseudo-header
-  fields, the exchange has no significant parts.
-
-  If the request includes a `Host` header field but no `:authority` header
-  field, it is treated for the purpose of computing its significant parts as
-  having an `:authority` header field with the value of the `Host` header field.
-* The response header fields whose names are listed in that exchange's
-  `Signed-Headers` header field ({{signed-headers}}), in the order they appear
-  in that header field. If a response header field name from `Signed-Headers`
-  does not appear in the exchange's response header fields, the exchange has no
-  significant parts.
+* The method ({{!RFC7231}}, section 4) and effective request URI ({{!RFC7230}},
+  section 5.5) of the request.
+* The response status code ({{!RFC7231}}, section 6) and the response header
+  fields whose names are listed in that exchange's `Signed-Headers` header field
+  ({{signed-headers}}), in the order they appear in that header field. If a
+  response header field name from `Signed-Headers` does not appear in the
+  exchange's response header fields, the exchange has no significant parts.
 * The exchange's payload body ({{!RFC7230}}, section 3.3). Note that the payload
   body is the message body with any transfer encodings removed.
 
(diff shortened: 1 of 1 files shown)


draft-yasskin-http-origin-signed-responses.md
```diff
@@ -321,17 +348,485 @@ frames, or could
...
+
+: A string ({{!I-D.nottingham-structured-headers}}, section 4.2) containing a
+  [valid URL string](https://url.spec.whatwg.org/#valid-url-string).
+
+"certSha256"
+
+: Binary content ({{!I-D.nottingham-structured-headers}}, section 4.5) holding
+  the SHA-256 hash of the first certificate found at "certUrl".
+
+"ed25519Key"
```
jyasskin: @otherdaniel/@mikewest Does this look like it'll handle what https://github.com/mikewest/signature-based-sri needs? (The algorithm that uses it is at https://wicg.github.io/webpackage/design-signatures/draft-yasskin-http-origin-signed-responses.html#signature-validity)

mikewest: In broad strokes, I think this would support the SRI proposal. I'm not sure about some details, but if we were designing something _just_ for the SRI use case, I think we'd end up with something sufficiently similar to this that we should just squint at it until it's the same. I'm willing to live with some "cruft" (from my feature's perspective) in order to not invent a custom new thing just for my use case. :)

That said, https://github.com/mikewest/signature-based-sri/issues/5#issuecomment-345212824 suggested looking into https://tools.ietf.org/html/draft-cavage-http-signatures-08, which does similar things to this proposal, with a different spelling, and somewhat different characteristics (no certs, for instance). It does tack headers into the message to be signed, which is appealing (binding the resource to a `Content-Type`, for instance, might be valuable).

I know you've thought about that use case. What do you think about supporting it in this way?

jyasskin: This is definitely still open for tweaking. For example, I could replace the `ed25519Key=*base64` field with a CSP-style `key=ed25519-base64` [label](https://mnot.github.io/I-D/structured-headers/#label). I didn't because that takes an extra level of parsing to actually check the signature in this spec, but it's certainly possible.

It's plausible that we should have totally separate headers for the "give extra assurance when the physical and logical request are the same" case from the "assert a request/response pair that my normal certificate doesn't cover" case. Even for the extra-assurance case, draft-cavage-http-signatures has some problems:
1. Specifying the algorithm next to the key is bad practice nowadays because it leads to things like https://auth0.com/blog/critical-vulnerabilities-in-json-web-token-libraries/.
1. Using a keyId to have a higher-level protocol look up the key instead of having the header specify it leaves room for those protocols to shoot themselves in the foot. I assume SRI would make something like "ed25519-base64" the keyId, so you won't be affected, but it's a problem for adopting the header more broadly.
1. Using the `Digest` header as the only way to sign the body seems risky: we'd have to make sure that every path through the browser notices the presence of a Digest header and avoids using the body until it's checked.
1. I'm still pretty sure leaving the hostname out of the signed data is a mistake, even though it requires extra signatures for testing vs prod. This draft doesn't require people to specify the hostname either, but it doesn't hardcode omitting it.
1. draft-cavage-http-signatures doesn't expire signatures, meaning a single signed resource with a vulnerability forces the server to rotate the key.
1. New things should try to use mnot's structured headers, although that's probably easy to fix.

draft-yasskin-http-origin-signed-responses.md
```diff
@@ -321,17 +348,485 @@ frames, or could
...
+
+: Binary content ({{!I-D.nottingham-structured-headers}}, section 4.5) holding
+  an Ed25519 public key ({{!RFC8032}}).
+
+{:#signature-validityurl} "validityUrl"
+
+: A string ({{!I-D.nottingham-structured-headers}}, section 4.2) containing a
+  [valid URL string](https://url.spec.whatwg.org/#valid-url-string).
+
+"date" and "expires
```
yoavweiss: missing "\""

yoavweiss: Also: why do we need to reinvent `cache-control` here? Or do those have a different purpose?

jyasskin: 9991607 now describes why I'm reinventing cache-control.

jyasskin:
I'm going to merge this and publish it as an IETF draft so we can start discussing on the httpbis mailing list. More comments are, of course, welcome, either here or there.

PR #92, opened 2017-12-06
Rewrite (X, section Y) to (Section Y of X).

jyasskin:
In the hopes of getting deep links.

Pointed out in https://lists.w3.org/Archives/Public/ietf-http-wg/2017OctDec/0399.html.

However, https://tools.ietf.org/html/draft-ietf-httpbis-origin-frame-04#section-1 does have a deep link to <a href="./rfc7540#section-9.1.2">[RFC7540], Section&nbsp;9.1.2</a>, so I don't think this ordering is the fundamental problem.

I can't get any of these syntaxes to result in deep links from https://xml2rfc.tools.ietf.org/, so I'm not sure what the ietf-httpbis drafts are doing to get them. The httpsbis editors' drafts do *not* have deep links: http://httpwg.org/http-extensions/expect-ct.html#response-header-field-syntax

@MikeBishop, halp! 😺

### DIFF
--- a/draft-yasskin-http-origin-signed-responses.md
+++ b/draft-yasskin-http-origin-signed-responses.md
@@ -76,7 +76,7 @@ Author
 Exchange (noun)
 : An HTTP request/response pair. This can either be a request from a client and
 the matching response from a server or the request in a PUSH_PROMISE and its
-matching response stream. Defined by {{!RFC7540}} section 8.
+matching response stream. Defined by Section 8 of {{!RFC7540}}.
 
 Intermediate
 : An entity that fetches signed HTTP exchanges from an author or another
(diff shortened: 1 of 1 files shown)


jyasskin:
It turns out that https://tools.ietf.org/html/draft-yasskin-http-origin-signed-responses-01 does have deep links because of magic from https://tools.ietf.org/tools/rfcmarkup/ that's not present in xml2rfc.

They're missing in https://tools.ietf.org/html/draft-yasskin-http-origin-signed-responses-01#page-10's "Section 4.2.3 of [I-D.ietf-tls-tls13]", so I'll do a little digging to see what's broken there.

jyasskin:
Ah, it's probably the commented-out "section x of draft-y markup" in https://tools.ietf.org/tools/rfcmarkup/code, so deep links into drafts probably just won't work.

PR #95, opened 2017-12-10
remove unneeded/misleading metadata

reschke:
(for instance, this is not a work item of the HTTP working group)

### DIFF
--- a/draft-yasskin-http-origin-signed-responses.md
+++ b/draft-yasskin-http-origin-signed-responses.md
@@ -6,9 +6,6 @@ docname: draft-yasskin-http-origin-signed-responses-latest
 category: std
 
 ipr: trust200902
-area: gen
-workgroup: http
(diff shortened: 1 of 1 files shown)


jyasskin:
Oh, that makes sense. Thanks! I wish I could remember what caused me to include those three elements so I could try to fix that documentation.

I have the same mistake in `draft-yasskin-dispatch-web-packaging.md` and `draft-yasskin-webpackage-use-cases.md` except with Dispatch as the workgroup. There's nothing special about Dispatch that would mean it's right to use them as the workgroup, right?

reschke:
If the draft is not an official work item of the WG, it shouldn't say so on the first page.

(FWIW, the related rule is a bit ambiguous as it only says what to put there if a draft *is* a WG draft -- but that's my interpretation...)

jyasskin:
Thanks, done.

PR #96, opened 2017-12-11
Add EdDSA certificates to the set with defined signature algorithms.

jyasskin:
I used the pre-hashed variants for those and switched the raw ed25519
public key to use ed25519ph, with an Open Question for whether that's
the right choice.

Ilari Liusvaara pointed out that draft-ietf-curdle-pkix is nearly published, so it makes sense to include its public key types in this document now. 

@sleevi, is this the right choice?

[Preview](https://jyasskin.github.io/webpackage/define-ed25519-certs/draft-yasskin-http-origin-signed-responses.html#signature-validity), [Diff](https://tools.ietf.org/rfcdiff?url1=https://jyasskin.github.io/webpackage/draft-yasskin-http-origin-signed-responses.txt&url2=https://jyasskin.github.io/webpackage/define-ed25519-certs/draft-yasskin-http-origin-signed-responses.txt)

### DIFF
--- a/draft-yasskin-http-origin-signed-responses.md
+++ b/draft-yasskin-http-origin-signed-responses.md
@@ -395,11 +395,17 @@ there are no non-significant response header fields in the exchange.
          : ecdsa_secp384r1_sha384 as defined in Section 4.2.3 of
            {{!I-D.ietf-tls-tls13}}.
 
+         Ed25519 and Ed448:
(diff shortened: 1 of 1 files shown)


sleevi:
There are no plans at this time to support such certificates in Chrome

jyasskin:
Thanks Ryan. I'll drop this until some client says they really want it.

PR #98, opened 2017-12-21
Start writing the bundling spec.

jyasskin:
[Preview](https://jyasskin.github.io/webpackage/bundles/draft-yasskin-dispatch-bundled-exchanges.html)

This fills a similar niche as [MHTML](https://tools.ietf.org/html/rfc2557), but
* is random access,
* efficiently encodes 8-bit payloads,
* can represent multiple top-level resources,
* can represent content-negotiated resources,
* and possibly other advantages.

I believe that combined with [signed exchanges](https://wicg.github.io/webpackage/draft-yasskin-http-origin-signed-responses.html), bundles fill all of the [web packaging use cases](https://wicg.github.io/webpackage/draft-yasskin-webpackage-use-cases.html) except for "Third-party security review", which will probably want to insist on a consistent version of contained resources.


draft-yasskin-dispatch-bundled-exchanges.md
```diff
@@ -0,0 +1,436 @@
...
+
+A valid request item *R*:
+
+* Matches the above CDDL.
+* Has only lower-case ASCII keys, matching the requirement in Section 8.1.2 of
+  {{?RFC7540}}.
+* Has exactly two keys starting with a ':' character, ':method' and ':url'.
+* *R*\[':method'] is an HTTP method defined as cacheable (Section 4.2.3 of
+  {{!RFC7231}}) and safe (Section 4.2.1 of {{!RFC7231}}), as required for
+  PUSH_PROMISEd requests (Section 8.2 of {{?RFC7540}}).
```
annevk: This is not really good enough for browsers I think. You need to have a safelist.

draft-yasskin-dispatch-bundled-exchanges.md
```diff
@@ -0,0 +1,436 @@
...
+* *R*\[':url'] is an absolute URI (Section 4.3 of {{!RFC3986}}).
```
annevk: It's unclear how to validate this in browsers. There's no such primitive.

draft-yasskin-dispatch-bundled-exchanges.md
```diff
@@ -0,0 +1,436 @@
...
+
+A valid request item *R* is interpreted as an HTTP request by interpreting
+*R*\[':method'] as the request's method (Section 4 of {{!RFC7231}}),
+*R*\[':url'] as the request's effective request URI (Section 5.5 of
+{{!RFC7230}}), and the remaining key/value pairs as the request's header fields.
```
annevk: Those key/value pairs are suitably constrained somehow? What's the parsing for them?

jyasskin: They're HTTP headers, so we should do whatever https://fetch.spec.whatwg.org/#http-network-fetch does.

Or are you worried about how claimed request headers are matched against browser request headers? That's not written yet. I was hoping to use the matching rules for PUSH_PROMISEs, but those aren't written yet either.

draft-yasskin-dispatch-bundled-exchanges.md
annevk: In the cache section, which is somewhat optional in a way.

draft-yasskin-dispatch-bundled-exchanges.md
annevk: Can I go with all three? E.g., `http://example.com/%` is not an "absolute URI", but will parse per the URL parser.

draft-yasskin-dispatch-bundled-exchanges.md
annevk: Well they're not quite the same as ordinary HTTP headers, since they are delimited differently, right? So e.g., can you have a 0x0A byte in the value?

annevk:
(Note that I replied even in the feedback GitHub considers "outdated".)

annevk:
Are you going to leave it to Fetch to drop the response body for a `HEAD` request? Or will that be an error to codify? It seems to me that you'd want to forbid `HEAD` to be encoded and only support it on the retrieval side and do the appropriate translations to `GET` minus response body there.

draft-yasskin-dispatch-bundled-exchanges.md
jyasskin: HTTP/2 headers are also delimited in a way that lets their values include a 0x0A byte. I wonder if browsers handle that correctly...

I suspect the right thing to do here is to explicitly create a [header list](https://fetch.spec.whatwg.org/#concept-header-list) from this, but since that concept doesn't exist in the IETF, I'll have to give up on the idea of making this an IETF draft.

I'm going to ping some folks to see if I'll hurt any feelings by doing that.

annevk: Oooh wow, I didn't know that about H/2. That's another reason to get a web-platform-tests setup.

annevk:
(Another thought I had was to take inspiration from the Cache API, as sort of the API surface on top of this format. The Cache API only supports `GET` thus far, but it does have `Vary` support. Haven't fully explored whether it's a complete fit, but it's interesting to think about as a way to avoid introducing too many new primitives.)

jyasskin:
I believe this is finally ready for a re-review. It's missing 3 things that are going to be important, but that I'd like to do in a subsequent patch:

1. The algorithm to take the request a browser would generate, e.g. with multiple values in its `Accept` header, and find a matching request inside a bundle, where requests will generally only have 1 value in their `Accept` headers.
1. Compression: the index can probably just be brotli-compressed, but response headers should be able to take advantage of a shared dictionary.
1. Signatures covering multiple resources: this both saves on signature verification costs and ensures that resources have matching versions.

draft-yasskin-dispatch-bundled-exchanges.md
```diff
@@ -0,0 +1,675 @@
...
+      * method is `pseudos[':method']`,
+      * url is `parsedUrl`,
+      * header list is `headers`, and
+      * client is null.
+
+   1. Let `streamOffset` be `sectionsStart + section-offsets\["responses"].offset
+      + offset`. That is, offsets in the index are relative to the start of the
+      "responses" section.
+   1. If `offset + length` is greater than
+      `section-offsets\["responses"].length`, return an error.
```
kinu: nit: extra \\

draft-yasskin-dispatch-bundled-exchanges.md
```diff
@@ -0,0 +1,675 @@
...
+   error, return that error.
+
+1. Let `url` be the result of parsing ({{URL}}) `urlString` with no base URL.
+
+1. If `url` is a failure, its fragment is not null, or it includes credentials,
+   return an error.
+
+1. Set `metadata`'s "manifest" item to `url`.
+
+### Parsing the critical section {#critical-section}
```
kinu: Wasn't fully unsure what could come here?

draft-yasskin-dispatch-bundled-exchanges.md
```diff
@@ -0,0 +1,675 @@
...
+
+~~~ cddl
+critical = [*tstr]
+~~~
+
+To parse the critical section, given its `sectionContents` and the `metadata`
+struct to fill in, the parser MUST do the following:
+
+1. Let `critical` be the result of parsing `sectionContents` as a CBOR item
+   matching the above `critical` rule ({{parse-cbor}}. If `critical` is an
```
kinu: nit: missing ) after {{parse-cbor}}

draft-yasskin-dispatch-bundled-exchanges.md
```diff
@@ -0,0 +1,675 @@
...
+1. If `stream.currentOffset + payloadLength != requestMetadata.offset +
+   requestMetadata.length`, return an error.
+
+1. Let `body` be a new body ({{FETCH}}) whose stream is a tee'd copy of `stream`
+   starting at the current offset and ending after `payloadLength` bytes.
+
+   TODO: Add the rest of the details of creating a `ReadableStream` object.
+
+1. Let `response` be a new response ({{FETCH}}) whose:
+   * Url list is `request`'s url list,
```
kinu: "Parsing the index section" seems to say request has one url, do you mean it could have a list / redirect chain?

draft-yasskin-dispatch-bundled-exchanges.md
```diff
@@ -0,0 +1,675 @@
...
+
+A bundle is parsed from a stream of bytes, which is assumed to have the attributes and operations described in {{stream-operations}}:
+
+Bundle parsers support two operations, defined in {{semantics-load-metadata}}
+and {{semantics-load-response}} each of which can return an error instead of
+their normal result.
+
+## Stream attributes and operations {#stream-operations}
+
+* A sequence of **available** bytes. As the stream delivers bytes, these are
```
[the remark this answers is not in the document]
jyasskin: Sure.

draft-yasskin-dispatch-bundled-exchanges.md
```diff
@@ -0,0 +1,675 @@
...
+1. Seek to the start of `stream`.
+
+1. Wait for 10 bytes to be available from `stream`.
+
+1. If the wait fails or if reading 10 bytes from `stream` doesn't return the
+   bytes with hex encoding "84 48 F0 9F 8C 90 F0 9F 93 A6" (the CBOR encoding of
+   the 4-item array initial byte and 8-byte bytestring initial byte, followed by
+   🌐📦 in UTF-8), return an error.
+
+1. Let `sectionOffsetsLength be the result of getting the length of the CBOR
```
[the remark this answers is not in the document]
jyasskin: Thanks.

draft-yasskin-dispatch-bundled-exchanges.md
jyasskin: Thanks.

draft-yasskin-dispatch-bundled-exchanges.md
jyasskin: Thanks.

draft-yasskin-dispatch-bundled-exchanges.md
jyasskin: If we add more sections in the future, and the publisher wants to make sure parsers don't just ignore one of those sections because they don't understand it, this section points out the ones not to skip.

I'm not certain this is the right way to do this. PNG does it with a particular capitalization in the section names: https://tools.ietf.org/html/rfc2083#page-13

draft-yasskin-dispatch-bundled-exchanges.md
jyasskin: This is just an asymmetry in how requests and responses default their values: https://fetch.spec.whatwg.org/#concept-request-url-list defaults from the request's url, while https://fetch.spec.whatwg.org/#concept-response-url is hard-coded to point to the response's url list. I believe both URL lists should always have 1 value for exchanges in the bundle.

I suspect Fetch will actually wind up copying or extending this response when serving browser-generated requests from bundles, so that a redirect into a bundle can work correctly.

draft-yasskin-dispatch-bundled-exchanges.md
```diff
@@ -0,0 +1,675 @@
...
+significantly affect performance when the bundle is loaded from a network
+stream.
+
+A bundle is parsed from a stream of bytes, which is assumed to have the attributes and operations described in {{stream-operations}}:
+
+Bundle parsers support two operations, defined in {{semantics-load-metadata}}
+and {{semantics-load-response}} each of which can return an error instead of
+their normal result.
+
+## Stream attributes and operations {#stream-operations}
```
jyasskin: @domenic, [here's the use of streams](https://jyasskin.github.io/webpackage/bundles/draft-yasskin-dispatch-bundled-exchanges.html#stream-operations) I mentioned on #whatwg. I'm thinking about re-doing it in terms of [`ReadableStream`](https://fetch.spec.whatwg.org/#readablestream) despite your advice because https://jyasskin.github.io/webpackage/bundles/draft-yasskin-dispatch-bundled-exchanges.html#load-response needs to create a [body's stream](https://fetch.spec.whatwg.org/#concept-body), which is defined as a `ReadableStream`. The stream a bundle's parsed from is likely to be a response body anyway, so it'll probably be a ReadableStream even if I pretend it isn't.

draft-yasskin-dispatch-bundled-exchanges.md
```diff
@@ -0,0 +1,675 @@
...
+  URL:
+    target: https://url.spec.whatwg.org/
+    title: URL
+    author:
+      org: WHATWG
+    date: Living Standard
+
+--- abstract
+
+Bundled exchanges provide a way to bundle up groups of HTTP request+response
```
annevk: request-response?

draft-yasskin-dispatch-bundled-exchanges.md
```diff
@@ -0,0 +1,675 @@
...
+pairs to transmit or store them together. The component exchanges can be signed
+using {{?I-D.yasskin-http-origin-signed-responses}} to establish their
+authenticity.
```
annevk: I'd leave this out, since the signing is controversial and not required.

draft-yasskin-dispatch-bundled-exchanges.md
```diff
@@ -0,0 +1,675 @@
...
+
+--- note_Note_to_Readers
+
+Discussion of this draft takes place on the ART area mailing list
+(art@ietf.org), which is archived
+at <https://mailarchive.ietf.org/arch/search/?email_list=art>.
```
annevk: It seems discussion also takes place on GitHub?

draft-yasskin-dispatch-bundled-exchanges.md
```diff
@@ -0,0 +1,675 @@
...
+To satisfy the use cases in {{?I-D.yasskin-webpackage-use-cases}}, this document
+proposes a new bundling format to group HTTP resources. Several of the use cases
+require the resources to be signed: that's provided by bundling signed exchanges
+({{?I-D.yasskin-http-origin-signed-responses}}) rather than natively in this
+format.
+
+## Terminology
+
+Exchange (noun)
+: An HTTP request/response pair. This can either be a request from a client and
```
annevk: I'd pick one separator and stick with it.

draft-yasskin-dispatch-bundled-exchanges.md
```diff
@@ -0,0 +1,675 @@
...
+  appended to the available bytes.
+* An **EOF** flag that's true if the available bytes include the entire stream.
+* A **current offset** within the available bytes.
+* A **wait for N bytes** operation, which blocks until the available bytes
+  include at least N bytes past the current offset. A wait operation fails if
+  the stream ends before enough bytes are received, either due to a network
+  error or because the stream has a finite length.
+* A **seek** operation to change the current offset, relative to either the
+  beginning of the available bytes or to the old current offset. A seek past the
+  end of the available bytes is an *error in this specification*.
```
annevk: "error in this specification" doesn't appear to mean anything. Wouldn't it be better to say that seek or read can return an error and that callers of them need to account for that?

draft-yasskin-dispatch-bundled-exchanges.md
```diff
@@ -0,0 +1,675 @@
...
+  of the available bytes is also an error in this specification.
+
+## Load a bundle's metadata {#semantics-load-metadata}
+
+This takes the bundle's stream and returns a struct ({{INFRA}}) of metadata
+containing at least items named:
+
+requests
+
+: A map ({{INFRA}}) whose keys are the HTTP requests ({{FETCH}}) for the
```
annevk: What Fetch defines as "requests" are not necessarily HTTP requests. They're more broad.

draft-yasskin-dispatch-bundled-exchanges.md
```diff
@@ -0,0 +1,675 @@
...
+This operation can be completed without inspecting bytes other than those that
+make up the loaded response, although higher-level operations like proving that
+an exchange is correctly signed ({{I-D.yasskin-http-origin-signed-responses}})
+may need to load other responses.
+
+Note that this operation uses the metadata for a particular request returned by
+{{semantics-load-metadata}}, while a client will generally want to load the
+response for a request that the client generated. This specification does not
+define how a client determines the best available bundled response, if any, for
+that client-generated request.
```
annevk: Why not? We define this for the Cache API, I think ideally we use the same algorithm here.

draft-yasskin-dispatch-bundled-exchanges.md
```diff
@@ -0,0 +1,675 @@
...
+
+# Format {#format}
+
+## Top-level structure {#top-level}
+
+*This section is non-normative.*
+
+The bundle is roughly a CBOR item ({{?I-D.ietf-cbor-7049bis}}) with the
+following CDDL ({{?I-D.ietf-cbor-cddl}}) schema, but bundle parsers are required
+to successfully parse some byte strings that aren't valid CBOR. For example,
```
annevk: Is there still use in CBOR then? Doesn't the MIME type lie with +cbor if you cannot actually use a CBOR parser?

draft-yasskin-dispatch-bundled-exchanges.md
```diff
@@ -0,0 +1,675 @@
...
+1. Let `knownSections` be the subset of the {{section-name-registry}} that this
+   client has implemented.
+
+1. Let `ignoredSections` be an empty set.
+
+1. For each `"name"` key in `section-offsets`, if `"name"`'s specification in
+   `knownSections` says not to process other sections, add those sections' names
+   to `ignoredSections`.
+
+1. Let `metadata` be a struct ({{INFRA}}) with no items.
```
annevk: The idea with structs is that they are immutable.

draft-yasskin-dispatch-bundled-exchanges.md
```diff
@@ -0,0 +1,675 @@
...
+1. For each pair `name`/`value` in `item`:
+   1. If `name` contains any non-ASCII or upper-case characters, return an
+      error. This matches the requirement in Section 8.1.2 of {{?RFC7540}}.
+   1. If `name` starts with a ':':
+      1. If `pseudos[name]` exists, return an error.
+      1. Set `pseudos[name]` to `value`.
+      1. Continue.
+   1. If `name` or `value` doesn't satisfy the requirements for a header in
+      {{FETCH}}, return an error.
+   1. If `headers` contains ({{FETCH}}) `name`, return an error.
```
annevk: You cannot have multiple headers of the same name? Might be worth calling out you cannot encode cookies.

draft-yasskin-dispatch-bundled-exchanges.md
jyasskin: I think you mean that the set of item names is immutable. It's simple to change this to a map, so I've done that. The set of keys is only not-fixed because I wanted to let other specifications add section names that might define new metadata, but I'm not totally sure that's a good idea, so this might go back to a struct later.

draft-yasskin-dispatch-bundled-exchanges.md
annevk: Even those are more broad as they carry fields HTTP requests don't have. I guess you could say requests whose url list contains a single URL whose scheme is an HTTP(S) scheme (unless this is restricted to secure contexts?).

annevk:
> Do you think I should say that clients MAY reject bundles that aren't a valid CBOR item? Or something else?

1. I think we want it to be fully deterministic across all conforming (browser) implementations how they would handle all possible sequences of bytes. I.e., at least as good as the HTML parser.
2. I don't think we should use +cbor if that isn't always the case or required to be the case.

Re: "best available bundled response" I'd prefer it as an open issue if you plan on addressing it rather than it saying it's implementation-defined.

(Note that we might get rid of pairs still: https://github.com/whatwg/infra/issues/127.)

jyasskin:
1. I haven't added the MAY.
1. I've removed the +cbor from the MIME type.
1. I marked the "best available bundled response" as a TODO.

Thanks!

draft-yasskin-dispatch-bundled-exchanges.md
jyasskin: I've removed the attempt to subset the kinds of Fetch requests this operation can return. If it would noticeably simplify the request-matching function I need to write, I might add the subsetting back later, but I don't really expect that.

jyasskin:
I'm going to merge this so it's easier to refer to and so I can start sending PRs to modify it, but please feel free to keep commenting and filing issues against it.

PR #108, opened 2018-01-17
Sign just the headers and identify a header to guard the payload.

jyasskin:
[Preview](https://jyasskin.github.io/webpackage/incremental-validity/draft-yasskin-http-origin-signed-responses.html), [Diff](https://tools.ietf.org/rfcdiff?url1=https://jyasskin.github.io/webpackage/draft-yasskin-http-origin-signed-responses.txt&url2=https://jyasskin.github.io/webpackage/incremental-validity/draft-yasskin-http-origin-signed-responses.txt) (When TravisCI recovers.)

This limits the length of the signed data, which helps validation
routines use the non-prehashed variant of ed25519, and if the MI header
field is used, it allows clients to incrementally trust the beginning
chunks of the payload as they are received instead of needing to wait
for the whole thing.

### DIFF
--- a/draft-yasskin-http-origin-signed-responses.md
+++ b/draft-yasskin-http-origin-signed-responses.md
@@ -157,11 +157,11 @@ The `Signature` header is a Structured Header as defined by
 {{!I-D.ietf-httpbis-header-structure}}).
 
 Each parameterised label MUST have parameters named "sig", "integrity",
-"validityUrl", "date", and "expires", and either "certUrl" and "certSha256"
-parameters or an "ed25519Key" parameter. This specification gives no meaning to
-the label itself, which can be used as a human-readable identifier for the
-signature (see {{parameterised-binary}}). The present parameters MUST have the
-following values:
+"validityUrl", "date", and "expires". Each parameterised label MUST also have
+either "certUrl" and "certSha256" parameters or an "ed25519Key" parameter. This
+specification gives no meaning to the label itself, which can be used as a
+human-readable identifier for the signature (see {{parameterised-binary}}). The
+present parameters MUST have the following values:
 
 "sig"
 
--- a/draft-yasskin-http-origin-signed-responses.md
+++ b/draft-yasskin-http-origin-signed-responses.md
@@ -444,14 +444,18 @@ there are no non-significant response header fields in the exchange.
    is present of `certificate-chain` or `ed25519Key`. Otherwise, return
    "invalid".
 
-   Note that the above algorithm can determine that an exchange's headers are
-   potentially-valid before the exchange's payload is received. The client MAY
-   process those headers as soon as they are validated. If `integrity`
-   identifies a header field like `MI` ({{?I-D.thomson-http-mice}}) that can
-   incrementally validate the payload, the client MAY also incrementally process
-   the validated parts of the payload as soon as they are validated. The client
-   MUST NOT process any part of the headers or payload before it has been
-   validated either by the signature or the header field named by `integrity`.
+Note that the above algorithm can determine that an exchange's headers are
+potentially-valid before the exchange's payload is received. Similarly, if
+`integrity` identifies a header field like `MI` ({{?I-D.thomson-http-mice}})
+that can incrementally validate the payload, early parts of the payload can be
+determined to be potentially-valid before later parts of the payload.
+Higher-level protocols MAY process parts of the exchange that have been
+determined to be potentially-valid as soon as that determination is made but
+MUST NOT process parts of the exchange that are not yet potentially-valid.
+Similarly, as the higher-level protocol determines that parts of the exchange
+are actually valid, the client MAY process those parts of the exchange and MUST
+wait to process other parts of the exchange until they too are determined to be
+valid.
 
 ## Updating signature validity ## {#updating-validity}
 
--- a/draft-yasskin-http-origin-signed-responses.md
+++ b/draft-yasskin-http-origin-signed-responses.md
@@ -100,9 +100,9 @@ appear in all capitals, as shown here.
 
 As a response to an HTTP request or as a Server Push (Section 8.2 of
 {{!RFC7540}}) the server MAY include a `Signed-Headers` header field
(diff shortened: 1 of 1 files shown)


draft-yasskin-http-origin-signed-responses.md
```diff
@@ -147,24 +147,33 @@ to include in the `Signed-Headers` header field.
...
 {{!I-D.ietf-httpbis-header-structure}}) of parameterised labels (Section 4.4 of
 {{!I-D.ietf-httpbis-header-structure}}).
 
-Each parameterised label MUST have parameters named "sig", "validityUrl",
-"date", and "expires", and either "certUrl" and "certSha256" parameters or an
-"ed25519Key" parameter. This specification gives no meaning to the label itself,
-which can be used as a human-readable identifier for the signature (see {{parameterised-binary}}). The present
-parameters MUST have the following values:
+Each parameterised label MUST have parameters named "sig", "integrity",
+"validityUrl", "date", and "expires", and either "certUrl" and "certSha256"
```
nyaxt: Not for this commit, but this line was a bit hard for me to understand. Can we split to something like:
```
Each parameterised label MUST have parameters named "sig", "integrity", "validityUrl", "date", and "expires".
In addition, each parameterised label MUST have either "certUrl" and "certSha256" parameters, or an "ed25519Key" parameter.
```

draft-yasskin-http-origin-signed-responses.md
```diff
@@ -286,18 +288,15 @@ extended diagnostic notation from {{?I-D.ietf-cbor-cddl}} appendix G:
...
-  "response",
-  [
-    ':status', '200',
-    'content-type', 'text/html'
-  ],
-  "payload",
-  '<!doctype html>\n<html>...'
+  {
+    ':url': 'https://example.com/'
+    ':method': 'GET',
```
nyaxt: I assume the keys need to be sorted?
https://wicg.github.io/webpackage/draft-yasskin-http-origin-signed-responses.html#canonical-cbor

draft-yasskin-http-origin-signed-responses.md
```diff
@@ -286,18 +288,15 @@ extended diagnostic notation from {{?I-D.ietf-cbor-cddl}} appendix G:
...
+  },
+  {
+    'digest': 'SHA-256=20addcf7368837f616d549f035bf6784ea6d4bf4817a3736cd2fc7a763897fe3',
+    ':status': '200',
+    'content-type': 'text/html'
```
nyaxt: this too?

draft-yasskin-http-origin-signed-responses.md
nyaxt: sorry maybe this is already sorted in "the bytewise lexicographic order"

jyasskin: Exactly: because the encoding of a string starts with its length, shorter strings sort before longer ones. This isn't obvious from the definition, but I do have the example of "z" sorting before "aa".

There's no actual need for the diagnostic form to appear in sorted order anyway, since canonicalization happens when encoding a semantic CBOR item to bytes, but I did sort it just to demonstrate the order.

draft-yasskin-http-origin-signed-responses.md
jyasskin: Done.

draft-yasskin-http-origin-signed-responses.md
```diff
@@ -351,21 +350,34 @@ This algorithm also accepts an `allResponseHeaders` flag, which insists that
...
-   `exchange`, then return "invalid".
+   `headers`, then return "invalid".
 1. Let:
    * `signature` be the signature (binary content in the parameterised value's
      "sig" parameter).
+   * `integrity` be the signature's "integrity" parameter.
    * `certUrl` be the signature's "certUrl" parameter, if any.
    * `certSha256` be the signature's "certSha256" parameter, if any.
    * `ed25519Key` be the signature's "ed25519Key" parameter, if any.
    * `date` be the signature's "date" parameter, interpreted as a Unix time.
```
nyaxt: [Not for this commit] Should this be a http://www.ietf.org/rfc/rfc2616.txt http-date?

draft-yasskin-http-origin-signed-responses.md
```diff
@@ -425,24 +437,21 @@ there are no non-significant response header fields in the exchange.
...
-whole thing before trusting any of it. {{?I-D.thomson-http-mice}} is designed to
-let us check the validity of just the `MI` header up front and then
-incrementally check blocks of the payload as they arrive. What's the best way to
-integrate that? Maybe add a flag to the `Signature` header field or its
-signatures saying that the payload is guarded by some other header field, so
-isn't included in the significant parts ({{significant-parts}}).
-{:#incremental-validity}
+   Note that the above algorithm can determine that an exchange's headers are
+   potentially-valid before the exchange's payload is received. The client MAY
+   process those headers as soon as they are validated. If `integrity`
```
nyaxt: > as soon as they are validated
Would you clarify if this means "potentially-validated" from Section 3.6 algorithm, or "valid" from the full cert chain authority check in Section 4.3.1.

draft-yasskin-http-origin-signed-responses.md
jyasskin: It's possible the HTTPWG will tell me to use text dates, but I'd rather avoid the change until then: text dates are longer, take more code to parse and generate, and we can provide the human readability in devtools if we want.

draft-yasskin-http-origin-signed-responses.md
jyasskin: This turns out to be hard to word. It's actually both: there's a stack of standards here, and the next layer up mustn't process past the potentially-valid point when determining actual validity, and the layer above that mustn't process past the actually-valid point when using the data for some other purpose. It's totally plausible that a protocol might need to see the end of a message body to determine whether any of it is valid, but still take advantage of the incremental validity we're providing here to give that answer faster.

I've updated the wording to try to capture that. I'm going to merge this to master to make my next PR easier to review, but please let me know if I need to change it more.

PR #109, opened 2018-01-19
Specifying the Accept-Signature header to ask for signatures.

jyasskin:
[Preview](https://jyasskin.github.io/webpackage/want-signature/draft-yasskin-http-origin-signed-responses.html), [Diff](https://tools.ietf.org/rfcdiff?url1=https://wicg.github.io/webpackage/draft-yasskin-http-origin-signed-responses.txt&url2=https://jyasskin.github.io/webpackage/want-signature/draft-yasskin-http-origin-signed-responses.txt)

[RFC3230](https://tools.ietf.org/html/rfc3230#section-4.3.1) uses `Want-Digest` for a similar purpose, but `Accept-*` seems more common.

@cramforce, does this look sensible? Have I over- or under-complicated it?

### DIFF
--- a/draft-yasskin-http-origin-signed-responses.md
+++ b/draft-yasskin-http-origin-signed-responses.md
@@ -650,6 +650,110 @@ Signature:
 `https://example.com/resource.validity.1511157180` could also expand the set of
 signatures if its `signatures` array contained more than 2 elements.
 
+## The Accept-Signature header ## {#accept-signature}
+
+`Signature` header fields cost on the order of 300 bytes for ECDSA signatures,
+so servers might prefer to avoid sending them to clients that don't intend to
+use them. A client can send the `Accept-Signature` header field to indicate that
+it does intend to take advantage of any available signatures and to indicate
+what kinds of signatures it supports.
+
+When a server receives an `Accept-Signature` header field in a client request,
+it SHOULD reply with any available `Signature` header fields for its response
+that the `Accept-Signature` header field indicates the client supports. However,
+if the `Accept-Signature` value violates a requirement in this section, the
+server MUST NOT send any `Signature` headers.
+
+The `Accept-Signature` header field is a Structured Header as defined by
+{{!I-D.ietf-httpbis-header-structure}}. Its value MUST be a list (Section 4.8 of
+{{!I-D.ietf-httpbis-header-structure}}) of parameterised labels (Section 4.4 of
+{{!I-D.ietf-httpbis-header-structure}}). The order of labels in the
+`Accept-Signature` list is not significant. Labels, ignoring any initial "-"
+character, MUST NOT be duplicated.
+
+Each label in the `Accept-Signature` header field's value indicates that a
+feature of the `Signature` header field ({{signature-header}}) is supported. If
+the label begins with a "-" character, it instead indicates that the feature
+named by the rest of the label is not supported. Unknown labels and parameters
+MUST be ignored because new labels and new parameters on existing labels may be
+defined by future specifications.
+
+### Integrity labels ### {#accept-signature-integrity}
+
+Labels starting with "digest/" indicate that the client supports the `Digest`
+header field ({{!RFC3230}}) with the digest-algorithm from the
+<https://www.iana.org/assignments/http-dig-alg/http-dig-alg.xhtml> registry
+named in lower-case by the rest of the label. For example, "digest/sha-512"
+indicates support for the SHA-512 digest algorithm, and "-digest/sha-256"
+indicates non-support for the SHA-256 digest algorithm.
+
+Labels starting with "mi/" indicate that the client supports the `MI` header
+field ({{!I-D.thomson-http-mice}}) with the parameter from the HTTP MI Parameter
+Registry registry named in lower-case by the rest of the label. For example,
+"mi/mi-blake2" indicates support for Merkle integrity with the
+as-yet-unspecified mi-blake2 parameter, and "-digest/mi-sha256" indicates
+non-support for Merkle integrity with the mi-sha256 content encoding.
+
+If the `Accept-Signature` header field is present, servers SHOULD assume support
+for "digest/sha-256" and "mi/mi-sha256" unless the header field states
+otherwise.
+
+### Key type labels ### {#accept-signature-key-types}
+
+Labels starting with "rsa/" indicate that the client supports certificates
+holding RSA public keys with a number of bits indicated by the digits after the
+"/".
+
+Labels starting with "ecdsa/" indicate that the client supports certificates
+holding ECDSA public keys on the curve named in lower-case by the rest of the
+label.
+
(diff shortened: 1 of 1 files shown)


draft-yasskin-http-origin-signed-responses.md
```diff
@@ -573,6 +573,110 @@ Signature: sig3;
 https://example.com/resource.validity could also expand the set of signatures if
 its `signatures` array contained more than 2 elements.
 
+## The Accept-Signature header ## {#accept-signature}
+
+`Signature` header fields cost on the order of 300 bytes for ECDSA signatures,
+so servers might prefer to avoid sending them to clients that don't intend to
+use them. A client can send the `Accept-Signature` header field to indicate that
```
cramforce: When would a client send this? Would you expect a UA to always send this header?

jyasskin: I would expect the AMP crawler to always send it. A browser might omit it on normal requests, but send it when there's an [`integrity="ed25519-xyz"` attribute](https://github.com/mikewest/signature-based-sri#the-proposal), and, when the user says "save page as", it might try re-requesting all the resources with the header to see if it can include proofs that the resources aren't modified. Or if cross-signatures become a thing, the browser might always send a version of the header that asks for the cross-signature it cares about.

nyaxt: I'd expect UA implementing pre-fetch mechanisms (incl. Chromium) to always send this header.
LINK rel={prefetch,preload} primitives may prefetch the resources that are later meant to be integrity checked. While re-sending the request on <script integrity> usage is theoretically possible, it would incur additional latency which was the whole point of the prefetch.

draft-yasskin-http-origin-signed-responses.md
```diff
@@ -573,6 +573,110 @@ Signature: sig3;
...
+label.
+
+If the `Accept-Signature` header field is present, servers SHOULD assume support
+for "rsa/2048", "ecdsa/secp256r1", and "ecdsa/secp384r1" unless the header field
+states otherwise.
+
+### Key value labels ### {#accept-signature-key-values}
+
+The "ed25519key" label has parameters indicating the public keys that will be
+used to validate the returned signature. Each parameter's name is re-interpreted
```
nyaxt: Would it be possible for the client (Chromium) to request ed25519key signature response, while not specifying the ed25519 public key (or its prefix)?
I'm not sure if the UA knows which key its going to use at the request time (similar to the prefetch argument above).

draft-yasskin-http-origin-signed-responses.md
jyasskin: We wouldn't need to pass the header for [`as`](https://html.spec.whatwg.org/multipage/semantics.html#attr-link-as) types that don't support [`integrity` attributes](https://w3c.github.io/webappsec-subresource-integrity/#interface-extensions), but yeah, that's a good point for preloaded scripts and stylesheets. @otherdaniel, would it make sense to put `integrity` attributes on prefetch links? [`<link>` already supports the attribute](https://html.spec.whatwg.org/multipage/semantics.html#attr-link-integrity) for the benefit of stylesheets.

Servers can also opt to send a Signature unconditionally when they expect it to be needed by their SRI directives, although we probably don't want to risk that performance footgun.

draft-yasskin-http-origin-signed-responses.md
jyasskin: Right now, no, although we could specify a way to say that. For example, `ed25519key; *`. I *think* that for the https://github.com/mikewest/signature-based-sri use case, the `integrity=ed25519-...` attribute is always available next to the `src`, so the browser could include the public key.
@otherdaniel

draft-yasskin-http-origin-signed-responses.md
theotherdaniel: Regarding _integrity attributes on prefetch links_: We've been [discussing this](https://bugs.chromium.org/p/chromium/issues/detail?id=677022) in a different context, and the consensus then was that from a web developer's perspective it makes little sense to require integrity attributes on prefetches. The present Chrome/Chromium implementation doesn't deal super well with this, but we thought our implementation choices shouldn't be the deciding factor.

I think one can argue this either way, but my gut feeling is that the prefetch shouldn't be burdened with this.

draft-yasskin-http-origin-signed-responses.md
theotherdaniel: According to Mike's proposal, this would always be the case.

I can imagine use cases where this wouldn't be the case. Particularly if [Origin Policy](https://wicg.github.io/origin-policy/) happens, it'd be very convenient if the origin could declare its supported keys, without requiring them to be duplicated within each page. That would then break this.

A related, but somewhat incomplete thought: Mike has been pushing around the idea that maybe each resource should carry its own integrity assertions (that is, the resource declares it's own hash, or maybe signature, etc.), which would make it somewhat self-contained. Then, much of the work could happen as each resource is downloaded, independently of where/how it's being used. That's a somewhat incomplete plan, but IMHO rather attractive since it offers a path to bring the cost of (additional) security down to near-zero, and predictably so. Much of what we discuss here isn't directly contradicting this, but does go in the opposite direction, in that a resource could only be integrity-checked if requested in specific ways.

draft-yasskin-http-origin-signed-responses.md
jyasskin: 'k, I realized that `ed25519key; *` already matches all keys, so the browser can use that on prefetches if we don't want those to include `integrity` attributes. I've added an example for this. Or servers can send their `Signature` headers unconditionally for those resources.

draft-yasskin-http-origin-signed-responses.md
jyasskin: Yep, I wouldn't want to get in the way of putting keys in the Origin Policy. Servers can definitely send the `Signature` even if the browser didn't explicitly request it, which fits the second half. If that becomes policy, we can drop the `ed25519key; ...` stuff entirely.

jyasskin:
@yoavweiss Here's your content negotiation mechanism.

yoavweiss:
> @yoavweiss Here's your content negotiation mechanism.

Indeed! missed that PR. Will go over it soonish

jyasskin:
I'm merging this to unblock publishing the next IETF draft, but please keep commenting and filing issues as you notice problems.

draft-yasskin-http-origin-signed-responses.md
theotherdaniel: This is a bit of a nitpick, but: As I read draft-ietf-httpbis-header-structure-02, a lone asterisk wouldn't match the grammar given there. It specifies 

>  binary = "*" 1\*21846(base64)

which I read as at least 1 and at most 21846 base64 characters. I guess there should be some way to specifiy an empty binary string, but I can't figure out how.

PR #110, opened 2018-01-20
Define application/http-exchange+cbor.

jyasskin:
[Preview](https://jyasskin.github.io/webpackage/envelope-format/draft-yasskin-http-origin-signed-responses.html#application-http-exchange)

This format wraps an HTTP exchange, which allows an HTTP/1 connection to
transfer cross-origin signed exchanges, and which may have other uses.

@kinu @yoavweiss Does this look reasonable?

### DIFF
--- a/draft-yasskin-http-origin-signed-responses.md
+++ b/draft-yasskin-http-origin-signed-responses.md
@@ -1146,7 +1146,7 @@ Additional information:
 
   Magic number(s):  8? 64 68 74 78 67
 
-  File extension(s):  N/A
+  File extension(s): .htxg
 
   Macintosh file type code(s):  N/A
 
--- a/draft-yasskin-http-origin-signed-responses.md
+++ b/draft-yasskin-http-origin-signed-responses.md
@@ -976,6 +976,13 @@ defined in Appendix G of {{?I-D.ietf-cbor-cddl}}:
 ]
 ~~~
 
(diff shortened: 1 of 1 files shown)


draft-yasskin-http-origin-signed-responses.md
```diff
@@ -764,6 +875,50 @@ signed exchange.
...
+
+Fragment identifier considerations:  N/A
+
+Additional information:
+
+  Deprecated alias names for this type:  N/A
+
+  Magic number(s):  8? 64 68 74 78 67
+
+  File extension(s):  N/A
```
nyaxt: Would it be possible to standardize one?

kinu:
Just read through the texts, LGTM

/cc @horo-t

draft-yasskin-http-origin-signed-responses.md
kinu: Reg: impl side concerns I think we're also fine with leaving it as N/A

jyasskin: I'll grab `htxg` which matches the magic number that I pulled out of my hat, but I'm going to invite the HTTPWG to propose other names for the mime type, which could easily change the extension.

jyasskin:
I'm merging this to unblock publishing the next IETF draft, but please keep commenting and filing issues as you notice problems.

PR #111, opened 2018-01-23
Changes inspired by Ryan's comments

jyasskin:
@sleevi, how's this look?

[Preview](https://jyasskin.github.io/webpackage/sleevi-comments/draft-yasskin-http-origin-signed-responses.html), [Diff](https://tools.ietf.org/rfcdiff?url1=https://wicg.github.io/webpackage/draft-yasskin-http-origin-signed-responses.txt&url2=https://jyasskin.github.io/webpackage/sleevi-comments/draft-yasskin-http-origin-signed-responses.txt)

### DIFF
--- a/draft-yasskin-http-origin-signed-responses.md
+++ b/draft-yasskin-http-origin-signed-responses.md
@@ -798,6 +798,14 @@ this leaks extra information.
 For non-executable resource types, a signed response can improve the privacy
 situation by hiding the client's interest from the original author.
 
+To prevent network operators other than `o1.com` or `o2.com` from learning which
+exchanges were read, clients SHOULD only load exchanges fetched over a transport
+that's protected from eavesdroppers. This can be difficult to determine when the
+exchange is being loaded from local disk, but when the client itself requested
+the exchange over a network it SHOULD require TLS ({{!I-D.ietf-tls-tls13}}) or a
+successor transport layer, and MUST NOT accept exchanges transferred over plain
+HTTP without TLS.
+
 # IANA considerations
 
 TODO: possibly register the validityUrl format.
--- a/draft-yasskin-http-origin-signed-responses.md
+++ b/draft-yasskin-http-origin-signed-responses.md
@@ -23,6 +23,12 @@ normative:
     author:
       org: WHATWG
     date: Living Standard
+  HTML:
+    target: https://html.spec.whatwg.org/multipage
+    title: HTML
+    author:
+      org: WHATWG
+    date: Living Standard
   POSIX:
     target: http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/
     title: The Open Group Base Specifications Issue 7
@@ -200,8 +206,8 @@ present parameters MUST have the following values:
 : An unsigned integer (Section 4.1 of {{!I-D.ietf-httpbis-header-structure}})
   representing a Unix time.
 
-The "certUrl" and "validityUrl" parameters are *not* signed, so intermediates can
-update them with pointers to cached versions.
+The "certUrl" parameter is *not* signed, so intermediates can update it with a
+pointer to a cached version.
 
 ### Examples ### {#example-signature-header}
 
@@ -280,10 +286,6 @@ Should the certUrl and validityUrl be lists so that intermediates can offer a
 cache without losing the original URLs? Putting lists in dictionary fields is
 more complex than {{?I-D.ietf-httpbis-header-structure}} allows, so they're
 single items for now.
-
-Should "validityUrl" be signed or optionally signed so that an exchange's author
-can prevent an intermediate from removing it, which would prevent clients from
-sharing the exchange among themselves without going back to the intermeidate?
 
 ## Significant headers of an exchange ## {#significant-headers}
 
@@ -428,6 +430,7 @@ there are no non-significant response header fields in the exchange.
    * `signature` be the signature (binary content in the parameterised value's
      "sig" parameter).
    * `integrity` be the signature's "integrity" parameter.
+   * `validityUrl` be the signature's "validityUrl" parameter.
    * `certUrl` be the signature's "certUrl" parameter, if any.
    * `certSha256` be the signature's "certSha256" parameter, if any.
    * `ed25519Key` be the signature's "ed25519Key" parameter, if any.
@@ -499,7 +502,10 @@ there are no non-significant response header fields in the exchange.
    1. The bytes of the canonical CBOR serialization ({{canonical-cbor}}) of a
       CBOR map mapping:
       1. If `certSha256` is set:
-         1. The text string "certSha256" to the byte string `certSha256`.
+         1. The text string "certSha256" to the byte string value of
+            `certSha256`.
+      1. The text string "validityUrl" to the byte string value of
+         `validityUrl`.
       1. The text string "date" to the integer value of `date`.
       1. The text string "expires" to the integer value of `expires`.
       1. The text string "headers" to the CBOR representation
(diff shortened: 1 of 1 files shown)


draft-yasskin-http-origin-signed-responses.md
```diff
@@ -655,33 +665,52 @@ authoritative for the PUSH_PROMISE's authority.
...
+
+## Downgrades ## {#seccons-downgrades}
 
 Signing a bad response can affect more users than simply serving a bad response,
 since a served response will only affect users who make a request while the bad
 version is live, while an attacker can forward a signed response until its
 signature expires. Authors should consider shorter signature expiration times
 than they use for cache expiration times.
 
+Clients MAY also check the ["validityUrl"](#signature-validityurl) of an
```
jyasskin: @jschuh, here's your suggestion that clients can check for validity more often.

draft-yasskin-http-origin-signed-responses.md
```diff
@@ -481,97 +487,103 @@ following CDDL ({{!I-D.ietf-cbor-cddl}}):
...
   certUrl="https://example.com/certs";
   certSha256=*kQAA8u33cZRTy7RHMO4+dv57baZL48SYA2PqmYvPPbg;
-  date=1511301183; expires=1511905983
-Signature: sig3;
+  date=1511301183; expires=1511905983,
+  sig3;
   sig=*MEYCIQCNxJzn6Rh2fNxsobktir8TkiaJYQFhWTuWI1i4PewQaQIhAMs2TVjc4rTshDtXbgQEOwgj2mRXALhfXPztXgPupii+;
   integrity="mi";
-  validityUrl="https://thirdparty.example.com/resource.validity";
+  validityUrl="https://thirdparty.example.com/resource.validity.1511128380";
```
sleevi: Readability wise, this is very subtle and tripped me up during review here, that we're talking thirdparty.example.com versus example.com

It might make sense to use a different suffix here on the resource.validity, so that the visual pattern matching can more easily notice the difference in origins?

draft-yasskin-http-origin-signed-responses.md
```diff
@@ -481,97 +487,103 @@ following CDDL ({{!I-D.ietf-cbor-cddl}}):
...
+  sig2;
   sig=*MEQCIGjZRqTRf9iKNkGFyzRMTFgwf/BrY2ZNIP/dykhUV0aYAiBTXg+8wujoT4n/W+cNgb7pGqQvIUGYZ8u8HZJ5YH26Qg;
   integrity="mi";
-  validityUrl="https://example.com/resource.validity";
+  validityUrl="https://example.com/resource.validity.1511128380";
   certUrl="https://example.com/certs";
   certSha256=*kQAA8u33cZRTy7RHMO4+dv57baZL48SYA2PqmYvPPbg;
-  date=1511301183; expires=1511905983
-Signature: sig3;
+  date=1511301183; expires=1511905983,
```
sleevi: As a further comment re: examples, it's unclear what the implications of the overlap are between the expires on sig1 and the date on sig2. That is, it's not clear the use case for the overlapping set of signatures with the same validityURL, versus discrete (non-overlapping)

This is not specific to this PR, but was within the set of concerns with the examples being relatively opaque. It may be that an 'explainer' type document is better suited to capture this nuance, but as an example, it's a bit confusing :)

draft-yasskin-http-origin-signed-responses.md
```diff
@@ -655,33 +671,52 @@ authoritative for the PUSH_PROMISE's authority.
...
 
 Signing a bad response can affect more users than simply serving a bad response,
 since a served response will only affect users who make a request while the bad
 version is live, while an attacker can forward a signed response until its
 signature expires. Authors should consider shorter signature expiration times
 than they use for cache expiration times.
 
+Clients MAY also check the ["validityUrl"](#signature-validityurl) of an
+exchange more often than the signature's expiration would require. Doing so
+provides a TLS guarantee that the exchange isn't out of date.
```
sleevi: This guarantee is provided iff you have the same-origin assertion from line 647. The open question (in line 674) makes it unclear whether this would be true, but if that's removed, then it could be that validityUrl is served by a different entity, and thus does not provide a TLS guarantee.

sleevi: Also, given validityUrls are unsigned and don't *seem* to be restricted to https:// in any of the normative requirements, I'm not sure that is inherently true, so much as assumed true?

draft-yasskin-http-origin-signed-responses.md
```diff
@@ -718,6 +753,10 @@ this leaks extra information.
 For non-executable resource types, a signed response can improve the privacy
 situation by hiding the client's interest from the original author.
 
+To prevent network operators other than `o1.com` or `o2.com` from learning which
+packages were read, clients SHOULD only load packages fetched over a transport
+at least as secure as TLS ({{?I-D.ietf-tls-tls13}}).
```
sleevi: 1) Why not MUST?
2) How do you determine "as secure as". What is the guidance meant to be provided to implementations in determining that assessment? Is HTTP on a local network 'at least as secure as' HTTPS, if it's over a VPN?

draft-yasskin-http-origin-signed-responses.md
```diff
@@ -516,37 +516,38 @@ reading):
...
   sig=*MEQCIGjZRqTRf9iKNkGFyzRMTFgwf/BrY2ZNIP/dykhUV0aYAiBTXg+8wujoT4n/W+cNgb7pGqQvIUGYZ8u8HZJ5YH26Qg;
   integrity="mi";
-  validityUrl="https://example.com/resource.validity.1511128380";
   certUrl="https://example.com/certs";
-  certSha256=*kQAA8u33cZRTy7RHMO4+dv57baZL48SYA2PqmYvPPbg;
+  certSha256=*W7uB969dFW3Mb5ZefPS9Tq5ZbH5iSmOILpjv2qEArmI;
   date=1511301183; expires=1511905983,
-  sig3;
+ sig3;
+  validityUrl="https://THIRDPARTY.com/resource.validity.1511128380";
```
sleevi: Ahhh, I don't think we'd want to use anything other than the IANA-reserved domains, especially in a draft, so I think you were doing great with thirdparty.example.com, which is why I mentioned the suffix part :)

Put more precise: I do not think we should deviate from https://tools.ietf.org/html/rfc6761#section-6.5

jyasskin: Thanks for expressing that opinion. Fixed.

draft-yasskin-http-origin-signed-responses.md
```diff
@@ -200,8 +206,70 @@ present parameters MUST have the following values:
...
+  validityUrl="https://thirdparty.example.com/resource.validity.1511161860";
+  certUrl="https://thirdparty.example.com/certs";
+  certSha256=*UeOwUPkvxlGRTyvHcsMUN0A2oNsZbU8EUvg8A9ZAnNc;
+  date=1511133060; expires=1511478660,
+~~~
+
+There are 4 signatures: 2 from different secp256r1 certificates within
+`https://example.com/`, one using a raw ed25519 public key that's also
+controlled by `example.com`, and a fourth using a secp256r1 cross-signing
+certificate owned by `thirdparty.example.com`.
```
sleevi: Could you explain what you mean by 'cross-signing certificate' here?

That is, 'normally' cross-signing refers to the same subject entity being validated by two different issuers - e.g., a given intermediate "Intermediate Foo" can be signed by both "CA A" and "CA B".

This makes me think that the use case here is `thirdparty.example.com` asserting that the content is truly from `example.com`, rather than being `example.com` proving that the `thirdpartysig` covers the resources from `thirdparty.example.com`.

If it's the latter (`thirdparty.example.com` certifying the content on `example.com`), then it seems we're introducing yet another Trusted Third Party into Web Packaging, beyond the CA model. That seems a radical departure from the TLS guarantees, and perhaps a subtle aspect I failed to realize - that this would allow non-CA entities to certify certificates.

jyasskin: I mean for all four of these signatures to sign the resource, not any other certificates. (Remember, this is currently about a single exchange, not for a bundle of exchanges that might contain resources from several origins.) One use case might be for `thirdparty.example.com` to be a binary transparency log whose signature certifies that the resource is contained in the log. I'm adding a more explicit statement of the cross-signing use cases in https://github.com/WICG/webpackage/pull/112/files#diff-a2197f2cbc199dc1463dc10213fa40d9R862.

I'm going to remove the use of "cross-signing" lest it imply signing certificates to anyone else, and then merge so that I can publish a new I-D.

PR #122, opened 2018-02-09
Avoid the TLS 1.3 Certificate message.

jyasskin:
Use a custom CBOR certificate chain format instead.

Also register this format with IANA.

@sleevi, better?

[Preview](https://jyasskin.github.io/webpackage/avoid-tls-messages/draft-yasskin-http-origin-signed-responses.html), [Diff](https://tools.ietf.org/rfcdiff?url1=https://wicg.github.io/webpackage/draft-yasskin-http-origin-signed-responses.txt&url2=https://jyasskin.github.io/webpackage/avoid-tls-messages/draft-yasskin-http-origin-signed-responses.txt)

### DIFF
--- a/draft-yasskin-http-origin-signed-responses.md
+++ b/draft-yasskin-http-origin-signed-responses.md
@@ -386,6 +386,58 @@ extended diagnostic notation from {{?I-D.ietf-cbor-cddl}} appendix G:
 ]
 ~~~
 
+## Loading a certificate chain ## {#cert-chain}
+
+The resource at a signature's `certUrl` MUST have the
+`application/cert-chain+cbor` content type, MUST be canonically-encoded CBOR
+({{canonical-cbor}}), and MUST match the following CDDL:
+
+~~~cddl
+cert-chain = [
+  "📜⛓", ; U+1F4DC U+26D3
+  + {
+    cert: bytes,
+    ? ocsp: bytes,
+    ? sct: bytes,
+    * tstr => any,
+  }
+]
+~~~
+
+The first item in the CBOR array is treated as the end-entity certificate, and
+the client will attempt to build a path ({{?RFC5280}}) to it from a trusted root
+using the other certificates in the chain.
+
+1. Each `cert` value MUST be a DER-encoded X.509v3 certificate ({{!RFC5280}}).
+   Other key/value pairs in the same array item define properties of this
+   certificate.
+1. The first certificate's `ocsp` value if any MUST be a complete, DER-encoded
+   OCSP response for that certificate (using the ASN.1 type `OCSPResponse`
+   defined in {{!RFC2560}}). Subsequent certificates MUST NOT have an `ocsp`
+   value.
+1. Each certificate's `sct` value MUST be a `SignedCertificateTimestampList` for
+   that certificate as defined by Section 3.3 of {{!RFC6962}}.
+
+Loading a `certUrl` takes a `forceFetch` flag and an `expectedSha256` hash
+value. The client MUST:
+
+1. Let `raw-chain` be the result of fetching ({{FETCH}}) `certUrl`. If
+   `forceFetch` is *not* set, the fetch can be fulfilled from a cache using
+   normal HTTP semantics {{!RFC7234}}. If this fetch fails, return
+   "invalid".
+1. Let `certificate-chain` be the array of certificates and properties produced
+   by parsing `raw-chain` using the CDDL above. If any of the requirements above
+   aren't satisfied, return "invalid". Note that this validation requirement
+   might be impractical to completely achieve due to certificate validation
+   implementations that don't enforce DER encoding or other standard
+   constraints.
+1. If the SHA-256 hash of `main-certificate`'s `cert_data` is not equal to
+   `certSha256`, return "invalid". Note that this intentionally differs from TLS
+   1.3, which signs the entire certificate chain in its Certificate Verify
+   (Section 4.4.3 of {{?I-D.ietf-tls-tls13}}), in order to allow updating the
+   stapled OCSP response without updating signatures at the same time.
+1. Return `certificate-chain`.
+
 ## Canonical CBOR serialization ## {#canonical-cbor}
 
 Within this specification, the canonical serialization of a CBOR item uses the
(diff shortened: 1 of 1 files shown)


sleevi:
@davidben should also look at this.

David: The concern was that at present, no TLS library exposes its Certificate parsing or generating functions (Modulo Boring’s CBS), and that, more generally speaking, the TLS libraries are maintained at separate cadences and release structures than UAs.

There’s also the separable concern about semantics of or to the TLS layer and their interaction with Web Packaging, and thus the suggestion to make it an explicit format.

draft-yasskin-http-origin-signed-responses.md
```diff
@@ -881,14 +917,13 @@ authoritative for the PUSH_PROMISE's authority.
...
-   1. Validate that `main-certificate` includes a "status_request" extension
+   1. Validate that `main-certificate` has an `ocsp` property ({{cert-chain}})
       with a valid OCSP response whose lifetime (`nextUpdate - thisUpdate`) is
       less than 7 days ({{!RFC6960}}). Note that this does not check for
       revocation of intermediate certificates, and clients SHOULD implement
       another mechanism for that.
-   1. Validate that all certificates in `path` include
-      "signed_certificate_timestamp" extensions containing valid SCTs from
-      trusted logs. ({{!RFC6962}})
+   1. Validate that all certificates in `path` include `sct` properties
```
jyasskin: Is this right, or should it only be the leaf certificate that holds an SCT?

draft-yasskin-http-origin-signed-responses.md
```diff
@@ -386,6 +386,58 @@ extended diagnostic notation from {{?I-D.ietf-cbor-cddl}} appendix G:
...
+The resource at a signature's `certUrl` MUST have the
+`application/cert-chain+cbor` content type, MUST be canonically-encoded CBOR
+({{canonical-cbor}}), and MUST match the following CDDL:
+
+~~~cddl
+cert-chain = [
+  "📜⛓", ; U+1F4DC U+26D3
+  + {
+    cert: bytes,
+    ? ocsp: bytes,
```
jyasskin: Oh, also, this is a simplification of the `status_request` TLS extension that omits the layer containing a `CertificateStatusType`. Is that the right choice?

draft-yasskin-http-origin-signed-responses.md
```diff
@@ -386,6 +386,58 @@ extended diagnostic notation from {{?I-D.ietf-cbor-cddl}} appendix G:
...
+    ? sct: bytes,
```
jyasskin: A `SignedCertificateTimestampList` appears to be a serialized TLS message, but I guess because other systems have to generate this, it's ok?

davidben:
Haven't looked closely at the PR text yet, but I agree avoiding the TLS message makes sense. That was one of the comments I had in my queue. :-) Even with BoringSSL's `CBS` stuff, we don't expose the message itself, just the low-level length-prefix bits.

You needed different parsing rules from TLS anyway. TLS doesn't allow unsolicited extensions, so you actually reject, not ignore, unknown extensions. (SCTs, OCSP stapling, etc., are advertised in ClientHello or CertificateRequest, so the other side knows whether to include them.) Your signing rules are also slightly different, with the `certSha256` stuff.

draft-yasskin-http-origin-signed-responses.md
sleevi: Under RFC 6962, only the leaf has (additional) SCTs provided by the TLS extension

Under 6962-bis, this can change, but there are no implementations of 6962-bis (nor announced) that I know of.

draft-yasskin-http-origin-signed-responses.md
sleevi: I believe so, at least as we discussed. This might be an area of future change.

The practical reality is that OCSP multi-stapling is rather inefficient for the browser case, given that "Every Modern Browser" (tm) effectively has an out-of-bands mechanism for dealing with intermediate revocation, treating it akin to root revocation, as well as heavily caching intermediate results (e.g. in the case of macOS, via an out-of-band datafile containing the revocation status of all intermediates, in the case of Windows, through background-fetching CRLs that cover CAs)

From a PKI pursist standpoint, however, you could try and argue that such systems don't handle nested sets of name-constrained subordinate CAs, and for those cases (and, effectively, those cases alone), it makes sense to allow OCSP responses for intermediates to be stapled.

draft-yasskin-http-origin-signed-responses.md
sleevi: Right, this is passed as (effectively) opaque input, much like the OCSP response, into the certificate validation system, which contains a minimal subset of the TLS decoder for this case.

draft-yasskin-http-origin-signed-responses.md
jyasskin: I've changed it to just the leaf certificate. Thanks.

draft-yasskin-http-origin-signed-responses.md
jyasskin: I'm going to omit the "nested sets of name-constrained subordinate CAs" case from this draft, and if someone tells me to add it, I'm going to make them write the wording that distinguishes certificates that need OCSPs from ones that don't.

draft-yasskin-http-origin-signed-responses.md
```diff
@@ -530,10 +560,17 @@ there are no non-significant response header fields in the exchange.
...
+   `cert_data` is not equal to `certSha256` (whose presence was checked when the
+   `Signature` header field was parsed), return "invalid".
+
+   Note that this intentionally differs from TLS 1.3, which signs the entire
+   certificate chain in its Certificate Verify (Section 4.4.3 of
+   {{?I-D.ietf-tls-tls13}}), in order to allow updating the stapled OCSP
+   response without updating signatures at the same time.
+1. If `signature` is `message`'s signature by `publicKey` using `signing-alg`,
+   return "potentially-valid" with `exchange` and whichever is present of
+   `certificate-chain` or `ed25519Key`. Otherwise, return "invalid".
```
davidben: Nit: There can be more than one valid signature for a given public key and message. I would say perhaps:

If `signature` is a valid signature of `message` by `publicKey` using `signing-alg` [...]

draft-yasskin-http-origin-signed-responses.md
```diff
@@ -386,6 +386,52 @@ extended diagnostic notation from {{?I-D.ietf-cbor-cddl}} appendix G:
...
+the client will attempt to build a path ({{?RFC5280}}) to it from a trusted root
+using the other certificates in the chain.
+
+1. Each `cert` value MUST be a DER-encoded X.509v3 certificate ({{!RFC5280}}).
+   Other key/value pairs in the same array item define properties of this
+   certificate.
+1. The first certificate's `ocsp` value if any MUST be a complete, DER-encoded
+   OCSP response for that certificate (using the ASN.1 type `OCSPResponse`
+   defined in {{!RFC2560}}). Subsequent certificates MUST NOT have an `ocsp`
+   value.
```
davidben: Re "subsequent certificates", I suspect this may end up being a bit controversial since the TLSWG intentionally went the other way with TLS 1.3 and made it easy to encode OCSP staples on intermediates. But I'll leave that for @sleevi and others to wrestle out later. :-)

(In our TLS 1.3 implementation, we tolerate metadata on intermediates but ignore them for now. We'll start processing them if we have something useful to do with them.)

draft-yasskin-http-origin-signed-responses.md
```diff
@@ -1568,6 +1648,12 @@ exchange argues for embedding a signature's lifetime into the signature.
 
 RFC EDITOR PLEASE DELETE THIS SECTION.
 
+draft-03
+
+* Define a CBOR structure to hold the certificate chain instead of re-using the
+  TLS1.3 message. Apparently TLS implementations don't expose their message
+  parsers enough to allow passing a message to a certificate verifier.
```
davidben: (Also the TLS 1.3 message has different parsing rules than what you want anyway.)

Issue #101, opened 2018-01-09
will support packaging only subresource ?


Jxck:
in ES modules, we need to fetch a lot of module files.
increasing requests cause a performance issue. ([10k 1kB load is 10 times slower than single 10MB load](https://docs.google.com/document/d/1ds9EEkcDGnt-iR8SAN-_7nsOfw7gsMfhZjzZ_QAIyjM/edit))

h2 will reduce connection handshake/establish, but doesn't reduce request/response itself which browser handles inside.

but reducing files via bundling like webpack is less merit of es modules(ex caching etc). 

same things happens for image. so we sometimes concatting them aka Sprite.

current spec seems packaging main-resource + sub-resource in 1 file.
so if webpackage support or have candidate to package lot of resources into 1 package(not include main-resoruce) seems solve fetching performance problem.
expected behavior is fetching single response but unpackaging in browser and handling/caching like separated ESModules/ImageIcons/StyleSheets etc.

yoavweiss:
This is a use case I discussed with @jyasskin, @addyosmani and @patrickkettner.

I think using webpackage as a bundling format is an interesting use case, and will provide the same compression benefits bundling currently provides without all its downsides. I think it can and should be addressed by these specifications, but will require some extra mechanisms to enable it.

I guess the first step is to send out a PR outlining this use-case.

yoavweiss:
Filed a use case PR. Beyond that, we also need to define a mechanism that will let the browser:
* Know that a package download is required
* Know which resources are covered by a certain package, so that their requests won't race with the package's response. Such a mechanism can use some form of digest that will include all the URLs in the package and will be delivered with the "package is required" markup.

Straw-man for such markup may look like:
`<link rel=package href="js_bundle" digest="bcfd23234989734abdcef">`

Jxck:
It's not only for esmodules but also image, style, font, favicon etc.

I have 2 plan for this in my mind.

dump them in rough.

## `<package>` like `<picture>` way for selective `<img>`

for example

```
<package src=bundle.wpac>
  <!-- backword compat for no-package -->
  <script type=module src=react.js>
  <script type=module src=moment.js>
  <script type=module src=main.js>

  <!-- backword compat for no-module -->
  <script type=nomodule src=bundle.js>

  <!-- css -->
  <link rel=stylesheet href=bootstrap.css>
  <link rel=stylesheet href=style.css>

  <!-- favicon -->
  <link rel="shortcut icon" href="favicon.ico">
</package>
[...]

PR #123, opened 2018-02-12
Add a use-case for subresource bundling

yoavweiss:
Closes https://github.com/WICG/webpackage/issues/101


yoavweiss:
Your update LGTM

Issue #124, opened 2018-02-20
gen-signedexchange should be able to capture arbitrary request/response HTTP headers.


nyaxt:
Originally requested by @kinu.

We should be able to set arbitrary request/response HTTP headers to be captured in the generated signed exchange.
```
$ ./gen-signedexchange \
    -requestHeader "Accept-Language: en" \
    -responseHeader "Server: Nginx" \
    -responseHeader "X-Frame-Options: Deny"
```

@hajimehoshi Would you work on this?

hajimehoshi:
Looking

PR #125, opened 2018-02-21
signedexchange: Add flags 'requestHeader' and 'responseHeader'

hajimehoshi:
#124 

This is WIP: I have no idea how to treat the passed request headers.


jyasskin:
Right now, the spec says not to sign request headers, which would imply that the tool shouldn't actually take a `--requestHeader` argument. If we want to sign request headers (which is plausible, especially inside `application/http-exchange+cbor` objects instead of plain HTTP responses), I think the natural thing is to include them in the map created by `encodeRequest()`.

hajimehoshi:
> I think the natural thing is to include them in the map created by encodeRequest().

Thank you, done.

go/signedexchange/signedexchange.go
```diff
@@ -97,6 +99,13 @@ func (e *Exchange) encodeRequest(enc *cbor.Encoder) error {
 			valueE.EncodeByteString([]byte(e.requestUri.String()))
 		}),
 	}
+	for name, value := range e.requestHeaders {
```
nyaxt: encodeRequest from L172 should include requestHeaders, while L149 should not

PR #132, opened 2018-03-02
Remove allResponseHeaders; handle signed request headers.

jyasskin:
@davidben, you'll probably like this one. It refactors the rules for trusting cross-origin exchanges so that they can be shared by Push and the application/http-exchange format. This let me cover all headers from those two transfer methods unconditionally, with only same-origin exchanges able to use `Signed-Headers` to select a subset.

There's a bunch of text motion, but sections id'ed `signed-headers` and `significant-headers` are basically unchanged in their new locations, and `authority-chain-validation` became the algorithm in `cross-origin-trust`.

[Preview](https://jyasskin.github.io/webpackage/unify-cross-origin/draft-yasskin-http-origin-signed-responses.html), [Diff](https://tools.ietf.org/rfcdiff?url1=https://wicg.github.io/webpackage/draft-yasskin-http-origin-signed-responses.txt&url2=https://jyasskin.github.io/webpackage/unify-cross-origin/draft-yasskin-http-origin-signed-responses.txt)

### DIFF
--- a/draft-yasskin-http-origin-signed-responses.md
+++ b/draft-yasskin-http-origin-signed-responses.md
@@ -122,15 +122,15 @@ appear in all capitals, as shown here.
 
 # Signing an exchange # {#proposal}
 
-As a response to an HTTP request or as a Server Push (Section 8.2 of
-{{!RFC7540}}) the server MAY include a `Signed-Headers` header field
-({{signed-headers}}) identifying [significant](#significant-headers) header
-fields and a `Signature` header field ({{signature-header}}) holding a list of
-one or more parameterised signatures that vouch for the content of the response.
+In the response of an HTTP exchange the server MAY include a `Signature` header
+field ({{signature-header}}) holding a list of one or more parameterised
+signatures that vouch for the content of the exchange. Exactly which content the
+signature vouches for can depend on how the exchange is transferred
+({{transfer}}).
 
 The client categorizes each signature as "valid" or "invalid" by validating that
 signature with its certificate or public key and other metadata against the
-significant headers and content ({{signature-validity}}). This validity then
+exchange's headers and content ({{signature-validity}}). This validity then
 informs higher-level protocols.
 
 Each signature is parameterised with information to let a client fetch assurance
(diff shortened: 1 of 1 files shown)


draft-yasskin-http-origin-signed-responses.md
```diff
@@ -893,51 +942,20 @@ error (Section 5.4.2 of {{!RFC7540}}) of type PROTOCOL_ERROR, as described in
 Section 8.2 of {{?RFC7540}}.
 
 Instead, the client MUST validate such a PUSH_PROMISE and its response by
-parsing the `Signature` header into a list of signatures according to the
-instructions in {{signature-validity}}, and searching that list for a valid
-signature using the algorithm in {{authority-chain-validation}}. If no valid
-signature is found, the client MUST treat the response as a stream error
+passing the exchange they represent to the algorithm in {{cross-origin-trust}}.
```
jyasskin: @nyaxt pointed out that this needs to exclude the `Signature` header from the exchange passed to {{cross-origin-trust}} (since you can't cover a signature with itself), so I'm working on that.

draft-yasskin-http-origin-signed-responses.md
```diff
@@ -1112,9 +1130,9 @@ Sleevi's recommendation.
 Clients MUST NOT trust an effective request URI claimed by an
 `application/http-exchange+cbor` resource ({{application-http-exchange}})
 without either ensuring the resource was transferred from a server that was
-authoritative (Section 9.1 of {{!RFC7230}}) for that URI's origin, or validating
-the resource's signature using a procedure like the one described in
-{{authority-chain-validation}}.
+authoritative (Section 9.1 of {{!RFC7230}}) for that URI's origin, or passing
+the exchange stored in the resource to the procedure in {{cross-origin-trust}},
```
jyasskin: I'm excluding the `Signature` header from this too.

draft-yasskin-http-origin-signed-responses.md
jyasskin: Ok, fixed, I think.

PR #134, opened 2018-03-03
Require a new certificate extension for packages

jyasskin:
This builds on #132, so probably review that one first.

[Preview](https://jyasskin.github.io/webpackage/cert-extension/draft-yasskin-http-origin-signed-responses.html), [Diff](https://tools.ietf.org/rfcdiff?url1=https://wicg.github.io/webpackage/draft-yasskin-http-origin-signed-responses.txt&url2=https://jyasskin.github.io/webpackage/cert-extension/draft-yasskin-http-origin-signed-responses.txt)

@davidben @sleevi

### DIFF
--- a/draft-yasskin-http-origin-signed-responses.md
+++ b/draft-yasskin-http-origin-signed-responses.md
@@ -41,6 +41,34 @@ normative:
     date: 2016
 
 informative:
+  DROWN:
+    target: https://drownattack.com/
+    title: The DROWN Attack
+    author:
+      - name: Nimrod Aviram
+      - name: Sebastian Schinzel
+      - name: Juraj Somorovsky
+      - name: Nadia Heninger
+      - name: Maik Dankel
+      - name: Jens Steube
+      - name: Luke Valenta
+      - name: David Adrian
+      - name: J. Alex Halderman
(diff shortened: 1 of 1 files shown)


draft-yasskin-http-origin-signed-responses.md
```diff
@@ -484,10 +512,7 @@ to retrieve an updated OCSP from the original server.
 1. If `expires` is more than 7 days (604800 seconds) after `date`, return
    "invalid".
 1. If the current time is before `date` or after `expires`, return "invalid".
-1. Let `message` be the concatenation of the following byte strings. This
-   matches the {{?I-D.ietf-tls-tls13}} format to avoid cross-protocol attacks
-   when TLS certificates are used to sign manifests.
-   1. A string that consists of octet 32 (0x20) repeated 64 times.
+1. Let `message` be the concatenation of the following byte strings:
```
jyasskin: Note that because these certificates can't be used for TLS servers anymore, I'm diverging from the TLS format. IIUC, the initial 64 spaces are only needed in TLS because the attacker controls the initial content of the message, but since the attacker doesn't control that for packages, it doesn't seem to be needed here. I've added an open question to check that my guess is right.

Issue #105, opened 2018-01-09
Archived / immutable content


craigfrancis:
While some packages need to be kept up to date, and expired when they become too old; there are also things that should never expire/change.

For example the content of a webpage a user wants to keep, a report or invoice created by a website, etc (this is instead of using a PDF or MS Word document).

So would it be possible to skip the Expires header, and have some way of marking the package as immutable?

That immutable flag could also instruct the browser to never download any resources, and for the JS `Date` object to return a date/time based on the Date header from the package (not the current date/time)... may be related to issue #104?

jyasskin:
"Archival" makes me think of formats like [WARC](http://iipc.github.io/warc-specifications/specifications/warc-format/warc-1.1/), which we're not trying to compete with: they focus on recording the state of the web, while we're trying to distribute content to browsers.

However, the rest of your comment looks like your use case is actually around saving a webpage for later personal use. Most of the need for expiration centers around preventing attackers from forging signatures or exploiting bugs, which isn't a risk for sites you've saved personally and don't distribute. We will support un-signed packages for this sort of thing, and there won't be anything forcing them to expire. Does that work for you?

bmeck:
The [`cache-control: immutable` header](https://tools.ietf.org/html/rfc8246) also seems to follow that line of thought where it does have an expiration associated with it, even if it is immutable. I think saving from signed->unsigned seems a reasonable utility to have for most tools.

TzviyaSiegman:
@prototypo Please see overlap with Archiving section (to come) in https://www.w3.org/TR/pwpub/

ibnesayeed:
The web archiving community is [taking notice of it](https://twitter.com/anjacks0n/status/950861384266416134). While WARC works well for its purpose, it would be good to give some more thoughts on the archiving aspect, be it in the form of some possibility of transformation between webpackage and WARC or some other way to interoperate. WARC might want to store the whole package as an object or unpack it to store individual resources.

rektide:
@craigfrancis according to https://wicg.github.io/webpackage/draft-yasskin-webpackage-use-cases.html#rfc.section.4.1 ,

> Packages are designed to hold public information and to be shared to people with whom the original author never has an interactive connection. In that situation, there’s no way to keep the contents confidential: even if they were encrypted, to make the data public, anyone would have to be able to get the decryption key.

[...]

PR #137, opened 2018-03-05
Add an archival use case.

jyasskin:
[Preview](https://jyasskin.github.io/webpackage/archival-use-case/draft-yasskin-webpackage-use-cases.html), [Diff](https://tools.ietf.org/rfcdiff?url1=https://wicg.github.io/webpackage/draft-yasskin-webpackage-use-cases.txt&url2=https://jyasskin.github.io/webpackage/archival-use-case/draft-yasskin-webpackage-use-cases.txt)

@ibnesayeed @craigfrancis @TzviyaSiegman, you've been interested in this topic in issues like #105. What should I write in this section? I don't know enough about the topic to even describe the use case well.

I'm probably going to merge this without review to get it in time for the IETF deadline, but I'll definitely fix any comments that show up after that.

### DIFF
--- a/draft-yasskin-webpackage-use-cases.md
+++ b/draft-yasskin-webpackage-use-cases.md
@@ -17,6 +17,12 @@ author:
     email: jyasskin@chromium.org
 
 informative:
+  ISO28500:
+    date: 2017
+    target: "https://www.iso.org/standard/68004.html"
+    title: "WARC file format"
+    seriesinfo:
+      ISO: "28500:2017"
   JAR:
     date: 2014
     target: "https://docs.oracle.com/javase/7/docs/technotes/guides/jar/jar.html"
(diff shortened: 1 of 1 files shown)


craigfrancis:
Thanks for adding these notes @jyasskin 

I'm not sure how much I can add at the moment, as you have noted that packages could be used for archival purposes (actual implementation will probably be for the browsers to decide), and the only other point I have is the suggestion/concerns about JavaScript basing the date/time off the timestamp of the package (which is already covered).

ibnesayeed:
Thanks @jyasskin for this. The language used seems a little too informal. That said, I think we need to invite more people from the web arching community to give their input on the archival aspect here or in #105. In case of WARC file-based replay, certificates of the archive host are used in TLS communication rather than the original ones. I am CCing @anjackson, @hvdsomp, @phonedude, @ikreymer, @galsondor, and @machawk1 here for any comments on it.

jmulliken:
I have some limited experience creating WARCs. Without knowing too much about the back-end, I can add that this approach would work for many html/css/javascript publications. We are currently using it for amenable web-based publications at Stanford University Press. Here, for example, is a web archive of our latest web publication: https://webrecorder.io/sup/enchanting-the-desert I think you could generalize a use case from this, but the type of publication is not the typical kind this group has looked at so far. That said I would LOVE for it to become part of the conversation.

BigBlueHat:
From a Web Publication WG perspective, we'll want Packaged Web Publications (PWP) to live well beyond the rental of a domain name or certificate. Reducing the contents to a sandbox-ed experience may be fine if the contents are primarily (or exclusively) descriptive (i.e. little to no JS involved).

Ideally, one could still experience a PWP they'd downloaded (or purchased) even if the domain or cert expired just days prior. Essentially, we don't want a reader's experience of a publication dependent on a rent-based architecture.

And huge 👍 to involving a wider audience (i.e. WARC and friends) in the development of this potentially very valuable addition to the Web platform.

PR #139, opened 2018-03-07
Start an implementation draft

jyasskin:
[Preview](https://jyasskin.github.io/webpackage/implementation-draft/draft-yasskin-httpbis-origin-signed-exchanges-impl.html), [Diff](https://tools.ietf.org/rfcdiff?url1=https://wicg.github.io/webpackage/draft-yasskin-http-origin-signed-responses.txt&url2=https://jyasskin.github.io/webpackage/implementation-draft/draft-yasskin-httpbis-origin-signed-exchanges-impl.txt)

This shows the state we expect to synchronize on for May.

PR #148, opened 2018-03-15
Replace the application/http-exchange+cbor format with a simpler format.

### LINKED ISSUE #135: Update http-exchange format to improve security+reliability
@davidben points out that it's hard to write streaming parsers, and that the most reliable protocols send a number of bytes to read, read that number of bytes, and then use a non-streaming parser to pull information out of that contiguous data. The format in https://tools.ietf.org/html/draft-yasskin-http-origin-signed-responses-03#section-5.3 doesn't accomplish this because CBOR items are prefixed by their length in items, not bytes.

In multi-process clients, we also want to allow the main/verification process to bound the amount of memory it uses. This implies (1) a length limit on the exchange header set, and (2) mandating use of the MI header instead of the Digest header to verify the payload authenticity, since MI has fixed-size records.

Finally, it would be nice to be able to check signatures against the literal bytes in the http-exchange format instead of needing to reserialize. Reserialization still needs to be *possible* so that intermediates can receive a signature in a normal same-origin HTTP response and then produce the http-exchange format, but final clients shouldn't need to do it.

This probably means the format will move away from being a general http-exchange serialization format and toward being specific to signed exchanges. This has the extra benefit of not needing to say that request payloads and response trailers cause the signature not to verify.

jyasskin:
This format:
* Doesn't require a streaming CBOR parser parse it from a network stream.
* Doesn't allow request payloads or response trailers, which don't fit into
  the signature model.
* Allows checking the signature before parsing the exchange headers.

This is similar to, but not exactly the same as the format @davidben proposed for #139: it pulls out the `Signature` header into bytestring before the other headers so that the second map can be literally the bytes in the signed string. This isn't extensible if we ever want to add new non-signed values, so I added a format version number into the format's magic number.

[Preview](https://jyasskin.github.io/webpackage/no-streaming/draft-yasskin-http-origin-signed-responses.html), [Diff](https://tools.ietf.org/rfcdiff?url1=https://wicg.github.io/webpackage/draft-yasskin-http-origin-signed-responses.txt&url2=https://jyasskin.github.io/webpackage/no-streaming/draft-yasskin-http-origin-signed-responses.txt)

Fixes #135.

### DIFF
--- a/draft-yasskin-http-origin-signed-responses.md
+++ b/draft-yasskin-http-origin-signed-responses.md
@@ -1063,12 +1063,12 @@ and header fields, and a response payload.
 
 This content type consists of the concatenation of the following items:
 
-1. The ASCII characters "sxg" followed by a 0 byte, to serve as a file
+1. The ASCII characters "sxg1" followed by a 0 byte, to serve as a file
    signature.
 
    Note: RFC EDITOR PLEASE DELETE THIS NOTE; The implementation of the final RFC
    MUST use this file signature, but implementations of drafts MUST NOT use it
-   and MUST use another implementation-specific string beginning with "sxg-" and
+   and MUST use another implementation-specific string beginning with "sxg1-" and
    ending with a 0 byte instead.
 1. 3 bytes storing a big-endian integer `sigLength`. If this is larger
    than TBD, parsing MUST fail.
@@ -1106,7 +1106,7 @@ defined in Appendix G of {{?I-D.ietf-cbor-cddl}}, and most of the `Signature`
 header field and payload elided with a ...:
 
 ~~~
-sxg\0<3-byte length of the following header
+sxg1\0<3-byte length of the following header
 value>sig1; sig=*...; integrity="mi"; ...<3-byte length of the encoding of the
 following array>[
   {
@@ -1312,11 +1312,12 @@ Subtype name:  signed-exchange
 
 Required parameters:
 
-* v: An integer denoting the version of the file format. When used with the
-  `Accept` header field (Section 5.3.1 of {{!RFC7231}}), this parameter can be a
-  hyphen (-)-separated range of version numbers, or a comma (,)-separated list
-  of such ranges or individual version numbers. The server is then expected to
-  reply with a resource using a particular version within those ranges.
+* v: An integer denoting the version of the file format. The version defined in
+  this specification is `1`. When used with the `Accept` header field (Section
+  5.3.1 of {{!RFC7231}}), this parameter can be a hyphen (-)-separated range of
+  version numbers, or a comma (,)-separated list of such ranges or individual
+  version numbers. The server is then expected to reply with a resource using a
+  particular version within those ranges.
 
   Note: RFC EDITOR PLEASE DELETE THIS NOTE; Implementations of drafts of this
   specification MUST NOT use simple integers to describe their versions, and
@@ -1342,7 +1343,7 @@ Additional information:
 
   Deprecated alias names for this type:  N/A
 
-  Magic number(s):  73 78 67 00
+  Magic number(s):  73 78 67 31 00
 
   File extension(s): .sxg
 
--- a/draft-yasskin-http-origin-signed-responses.md
+++ b/draft-yasskin-http-origin-signed-responses.md
@@ -1064,7 +1064,8 @@ and header fields, and a response payload.
 This content type consists of the concatenation of the following items:
 
 1. The ASCII characters "sxg1" followed by a 0 byte, to serve as a file
-   signature.
+   signature. This is redundant with the MIME type, and receipients that receive
+   both MUST ignore the values in these first 5 bytes.
 
    Note: RFC EDITOR PLEASE DELETE THIS NOTE; The implementation of the final RFC
    MUST use this file signature, but implementations of drafts MUST NOT use it
@@ -1072,10 +1073,10 @@ This content type consists of the concatenation of the following items:
    ending with a 0 byte instead.
 1. 3 bytes storing a big-endian integer `sigLength`. If this is larger
    than TBD, parsing MUST fail.
-1. `sigLength` bytes holding the `Signature` header field's value
-   ({{signature-header}}).
 1. 3 bytes storing a big-endian integer `headerLength`. If this is larger than
    TBD, parsing MUST fail.
+1. `sigLength` bytes holding the `Signature` header field's value
+   ({{signature-header}}).
 1. `headerLength` bytes holding the signed headers, the canonical serialization
    ({{canonical-cbor}}) of the CBOR representation of the request and response
    headers of the exchange represented by the `application/signed-exchange`
@@ -1086,8 +1087,10 @@ This content type consists of the concatenation of the following items:
 1. The payload body (Section 3.3 of {{!RFC7230}}) of the exchange represented by
    the `application/signed-exchange` resource.
 
-   Note that the use of the payload body here means that the `Transfer-Encoding`
-   header field has no effect.
+   Note that the use of the payload body here means that a `Transfer-Encoding`
+   header field inside the `application/signed-exchange` header block has no
+   effect. A `Transfer-Encoding` header field on the outer HTTP response that
+   transfers this resource still has its normal effect.
 
 ### Cross-origin trust in application/signed-exchange {#co-trust-app-signed-exchange}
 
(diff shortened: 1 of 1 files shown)


draft-yasskin-http-origin-signed-responses.md
```diff
@@ -1054,112 +1054,76 @@ This allows the mitigation against downgrades in {{seccons-downgrades}}, but
...
+   ({{signature-header}}).
+1. 3 bytes storing a big-endian integer `headerLength`. If this is larger than
+   TBD, parsing MUST fail.
+1. `headerLength` bytes holding the signed headers, the canonical serialization
+   ({{canonical-cbor}}) of the CBOR representation of the request and response
+   headers of the exchange represented by the `application/signed-exchange`
+   resource ({{cbor-representation}}), excluding the `Signature` header field.
+
+   Note that this is exactly the bytes used when checking signature validity in
+   {{signature-validity}}.
```
davidben: Yay!

draft-yasskin-http-origin-signed-responses.md
```diff
@@ -1054,112 +1054,76 @@ This allows the mitigation against downgrades in {{seccons-downgrades}}, but
...
+
+   Note: RFC EDITOR PLEASE DELETE THIS NOTE; The implementation of the final RFC
+   MUST use this file signature, but implementations of drafts MUST NOT use it
+   and MUST use another implementation-specific string beginning with "sxg1-" and
+   ending with a 0 byte instead.
+1. 3 bytes storing a big-endian integer `sigLength`. If this is larger
+   than TBD, parsing MUST fail.
+1. `sigLength` bytes holding the `Signature` header field's value
+   ({{signature-header}}).
+1. 3 bytes storing a big-endian integer `headerLength`. If this is larger than
```
davidben: WDYT about putting the two lengths first? (I.e. swap this one and the one above.) One less state in your parsing state machine. (First you read a fixed 5 + 3 + 3 bytes, then you check lengths, then you read additional `sigLength + headerLength` bytes.)

draft-yasskin-http-origin-signed-responses.md
```diff
@@ -1054,112 +1054,76 @@ This allows the mitigation against downgrades in {{seccons-downgrades}}, but
...
+   headers of the exchange represented by the `application/signed-exchange`
+   resource ({{cbor-representation}}), excluding the `Signature` header field.
+
+   Note that this is exactly the bytes used when checking signature validity in
+   {{signature-validity}}.
+1. The payload body (Section 3.3 of {{!RFC7230}}) of the exchange represented by
+   the `application/signed-exchange` resource.
+
+   Note that the use of the payload body here means that the `Transfer-Encoding`
+   header field has no effect.
```
davidben: Nit: Specifically the `Transfer-Encoding` header inside the header block has no effect, whereas the outer `Transfer-Encoding` header which transfers the overall `application/signed-exchange` blob does. Dunno if that's worth clarifying.

draft-yasskin-http-origin-signed-responses.md
```diff
@@ -1054,112 +1054,76 @@ This allows the mitigation against downgrades in {{seccons-downgrades}}, but
...
+  }
+]<!doctype html>\r\n<html>...
 ~~~
 
-### Open Questions ## {#oq-application-http-exchange}
+### Open Questions ## {#oq-application-signed-exchange}
 
-Should `application/http-exchange+cbor` support request payloads and trailers,
-or only the aspects needed for signed exchanges?
+Should this be a CBOR format, or is the current mix of binary and CBOR better?
```
davidben: (I like the current mix, of course. :-) )

draft-yasskin-http-origin-signed-responses.md
```diff
@@ -1054,112 +1054,76 @@ This allows the mitigation against downgrades in {{seccons-downgrades}}, but
...
+
+To allow signed exchanges to be the targets of `<link rel=prefetch>` tags, we
+define the  `application/signed-exchange` content type that represents a signed
+HTTP exchange, including request metadata and header fields, response metadata
+and header fields, and a response payload.
+
+This content type consists of the concatenation of the following items:
+
+1. The ASCII characters "sxg1" followed by a 0 byte, to serve as a file
+   signature.
```
davidben: Does the receiver need to check this value?

draft-yasskin-http-origin-signed-responses.md
```diff
@@ -1054,112 +1054,76 @@ This allows the mitigation against downgrades in {{seccons-downgrades}}, but
...
+
+   Note: RFC EDITOR PLEASE DELETE THIS NOTE; The implementation of the final RFC
+   MUST use this file signature, but implementations of drafts MUST NOT use it
+   and MUST use another implementation-specific string beginning with "sxg1-" and
+   ending with a 0 byte instead.
```
davidben: I might suggest instead saying that draft impls not use the final MIME type instead/additionally. Consider if this format changes a bit. Probably someone will end up in a situation where they need to support two iterations at the same time. That means you first need to read enough to dispatch on the type (which isn't even fixed-width), then parse the rest out. Or you cleverly meld the state machines.

Whereas, if it's in the content type, the HTTP stack has already given you the headers as one atomic thing, so you can go ahead and dispatch there.

Edit: Ah, I see you've put the parameter in the MIME type registration. Maybe make it clearer here that that's going on?

draft-yasskin-http-origin-signed-responses.md
```diff
@@ -1342,24 +1304,36 @@ signed exchange.
...
 
-Required parameters:  N/A
+Required parameters:
+
+* v: An integer denoting the version of the file format. The version defined in
+  this specification is `1`. When used with the `Accept` header field (Section
+  5.3.1 of {{!RFC7231}}), this parameter can be a hyphen (-)-separated range of
+  version numbers, or a comma (,)-separated list of such ranges or individual
+  version numbers. The server is then expected to reply with a resource using a
+  particular version within those ranges.
```
davidben: That seems a lot of complexity to put on the receiver. Why not just define application/signed-exchange-v1, application/signed-exchange-v2, etc. You don't need to parse out all this complex versioning information. We use a lot more bytes but header compression is nice, and is this format really going to have to change that much?

You might be able to make this extensible by saying the headers block is allowed to have unknown keys that you just ignore. This is how TLS does a lot of things. Though then you need to be sure that extensions are all safe for unknowing receivers to ignore.

draft-yasskin-http-origin-signed-responses.md
jyasskin: I think the version identifier has to go in both because people will store these as files, and their servers need to be able to pick the right mime type. But yeah, I've now said the client needs to trust the mime type.

draft-yasskin-http-origin-signed-responses.md
jyasskin: Sure.

draft-yasskin-http-origin-signed-responses.md
jyasskin: Clarified.

draft-yasskin-http-origin-signed-responses.md
jyasskin: Hopefully the format won't change much once it's an RFC, but it is going to change until then. I think I talked to @sleevi about whether to make the headers block (especially the non-signed block) extensible, and he liked making it really simple, hence 21692e13a3b505292fbb248883ba368c4a2f178c.

I think I can remove ranges from the version numbers, since most likely a client will only support 2 versions at a time.

draft-yasskin-http-origin-signed-responses.md
```diff
@@ -1054,112 +1054,79 @@ This allows the mitigation against downgrades in {{seccons-downgrades}}, but
...
+To allow signed exchanges to be the targets of `<link rel=prefetch>` tags, we
+define the  `application/signed-exchange` content type that represents a signed
+HTTP exchange, including request metadata and header fields, response metadata
+and header fields, and a response payload.
+
+This content type consists of the concatenation of the following items:
+
+1. The ASCII characters "sxg1" followed by a 0 byte, to serve as a file
+   signature. This is redundant with the MIME type, and receipients that receive
+   both MUST ignore the values in these first 5 bytes.
```
davidben: I would probably suggest checking it, rather than ignoring it, just to make sure folks get it right. https://tools.ietf.org/html/draft-thomson-postel-was-wrong-02 and stuff.

draft-yasskin-http-origin-signed-responses.md
```diff
@@ -1342,24 +1307,36 @@ signed exchange.
...
+  `version = DIGIT/%x61-7A`) The version defined in this specification is `1`.
+  When used with the `Accept` header field (Section 5.3.1 of {{!RFC7231}}), this
+  parameter can be a comma (,)-separated list of version strings. ({{!RFC5234}}
+  ABNF: `version-list = version *( "," version )`) The server is then expected
+  to reply with a resource using a particular version from that list.
+
+  Note: RFC EDITOR PLEASE DELETE THIS NOTE; Implementations of drafts of this
+  specification MUST NOT use simple integers to describe their versions, and
+  MUST instead define implementation-specific strings to identify which draft is
+  implemented.
```
davidben: Nit: That then suggests that versions aren't digits but opaque strings, right?

jyasskin: Yep, I believe I included that change in b95e497dd2bfb780504220ccd4ebe85fe64baf56 (especially `v: A string denoting the version`), but if you see a word I missed, let me know.

draft-yasskin-http-origin-signed-responses.md
jyasskin: Done.

PR #150, opened 2018-03-16
gen-se: Allow override of content-type response header

nyaxt:
Found by @irori


hajimehoshi:
So the problem was that `Add` could be called multiple times for the same header `content-type`, which was problematic?

irori:
`Add` can be called multiple times for the same header, but `encodeResponseHeaders` uses only the first one (`value[0]` here:
https://github.com/WICG/webpackage/blob/master/go/signedexchange/signedexchange.go#L104)

hajimehoshi:
lgtm

nyaxt:
> encodeResponseHeaders uses only the first one (value[0] here:

This sounds like an another bug.

nyaxt:
@twifkak FYI, while this shouldn't affect "library" usage of gen-se.

PR #155, opened 2018-03-18
Insulate against changes in later versions of draft-thomson-http-mice-02.

jyasskin:
[Preview](https://jyasskin.github.io/webpackage/draft-mice/draft-yasskin-httpbis-origin-signed-exchanges-impl.html), [Diff](https://tools.ietf.org/rfcdiff?url1=https://wicg.github.io/webpackage/draft-yasskin-httpbis-origin-signed-exchanges-impl.txt&url2=https://jyasskin.github.io/webpackage/draft-mice/draft-yasskin-httpbis-origin-signed-exchanges-impl.txt)


nyaxt:
I think the change totally makes sense to go in `draft-yasskin-http-origin-signed-responses.md`, but I'd like to avoid changing the impl snapshot `draft-yasskin-httpbis-origin-signed-exchanges-impl.md` at this point if there is no security concern.

Chromium has already implemented the current `draft-yasskin-httpbis-origin-signed-exchanges-impl.md` under flag, so any further changes will confuse the users who are currently experimenting with the feature.

jyasskin:
Ok, now that https://tools.ietf.org/html/draft-yasskin-httpbis-origin-signed-exchanges-impl-00 is published, this is updated and ready for the -01 implementation draft.

And I should check with @martinthomson whether this is a sensible way to handle changes in his MI draft. I've added a -draft2 suffix to all of the identifiers.

martinthomson:
We kinda screwed up the drafts leading to RFC 8188 by not including a draft-version in the name that we used for deployments.  That's a hard thing to reverse.  If you are serious about changing the mechanism to handle empty bodies properly, then there is a risk that things will change.

davidben:
I'm not sure I follow you. RFC 8188 and draft-thomson-http-mice-02 don't cover the same thing.

martinthomson:
In 8188 we made a few (minor) changes over time.  Each time, we had to deploy a new name.  Given that the chances of changes here are non-trivial, maybe prepare for those changes by adding a draft-number suffix.

jyasskin:
I'll take that as an endorsement of this change, where I'm adding a draft number suffix to the header and content type. Thanks!

PR #157, opened 2018-03-19
gen-signedexchange: Encode multiple values for the same header

### LINKED ISSUE #151: signedexchange: Only first request/response header is adopted when multiple values are specified
https://github.com/WICG/webpackage/blob/master/go/signedexchange/signedexchange.go#L87

In multiple values are specified, it looks like ok to join all the values with comma:

https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html

> Multiple message-header fields with the same field-name MAY be present in a message if and only if the entire field-value for that header field is defined as a comma-separated list [i.e., #(values)]. It MUST be possible to combine the multiple header fields into one "field-name: field-value" pair, without changing the semantics of the message, by appending each subsequent field-value to the first, each separated by a comma. The order in which header fields with the same field-name are received is therefore significant to the interpretation of the combined field value, and thus a proxy MUST NOT change the order of these field values when a message is forwarded.

hajimehoshi:
Fixes #151

CC @twifkak 

PTAL


nyaxt:
cherry-picked into nyaxt/master

twifkak:
Note that https://tools.ietf.org/html/rfc7230#section-3.2.2 calls out an exception for Set-Cookie. Obviously, Set-Cookie shouldn't appear in web packages, but I'm wondering if we can avoid the potential for semantic change (e.g. in any other non-standard headers) by instead looping over `value` and emitting several key/value pairs.

hajimehoshi:
@twifkak Good point. I feel like we should eliminate `Set-Cookie` explicitly. Did you mean we should have a kind of whitelist for standard headers?

twifkak:
Sorry! My comment was based on a mistaken understanding of CBOR. I see now that it supports only one value per key. In that case, we should probably eliminate `Set-Cookie` here, or return an error if more than one `Set-Cookie` is present.

FWIW, my packager tool currently strips all response headers in this list: https://jyasskin.github.io/webpackage/implementation-draft/draft-yasskin-httpbis-origin-signed-exchanges-impl.html#stateful-headers

This could be a useful thing to do at the library level, too, or you may want to keep it generic (e.g. allow users to generate stateful htxg's for testing). Both options seem reasonable to me.

hajimehoshi:
OK, I think stripping the stateful header fields should be done on the library side. I'll create an issue for this and PR soon.

Issue #152, opened 2018-03-16
gen-signedexchange: Add a flag to override signature's date


irori:
That would be useful for generating test htxg.

cc: @hajimehoshi

hajimehoshi:
Done

PR #159, opened 2018-03-20
cmd/gen-signedexchange: Add date flag

hajimehoshi:
#152 

PTAL


irori:
Thanks @hajimehoshi!

@nyaxt would you cherry-pick this into nyaxt/master?

Issue #160, opened 2018-03-20
gen-singedexchange: Cause an error when a stateful header field is found


hajimehoshi:
https://jyasskin.github.io/webpackage/implementation-draft/draft-yasskin-httpbis-origin-signed-exchanges-impl.html#stateful-headers

CC @twifkak @nyaxt

hajimehoshi:
Instead of causing an error, just eliminating sounds fine.

PR #162, opened 2018-03-20
gen-signedexchange: Error with stateful request/response header fields

hajimehoshi:
#160 

PTAL

### DIFF
--- a/go/signedexchange/signedexchange.go
+++ b/go/signedexchange/signedexchange.go
@@ -52,12 +52,12 @@ var (
 func NewExchange(uri *url.URL, requestHeaders http.Header, status int, responseHeaders http.Header, payload []byte, miRecordSize int) (*Exchange, error) {
 	for h, _ := range statefulRequestHeaders {
 		if _, ok := requestHeaders[h]; ok {
-			return nil, fmt.Errorf("signedexchange: request header %s is not available", h)
+			return nil, fmt.Errorf("signedexchange: stateful request header %q can't be captured inside signed exchange", h)
 		}
 	}
 	for h, _ := range statefulResponseHeaders {
 		if _, ok := responseHeaders[h]; ok {
-			return nil, fmt.Errorf("signedexchange: response header %s is not available", h)
+			return nil, fmt.Errorf("signedexchange: stateful response header %q can't be captured inside signed exchange", h)
 		}
 	}
 
--- a/go/signedexchange/signedexchange.go
+++ b/go/signedexchange/signedexchange.go
@@ -51,10 +51,14 @@ var (
 
 func NewExchange(uri *url.URL, requestHeaders http.Header, status int, responseHeaders http.Header, payload []byte, miRecordSize int) (*Exchange, error) {
 	for h, _ := range statefulRequestHeaders {
-		requestHeaders.Del(h)
+		if _, ok := requestHeaders[h]; ok {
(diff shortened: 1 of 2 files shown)


kinu:
Should these rather error out than ignore? (No need to take my comment authoritative / you can wait for nyaxt's response)

hajimehoshi:
First I thought that, but now I am not sure since gen-singnedexchange would not accept any response header including stateful headers, and this would not be useful. I'm fine with either.

nyaxt:
Commented on #160, but +1 on kinu. This should be an error

hajimehoshi:
Done.

go/signedexchange/signedexchange.go
```diff
@@ -26,7 +26,41 @@ type Exchange struct {
 	payload []byte
 }
 
+// https://jyasskin.github.io/webpackage/implementation-draft/draft-yasskin-httpbis-origin-signed-exchanges-impl.html#stateful-headers.
+var (
+	statefulRequestHeaders = map[string]struct{}{
```
twifkak: Consider exporting these two sets. I'd like to offer packager users the option to remove these headers rather than error, as I imagine some web servers probably Set-Cookie automatically. (Let me know if you think that's a bad idea.)

hajimehoshi: Good point. I'd add a new API to strip. I'd like to keep `NewExchange` as it is. The gen-signedexchange user would be able to choose to strip such headers by specifying flag. Does this make sense to you?

hajimehoshi:
@nyaxt Are you happy with this PR? Thanks!

go/signedexchange/signedexchange.go
```diff
@@ -26,7 +26,41 @@ type Exchange struct {
...
+		"Set-Cookie2":               struct{}{},
+		"SetProfile":                struct{}{},
+		"WWW-Authenticate":          struct{}{},
+	}
+)
+
 func NewExchange(uri *url.URL, requestHeaders http.Header, status int, responseHeaders http.Header, payload []byte, miRecordSize int) (*Exchange, error) {
+	for h, _ := range statefulRequestHeaders {
+		if _, ok := requestHeaders[h]; ok {
+			return nil, fmt.Errorf("signedexchange: request header %s is not available", h)
```
nyaxt: stateful request header %q can't be captured inside signed exchange

go/signedexchange/signedexchange.go
```diff
@@ -26,7 +26,41 @@ type Exchange struct {
...
+		}
+	}
+	for h, _ := range statefulResponseHeaders {
+		if _, ok := responseHeaders[h]; ok {
+			return nil, fmt.Errorf("signedexchange: response header %s is not available", h)
```
nyaxt: stateful response header %q can't be captured inside signed exchange

go/signedexchange/signedexchange.go
nyaxt: stateful request header %q can't be captured inside signed exchange

go/signedexchange/signedexchange.go
twifkak: Yep, that works for me.

go/signedexchange/signedexchange.go

go/signedexchange/signedexchange.go

hajimehoshi:
ping @nyaxt

PR #171, opened 2018-04-06
Fix #169: Explicitly ignore SCTs in X.509 and OCSP extensions.

### LINKED ISSUE #169: Cross-origin trust: Source of SCTs
https://wicg.github.io/webpackage/draft-yasskin-http-origin-signed-responses.html#cross-origin-trust

> 6.4 Validate that main-certificate has an sct property (Section 3.3) containing valid SCTs from trusted logs. ([RFC6962])

Does this mean that SCTs have to be included as a map item of the cert-chain CBOR? Or SCTs can be provided as an X.509v3 extension of `main-certificate` or as an OCSP extension of the `ocsp` property?

jyasskin:
Thanks @irori.

@davidben @sleevi, is this the right thing to do? Or should I match [RFC6962](https://tools.ietf.org/html/rfc6962#section-3.3) by requiring clients to look in all three places?

### DIFF
--- a/draft-yasskin-http-origin-signed-responses.md
+++ b/draft-yasskin-http-origin-signed-responses.md
@@ -863,6 +863,9 @@ returns "valid", return "valid". Otherwise, return "invalid".
    1. Validate that `main-certificate` has an `sct` property
       ({{cert-chain-format}}) containing valid SCTs from trusted logs.
       ({{!RFC6962}})
+
+      Note that SCTs embedded in the certificate via an X.509 extension or in
+      the OCSP response via an OCSP extension are ignored.
 1. Return "valid".
 
 ## Stateful header fields {#stateful-headers}


davidben:
I'm confused. What's the motivation for ignoring them? That seems inconsistent, especially considering that SCTs will often be embedded in the certificate.

It also seems more complicated to ignore them for the implementer too. Implementation-wise, that should wants to be thoroughly abstracted away from the web packaging implementation anyway. If not, something is probably structured wrong.

sleevi:
@jyasskin I think we can close as WontFix.

All three sources (certificate, OCSP, extension) should be acceptable. The most likely workflow is that all of the necessary SCTs will be within the certificate itself. However, we need an additional channel to supply SCTs (in the event the cert or OCSP response does not contain them), hence the need for the CBOR side (to match the TLS extension).

So... They should all work :) This is more about mirroring the TLS channels' set of authority expressions

jyasskin:
My argument for ignoring the other two methods is that it's simpler to have only one way to do things. However, if y'all think it's actually simpler to accept them from anywhere, I'll do that instead.

It's definitely not WontFix if we want to accept them from anywhere: the current text says that the SCT has to appear by itself. I have to fix that.

@irori, are you the person implementing this, and were you asking because it's less code to accept them from any of the three places?

sleevi:
Yeah, from a CT ecosystem point of view, it's highly desirable that the CA is responsible for ensuring the compliance of the certificate with various CT policies, either via embedding SCTs in the certificate or via the OCSP response. That ensures a more consistent approach less likely to break due to changes in the CT ecosystem. The availability for the TLS extension (or, in this case, the dictionary) is meant to address scenarios in which the CA does not manage that effectively/at all, or the validity period of the certificate is so great that it exceeds the normal log ecosystem stability range. The former is increasingly less likely as CT deployment rolls out, while the latter has not yet happened and is only theoretical.

The ideal outcome is that Web Packagers do not have to 'think' about CT when packaging/updating signatures, and that it just works. This is a mechanism to allow for explicit control (as some large CDNs prefer, in order to keep certificates and OCSP responses small for clients that don't use CT) or rapid response.

davidben:
CT has a bunch of ways to deliver SCTs because it needed to slot into an existing and complex CA and TLS server ecosystem. You may have CAs that haven't gotten SCT support yet, you may have TLS servers that don't support the extension, you may have server operators that don't know to configure CT, etc. In a vacuum, yes, having only one mechanism would be by far the most preferable, but there were other considerations for it to succeed.

Given that CT thus looks like this, it should be consistent across uses, so that we *don't* need to make the implementation more complex. Chromium's implementation already bundles together the three sources. If the in-progress web packaging implementation isn't using that abstraction, we should fix that, and not the spec.

jyasskin:
Done: https://jyasskin.github.io/webpackage/fix-169/draft-yasskin-http-origin-signed-responses.html#cross-origin-trust

davidben:
In hindsight, I suspect CT could have managed with just the in-cert mechanism and the TLS extension, without leeching off of OCSP, but so it goes. :-)

irori:
Yeah, accepting them is simpler from the implementer's point of view.
Thanks!

PR #173, opened 2018-04-09
Sketch loading and caching in the explainer based on Kinuko's description

jyasskin:
Thanks @kinu! This is a subset of your document that I think will give folks like the TAG enough to go on. The rest of your doc is most of the specification.

https://github.com/jyasskin/webpackage/blob/sketch-loading/explainer.md#signed-exchange-loading-sketch

### DIFF
--- a/explainer.md
+++ b/explainer.md
@@ -49,7 +49,9 @@ intermediate) and forwards it to a client (or another intermediate).
 
 When an HTTP exchange is encoded into a resource, the resource can be fetched
 from a **physical URL** that is different from the **logical URL** of the
-encoded exchange.
+encoded exchange. We talk about the **inner** exchange and its inner request and
+response, the **outer** resource it's encoded into, and sometimes the outer
+exchange that fetches the outer resource.
 
 ## Component documents
 
@@ -91,8 +93,8 @@ Signed exchanges can also be sent to a client in three ways:
    non-origin signatures and to provide an origin-trusted signature to
    intermediates.
 1. Enveloped into the `application/signed-exchange` content type. In this case,
-   the signed exchange has both the logical URL of its embedded request, and the
-   physical URL of the envelope itself.
+   the signed exchange has both the logical URL of its inner request, and the
+   physical URL of the outer envelope.
 1. In an HTTP/2-Pushed exchange.
 
 We publish [periodic snapshots of this
@@ -246,10 +248,10 @@ reviewed for.
 
 ## Signed Exchange Loading Sketch
 
-When an **embedder** prefetches or embeds an enveloped signed exchange, or a
-client navigates from the embedder to an enveloped signed exchange, the
-client goes through several steps to open the envelope and load the signed
-resource.
+When an **embedder** prefetches or embeds an application/signed-exchange
+resource, or a client navigates from the embedder to an
+application/signed-exchange, the client goes through several steps to open the
+outer envelope and load the inner exchange.
 
 ### Fetch the physical URL
 
@@ -260,11 +262,10 @@ embedder's and any parent frame's Content Security Policy, and goes through the
 physical URL's Service Worker for navigations or the embedder's otherwise.
 
 Once the response comes back, its `Content-Type: application/signed-exchange`
-header tells the client that it represents a signed exchange, but it's initially
-treated like any other response: the Response object shown to the Service Worker
-(if any) is the bytes of the `application/signed-exchange` resource, not the
-response inside it, and the `application/signed-exchange` participates in caches
-the same way as any other resource.
+header tells the client that it's the outer resource of a signed exchange, but
+it's initially treated like any other response: the Response object shown to the
+Service Worker (if any) is the bytes of the outer, not the inner response, and
+the outer resource participates in caches the same way as any other resource.
 
 ### Fetch the certificate chain
 
@@ -290,14 +291,13 @@ Each signature with a valid certificate chain is passed on to the next step.
 
 ### Signature verification
 
-Once the certificates are validated and enough of the
-`application/signed-exchange` resource is received to parse the claimed
-signed-exchange headers, the client extracts the logical URL from those headers
-and then tries to find a valid signature over the headers that is trusted for
-the claimed origin. If none of the signatures are valid, it either
+Once the certificates are validated and enough of the outer resource is received
+to parse the claimed inner headers, the client extracts the logical URL from
+those headers and then tries to find a valid signature over the headers that is
+trusted for the logical URL's origin. If none of the signatures are valid, it
+either
 
-1. redirects to the logical URL as if the whole signed exchange were a 302
-   response, or
+1. redirects to the logical URL as if the outer response were a 302 redirect, or
 2. fails with a network error.
 
 We're not yet certain which behavior is best. The first is slightly more
@@ -311,7 +311,7 @@ records into a response stream as they arrive.
 
 ### Prefetching stops here
 
-At this point, the client's behavior depends on whether the signed exchange was
+At this point, the client's behavior depends on whether the outer exchange was
 requested as a [prefetch](https://w3c.github.io/resource-hints/#prefetch). To
 satisfy the [privacy-preserving prefetch](#privacy-preserving-prefetch) use
 case, prefetches can't fully load the logical URL, which would create an HTTP
@@ -330,34 +330,37 @@ simpler than expected.
 ### Caching the signed response
 
 If the signed exchange was requested as a navigation or subresource (i.e. *not*
-prefetches), the client tries to cache the request→response pair it contains.
+prefetches), the client tries to cache the inner exchange.
 
 This *doesn't* happen if:
 
-* The signed exchange's request headers aren't sufficiently similar (TBD) to the
+* The inner request headers aren't sufficiently similar (TBD) to the
   request headers the client would use for a normal request in the same context.
   This prevents a malicious intermediate from sticking the wrong
   content-negotiated resource in the HTTP cache.
-* There's a response in the HTTP (or any?) cache with a newer Date header
-  than the signed exchange's response. This prevents some downgrade attacks.
+* There's a response in the HTTP (or any?) cache with a newer Date header than
+  the inner response's Date header. This prevents some downgrade attacks.
 
 In either of these cases, the client just skips to the redirect in the next
 step.
 
-If we're still here, the signed exchange is put into a "use-once" cache similar
-to the [preload cache](https://github.com/whatwg/fetch/issues/590) and, if the
-response headers allow it (and review of this explainer indicates it's a good
-idea), the [HTTP cache](https://tools.ietf.org/html/rfc7234).
+If we're still here, the inner exchange is treated as if it had been prefetched
+from the original origin. Note that the [prefetch
+specification](https://w3c.github.io/resource-hints/) doesn't say exactly what
+this means. It is similar to the [preload
+cache](https://github.com/whatwg/fetch/issues/590) but situated between the HTTP
+cache and the Service Worker, and not exactly the same in other respects.
 
-If we put the signed exchange in the HTTP cache, its freshness has to be bounded
-by the shorter of the normal HTTP cache lifetime or the signature's expiration.
+If this behavior puts the inner exchange in the [HTTP
+cache](https://tools.ietf.org/html/rfc7234), its freshness has to be bounded by
+the shorter of the normal HTTP cache lifetime or the signature's expiration.
 This is more strict than the bound on cache freshness when it crosses a
 certificate or OCSP response expiration in order to partially mitigate
 [downgrade
 attacks](https://wicg.github.io/webpackage/draft-yasskin-http-origin-signed-responses.html#seccons-downgrades).
 For *later* loads of the logical URL (in particular, not the load that's
 happening through the signed exchange, since it's fulfilled using the
-above-mentioned "use-once" cache), a stale entry can be revalidated in the
+above-mentioned prefetch cache), a stale entry can be revalidated in the
 following ways:
 
 * If the `Signature` is expired but the HTTP caching information is fresh, the
@@ -378,40 +381,54 @@ following ways:
 
 At this point the client redirects to the signed exchange's logical URL. For
 navigations, this request goes through the logical URL's Service Worker. For
-subresources, the embedder's Service Worker already handled the request when it
-chose to return a signed exchange, so we currently think it doesn't need a
-second chance.
+subresources, it goes through the embedder's Service Worker again.
 
-If the request goes through a Service Worker (and caching wasn't skipped above),
-the Service Worker will probably want the `FetchEvent` to include some
-notification that there's a response available without a network request. It may
-be reasonable to ship without this, but we currently think the
-[`preloadResponse`](https://w3c.github.io/ServiceWorker/#fetch-event-preloadresponse)
-field in `FetchEvent` may be enough, although this doesn't provide a place to
-tell the Service Worker about any differences between the signed exchange's
-request and the client's request.
+If the Service Worker forwards its `FetchEvent` to the network, either by
+returning without calling `e.respondWith()` or by calling
+`e.respondWith(fetch(e.request))`, this returns the inner response. The Service
+Worker can also explicitly return something else by calling
+`e.respondWith(somethingElse)`.
 
-The Service Worker can explicitly return the signed exchange's response using
-`e.respondWith(e.preloadResponse)`, implicitly return it either by returning
-without calling `e.respondWith()` or by calling
-`e.respondWith(fetch(e.request))`, or explicitly return something else by
-calling `e.respondWith(somethingElse)`.
+For now, there's no explicit notification to the Service Worker that this
+request is part of handling a signed exchange. Because the inner exchange is in
+the prefetch cache, the Service Worker can check whether it's there by doing a
+fetch with the "only-if-cached" [cache
+mode](https://fetch.spec.whatwg.org/#concept-request-cache-mode), and it might
+make sense to attach some metadata to that response describing the situation,
+but we don't currently propose that metadata. The Service Worker also currently
+gets no description of any differences between the signed inner request and the
+browser's request in `e.request`.
 
 ### Matching prefetches with subresources
 
 If page `A` prefetches two signed exchanges `B.sxg` and `C.sxg` containing
 logical URLs `B` and `C`, respectively, and the user then navigates to `B.sxg`,
 we'd like as many of `B`'s subresource fetches as possible to be fulfilled from
-the prefetched content. It's an open question whether:
+the prefetched content. However, the privacy restriction that prefetches can't
+populate the HTTP cache, combined with considerations around implementation
+complexity, limit which ones we can wire up.
 
 1. B can use `C.sxg` as a subresource and have that fulfilled by `A`'s prefetch
    of `C.sxg`.
-1. B can use `C` as a subresource and have that fulfilled by `A`'s prefetch of
-   `C.sxg`.
 1. B can include a `Link: <C.sxg>; rel=preload` header and then have a `C`
    subresource fulfilled by `A`'s prefetch.
-1. B can include a `Link: <C>; rel=preload` header and then have a `C`
+
+However,
+
+1. B *cannot* use `C` as a subresource and have that fulfilled by `A`'s prefetch of
+   `C.sxg`.
+1. B *cannot* include a `Link: <C>; rel=preload` header and then have a `C`
    subresource fulfilled by `A`'s prefetch.
+
+This means that a page with subresources that wants its referring sites to be
+able to prefetch it in a privacy-preserving way, has to use physical URLs for
+internal links, which means it must be re-signed for each distributing cache.
+
+We consider the centralizing properties of that restriction to be a bug. When
+[bundles](#bundled-exchanges) are more fully specified, we expect sites to be
+able to use logical URLs in their internal links by including both `B` and `C`
+in a single bundle, and then a single signed bundle can be used for multiple
+distributing caches.
 
 ## FAQ
 
--- a/explainer.md
+++ b/explainer.md
@@ -20,11 +20,12 @@ TAG's Web Packaging Draft](https://w3ctag.github.io/packaging-on-the-web/)~~.
   - [Privacy-preserving prefetch](#privacy-preserving-prefetch)
   - [Packaged Web Publications](#packaged-web-publications)
   - [Third-party security review](#third-party-security-review)
-- [Loading sketch](#loading-sketch)
+- [Signed Exchange Loading Sketch](#signed-exchange-loading-sketch)
   - [Fetch the physical URL](#fetch-the-physical-url)
   - [Fetch the certificate chain](#fetch-the-certificate-chain)
   - [Signature verification](#signature-verification)
   - [Prefetching stops here](#prefetching-stops-here)
+  - [No nested signed exchanges](#no-nested-signed-exchanges)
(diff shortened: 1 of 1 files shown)


explainer.md
```diff
@@ -215,6 +243,162 @@ third-party then signs either the exchanges in a package or the package as a
...
+client goes through several steps to open the envelope and load the signed
+resource.
+
+### Fetch the physical URL
+
+The client won't know that a URL holds a signed exchange until it receives the
+`Content-Type` in the response, so the initial request is identical to any other
+request in the same context. It follows redirects, is constrained by the
+embedder's and any parent frame's Content Security Policy, and goes through the
+embedder's or the physical URL's Service Worker.
```
kinu: nit: would it be better to say it depends on if the request is for a navigation or not? (Not sure how this text is confusing or not to the readers)

explainer.md
```diff
@@ -215,6 +243,162 @@ third-party then signs either the exchanges in a package or the package as a
...
+
+Once the response comes back, its `Content-Type: application/signed-exchange`
+header tells the client that it represents a signed exchange, but it's initially
+treated like any other response: the Response object shown to the Service Worker
+is the bytes of the `application/signed-exchange` resource, not the response
+inside it, and the client follows the remaining steps only if the Service Worker
```
kinu: 'only if the Service Worker responds with...' part felt a bit confusing (in this paragraph it's not clear if we're talking about fetching the physical URL in general or the case that with the embedder's SW).

explainer.md
```diff
@@ -215,6 +243,162 @@ third-party then signs either the exchanges in a package or the package as a
...
+
+Each signature with a valid certificate chain is passed on to the next step.
+
+### Signature verification
+
+Once the certificates are validated and enough of the
+`application/signed-exchange` resource is received to parse the claimed
+signed-exchange headers, the client extracts the logical URL from those headers
+and then tries to find a valid signature over the headers that is trusted for
+the claimed origin. If it can't find any, it either
```
kinu: Should it be clearer if we say 'If it fails to validate the signature' given that this talks about general validation failure cases (e.g. clock skew etc)?  (While I also found that we're using the text 'find a valid signature' throughout this change, so this is probably intentional?)

explainer.md
```diff
@@ -215,6 +243,162 @@ third-party then signs either the exchanges in a package or the package as a
...
+### Caching the signed response
+
+If the signed exchange was requested as a navigation or subresource (i.e. *not*
+prefetches), the client tries to cache it.
+
+This *doesn't* happen if:
+
+* The signed exchange's request headers aren't sufficiently similar (TBD) to the
+  request headers the client would use for a normal request in the same context.
+  This may avoid confusing the client?
```
kinu: Do you have some example scenarios?

explainer.md
```diff
@@ -215,6 +243,162 @@ third-party then signs either the exchanges in a package or the package as a
...
+At this point, the client's behavior depends on whether the signed exchange was
+requested as a [prefetch](https://w3c.github.io/resource-hints/#prefetch). To
+satisfy the [privacy-preserving prefetch](#privacy-preserving-prefetch) use
+case, prefetches can't fully load the logical URL, which would create an HTTP
+cache entry that would be visible to the logical URL's server.
+
+Prefetches can and should process any `Link: <>; rel=preload` headers they find,
+as prefetches. If those point at signed exchanges, this process repeats.
+
+### Caching the signed response
```
kinu: Let me clarify... at this point we start talking about caching the response(s) in the signed exchange, is that right?

explainer.md
```diff
@@ -215,6 +243,162 @@ third-party then signs either the exchanges in a package or the package as a
...
+  This may avoid confusing the client?
+* There's a response in the HTTP (or preload?) cache with a newer Date header
+  than the signed exchange's response. This prevents some downgrade attacks.
+
+In either of these cases, the client just skips to the redirect in the next
+step.
+
+If we're still here, the signed exchange is put into the [preload
+cache](https://github.com/whatwg/fetch/issues/590) and, if the response headers
+allow it, the [HTTP cache](https://tools.ietf.org/html/rfc7234).
```
kinu: I don't feel super comfortable having this text yet, UA should respect the cache headers but not sure if it should cache the response in the signed exchange.  Also: mentioning 'preload cache' here feels a bit confusing, it should come into play only when the signed exchange is 'preloaded', right?

explainer.md
jyasskin: My idea here is that signed exchanges can have several signatures, some of which are invalid for this client (but might be valid for some other client), some of which are valid but not origin-trusted, and some of which are both valid and origin-trusted. The basic verification process is a search through those signatures for one that's both valid and origin-trusted. Does that make sense? I definitely want to avoid implying that there's exactly one signature, but I'm happy to reword to make this clearer.

explainer.md
jyasskin: Heh, yes, that wasn't clear at all. Is this new wording better?

explainer.md
jyasskin: 1. The Accept header includes `image/newformat`, the response is actually in that format and so Chrome can't display it, and this winds up in the cache, preventing Chrome from doing a later request with our actual Accept headers, which would content-negotiate for an image we could actually display.
2. The Accept-Language header doesn't match the user's language preferences, and the response is something they can't read. If we dropped the exchange with mis-matched request headers, we might negotiate the right response.

I don't have an example of confusing the client beyond just caching an unusable resource that could have been content-negotiated using a connection to the real server. I guess that any real confusion could also be caused by a server that served a malicious response, so Chrome's already hardened against that.

We could put mismatched request headers in the preload cache and just exclude them from the HTTP cache, so a malicious or buggy intermediate would only hurt itself? It seems simpler to skip all caching, but you'd know better what's easier to implement.

explainer.md
jyasskin: Ah, I hadn't caught that whether to HTTP-cache it was still up in the air. I'll mention that.

I could be misunderstanding the preload cache in general, but my impression was that (whatever we call it) it's the place that stores prefetches, preloads, the thing https://w3c.github.io/ServiceWorker/#navigation-preload-manager manages, and anything else where we want to use an already-stale response "once".

kinu: I see. Since the issue (https://github.com/whatwg/fetch/issues/590) isn't closed yet I added a question to ask for the clarification there, that would probably give you some additional context too. For now I just assume this is talking about 'a cache' for preload/prefetch or whatever.

explainer.md
kinu: Loos a lot clearer, thanks!

explainer.md
```diff
@@ -215,6 +243,162 @@ third-party then signs either the exchanges in a package or the package as a
...
+* There's a response in the HTTP (or preload?) cache with a newer Date header
+  than the signed exchange's response. This prevents some downgrade attacks.
+
+In either of these cases, the client just skips to the redirect in the next
+step.
+
+If we're still here, the signed exchange is put into the [preload
+cache](https://github.com/whatwg/fetch/issues/590) and, if the response headers
+allow it (and review of this explainer indicates it's a good idea), the [HTTP
+cache](https://tools.ietf.org/html/rfc7234).
```
kinu: Reading this further I started to worry that we say a lot about the HTTP cache here.

Does it make sense to have this part (putting it in HTTP cache) as an optional behavior, or could the spec leave which cache to put the resource up to UA?  Say, we can define a conceptual `signed response cache` and note that the behavior can be implemented in the HTTP cache if UA wants?

My concern is layering, efficiency and complexity: The signed exchange itself is cacheable, which seems to mean that the signed exchange layer sits on top of HTTP cache, therefore populating HTTP cache with the contents extracted from the signed exchange could mean:

1. we may waste disk space by caching the data in dup'ed way (signed and non-signed ones)
2. we may need to populate HTTP cache from the layer above it, which is probably unprecedented
3. we need to make the HTTP cache understand the signed exchange logic and associate each entry with the corresponding signature in order to process the revalidation logic stated below

explainer.md
```diff
@@ -215,6 +243,162 @@ third-party then signs either the exchanges in a package or the package as a
...
+
+If we put the signed exchange in the HTTP cache, its freshness has to be bounded
+by the shorter of the normal HTTP cache lifetime or the signature's expiration.
```
kinu: Does this imply that the signature's expiration has somewhat sticky effect on the resources that are installed from the signed exchange / bundle?  For some use cases like offline PWA installation this may not make a lot sense, or may have some inconsistency?  E.g. I think we'd want to keep the PWA's SW and assets installed from a bundle/exchange valid even after the signature expires, is that right?

explainer.md
```diff
@@ -215,6 +243,162 @@ third-party then signs either the exchanges in a package or the package as a
...
+
+At this point the client redirects to the signed exchange's logical URL. For
+navigations, this request goes through the logical URL's Service Worker. For
+subresources, the embedder's Service Worker already handled the request when it
+chose to return a signed exchange, so we currently think it doesn't need a
+second chance.
+
+If the request goes through a Service Worker (and caching wasn't skipped above),
+the `FetchEvent` needs to include some notification that there's a response
+available in the preload cache. We currently think the
```
kinu: I don't think the `FetchEvent` *needs* to include the notification, but if SW wants to know about that we can do that by utilizing navigation preload (only for the SWs that have enabled the feature), is the argument.

explainer.md
jyasskin: Blargh. https://github.com/whatwg/fetch/issues/590#issuecomment-357995100 makes me realize that this is definitely *not* the preload cache. The preload cache has a bit of behavior that I want, but it sits after the Service Worker (i.e. caches SW responses), while we're currently thinking of this thing as sitting before the SW and providing input to it.

You mentioned prefetch "basically just puts things in HTTP cache": does that mean that when the subsequent page requests the prefetched thing, the request goes through the subsequent page's Service Worker, which finds the prefetched thing if it goes to the network?

explainer.md
jyasskin: I agree that we can probably ship without this notification, but I think people will want it sooner or later. I'll downgrade my language here.

The existing navigation preload only applies to navigations, while signed exchanges also offer a response for subresources, so I think we'll need to extend its current spec. I also don't think we need opt-in to expose this: navigation preload needs opt-in because it adds an extra maybe-useless request, but this just exposes something that's happening anyway.

explainer.md
kinu: > You mentioned prefetch "basically just puts things in HTTP cache": does that mean that when the subsequent page requests the prefetched thing, the request goes through the subsequent page's Service Worker, and the SW finds the prefetched thing if the it calls down to the network?

Yeah that what happens.

explainer.md
kinu: I see reg: offline use case.  What do you think about the following two cases:

* UA populates HTTP cache with resources extracted from the signed exchange, and they expire when the signature of them expire
* SW populates Cache Storage with resources come from the signed exchange (and SW may know the fact or not), and they will be around even after the signature of the original exchange expires

My mental model has been that the expiration matters when we process the signed exchange, but once if it's processed the resources that come from the signed exchange could just look similar to others. I think I'm fine with the current text, but wanted to note that (partly because not sticking the signature info is easier to implement).

explainer.md
kinu: I still think having the signed responses only in so-called `preload cache` (or undefined memory cache thing) could make sense, at least from impl pov.  We sill cache the signed exchange, so having a layer that processes it and potentially caches it up to a certain period seems to be fairly natural to me (this is what we do for images, scripts etc), and that resolves your perf concern?  That model fits well with the expiration model too.

explainer.md
```diff
@@ -215,6 +244,175 @@ third-party then signs either the exchanges in a package or the package as a
...
+
+In either of these cases, the client just skips to the redirect in the next
+step.
+
+If we're still here, the signed exchange is put into a "use-once" cache similar
+to the [preload cache](https://github.com/whatwg/fetch/issues/590) and, if the
+response headers allow it (and review of this explainer indicates it's a good
+idea), the [HTTP cache](https://tools.ietf.org/html/rfc7234).
+
+If we put the signed exchange in the HTTP cache, its freshness has to be bounded
```
mnot: HTTP freshness can't be modified like this; any upstream caches (e.g., CDN) won't be aware of these special semantics. 

I think the two approaches you have are:
1. Don't do that -- i.e., make sure HTTP cache lifetime is always shorter than the signature expiration.
2. If a client encounters a fresh cached response with an expired signature, it can force-refresh using things like `Cache-Control: max-age` in requests. Be aware, though, that request cache directives are ignored by many...

explainer.md
```diff
@@ -215,6 +244,175 @@ third-party then signs either the exchanges in a package or the package as a
...
+For *later* loads of the logical URL (in particular, not the load that's
+happening through the signed exchange, since it's fulfilled using the
+above-mentioned "use-once" cache), a stale entry can be revalidated in the
+following ways:
+
+* If the `Signature` is expired but the HTTP caching information is fresh, the
+  client can fetch the `validityUrl `to update just the signature. It *must not*
+  send an `If-None-Match` or `If-Modified-Since` request to update the cache,
+  because the `Signature` expiring means we don't trust the claimed ETag or date
+  anymore.
```
mnot: As above, if the response is still fresh, upstream caches will consider it so, so you'll need to cache-bust.

explainer.md
```diff
@@ -215,6 +244,175 @@ third-party then signs either the exchanges in a package or the package as a
...
+* If the `Signature` is valid but the HTTP caching information is stale, the
+  client can send the logical URL an `If-None-Match` or `If-Modified-Since`
+  request hoping for a 304. Note that the client has to keep the original
+  response headers if it intends to use the `validityUrl` to update the
+  signature in the future.
+* If neither is valid, the client could first update the signature and then
+  check for a 304 (or even do both concurrently), but it may be easier to just
+  do an unconditional request.
```
mnot: I feel like I'm missing something here -- what does "first update the signature and then check for a 304" mean?

explainer.md
```diff
@@ -72,9 +89,15 @@ These signatures can be used in three related ways:
...
-so that test implementations can interoperate.
+Signed exchanges can also be sent to a client in three ways:
+
+1. Using a `Signature` header in a normal HTTP response. This way is used for
+   non-origin signatures and to provide an origin-trusted signature to
+   intermediates.
+1. Enveloped into the `application/signed-exchange` content type. In this case,
+   the signed exchange has both the logical URL of its embedded request, and the
+   physical URL of the envelope itself.
+1. In an HTTP/2-Pushed exchange.
```
mnot: Until their interaction with Push is specified, I wonder if it's better to just leave this out.

explainer.md
```diff
@@ -38,6 +47,10 @@ Firefox) downloads content and uses it. An **intermediate** (like Fastly, the
 AMP cache, or old HTTP proxies) downloads content from its author (or another
 intermediate) and forwards it to a client (or another intermediate).
 
+When an HTTP exchange is encoded into a resource, the resource can be fetched
+from a **physical URL** that is different from the **logical URL** of the
+encoded exchange.
```
mnot: This terminology feels a bit generic; could we come up with something a bit more specific?

jyasskin: I don't have any better terms offhand, but I'll look for some.

explainer.md
jyasskin: In the explainer, I think it makes sense to lay out the goal, even though we haven't written down exactly how we're getting there yet. In the loading specification, I think you're right that I should only add the PUSH route when I can actually write down how the browser processes it.

Does that make sense? If you're not convinced, I'm not dead set on keeping this entry.

There's also a question of whether it makes sense to target Push in any new feature...

explainer.md
jyasskin: If I rewrite this sentence as "If we put the *content* of the signed exchange in the HTTP cache", does that fix things for you? I definitely don't mean that the envelope needs to have different cache behavior, just the signed bits we pull out of the envelope. (I'm looking for clearer words about this; distinguishing the envelope format from the logical HTTP exchange inside it has been tricky.)

I *think* that's ok because the upstream caches like CDNs can only put the contents of a signed exchange into their cache if they understand signed exchanges, which means they're modifying code anyway and can also modify it in this way. I could even be wrong about that because it'd be code owned by two different groups?

explainer.md
jyasskin: To update the signature, the client follows https://wicg.github.io/webpackage/draft-yasskin-http-origin-signed-responses.html#rfc.section.3.6. Then "check for a 304" is probably equivalent to "makes a conditional request". It's intended to be the two bullets above.

explainer.md
jyasskin: Here, the client would make an unconditional request to the `validity-url`, as described by https://wicg.github.io/webpackage/draft-yasskin-http-origin-signed-responses.html#updating-validity, and it's fine for intermediates to cache that with normal HTTP semantics, and if the caching headers are set wrong, the client will probably wind up re-fetching the resource with a normal TLS request to its logical URL (the original publisher's origin), at which point any CDNs are welcome to serve their cached copy if they have one.

explainer.md
```diff
@@ -215,6 +244,175 @@ third-party then signs either the exchanges in a package or the package as a
...
+the prefetched content. It's an open question whether:
+
+1. B can use `C.sxg` as a subresource and have that fulfilled by `A`'s prefetch
+   of `C.sxg`.
+1. B can use `C` as a subresource and have that fulfilled by `A`'s prefetch of
+   `C.sxg`.
+1. B can include a `Link: <C.sxg>; rel=preload` header and then have a `C`
+   subresource fulfilled by `A`'s prefetch.
+1. B can include a `Link: <C>; rel=preload` header and then have a `C`
+   subresource fulfilled by `A`'s prefetch.
```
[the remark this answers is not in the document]
jyasskin: I need to think through how this affects how authors need to write their content, using more time than I have today, but thanks for also thinking about it.

explainer.md
```diff
@@ -215,6 +244,175 @@ third-party then signs either the exchanges in a package or the package as a
...
+We're not yet certain which behavior is best. The first is slightly more
+resilient to clock skew, while the second encourages intermediates to get their
+implementations right.
+
+If the client does find a valid signature, it "releases" the headers for further
+processing and starts validating
+[mi-sha256](https://tools.ietf.org/html/draft-thomson-http-mice-02#section-2)
+records into a response stream as they arrive.
+
+### Prefetching stops here
```
kinu: Looking at the comments by mnot, it seems intermediates may just go on and can cache the responses in the signed exchanges regardless of the request was for prefetches or not. Is that right, and if so should we note something here?

jyasskin: I _think_ @mnot is talking about proxies caching the enveloped signed exchange instead of proxies unwrapping the signed exchange and caching its contents (https://github.com/WICG/webpackage/pull/173#discussion_r181182126), but if he does think they'll blindly unwrap things, I need to think about the implications.

jyasskin:
I've now made the layering more explicit and removed the claim that we'll put the inner exchange in the HTTP cache. There's still a section saying to explore that later.

How's it look now?

explainer.md
```diff
@@ -215,6 +246,226 @@ third-party then signs either the exchanges in a package or the package as a
 whole using a certificate whose metadata reflects whichever property was
 reviewed for.
 
+## Signed Exchange Loading Sketch
+
+Signed exchanges fit into the loading stack between the prefetch cache and
+Service Workers, leading to a stack with the following layers:
+
+Network → HTTP/2 Push cache → HTTP Cache → prefetch cache → **signed exchange
+handling** → Service Workers → preload cache → memory/image cache → actual
```
jyasskin: Does the preload cache might look equivalent to the memory/image cache to developers? If so, I should probably merge them in this list.

explainer.md
```diff
@@ -215,6 +246,226 @@ third-party then signs either the exchanges in a package or the package as a
...
+[Fetch](https://fetch.spec.whatwg.org/)'s terminology.) For navigations, this
+request goes through the publishing URL's Service Worker. For subresources, it
+goes through the embedder's Service Worker again. (TODO: Check that redirects
+actually work this way.)
+
+If the Service Worker fetches a matching URL from the network, either by
+returning without calling `e.respondWith()` or by calling `fetch(...)`, this
+tries to return the response stream that was attached to the redirect. However,
+if either of the following conditions is met, the fetch bypasses the attached
+exchange and continues down to the lower caches and the network:
```
kinu: Does this extra check run only when the fetch comes from/via the Service Worker?  We also say the response is not really distinguishable between inner responses as a part of signed exchange vs regular resources from SW for now, they feel slightly inconsistent to me.

explainer.md
```diff
@@ -215,6 +246,226 @@ third-party then signs either the exchanges in a package or the package as a
...
+for internal links, which means it must be re-signed for each distributing
+cache.
+
+We consider the centralizing properties of that restriction to be a bug. When
+[bundles](#bundled-exchanges) are more fully specified, we expect sites to be
+able to use publishing URLs in their internal links by including both `B` and
+`C` in a single bundle, and then a single signed bundle can be used for multiple
+distributing caches.
+
+### To consider: Cache the inner exchange
```
kinu: Yeah +1 to keep this as an open discussion for now

explainer.md
```diff
@@ -215,6 +246,226 @@ third-party then signs either the exchanges in a package or the package as a
...
+If the Service Worker fetches a matching URL from the network, either by
+returning without calling `e.respondWith()` or by calling `fetch(...)`, this
+tries to return the response stream that was attached to the redirect. However,
+if either of the following conditions is met, the fetch bypasses the attached
+exchange and continues down to the lower caches and the network:
+
+* The inner request headers aren't sufficiently similar (TBD) to the headers in
+  the Request the SW sent. This prevents a malicious intermediate from causing
+  the client to use the wrong content-negotiated resource. If we later put inner
+  responses in the HTTP cache, this also prevents the intermediate from putting
```
kinu: nit: annotate this as (TBD) too?

explainer.md
```diff
@@ -38,6 +47,12 @@ Firefox) downloads content and uses it. An **intermediate** (like Fastly, the
 AMP cache, or old HTTP proxies) downloads content from its author (or another
 intermediate) and forwards it to a client (or another intermediate).
 
+When an HTTP exchange is encoded into a resource, the resource can be fetched
+from a **distributing URL** that is different from the **publishing URL** of the
+encoded exchange. We talk about the **inner** exchange and its inner request and
+response, the **outer** resource it's encoded into, and sometimes the outer
+exchange that fetches the outer resource.
```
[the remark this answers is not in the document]
jyasskin: Hm, yeah, that's unclear. How about "whose response's payload is the outer resource"?

explainer.md
```diff
@@ -60,6 +75,10 @@ draft](https://tools.ietf.org/html/draft-yasskin-http-origin-signed-responses))
 allows a publisher to sign their HTTP exchanges and intermediates to forward
 those exchanges without breaking the signatures.
 
+We publish [periodic snapshots of this
+draft](https://tools.ietf.org/html/draft-yasskin-httpbis-origin-signed-exchanges-impl)
+so that test implementations can interoperate.
```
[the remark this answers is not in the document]
jyasskin: I can avoid moving it in this PR.

explainer.md
jyasskin: The extra check should happen even when there's no SW or no `fetch` handler in the SW. I tend to think of a page without a SW as equivalent to a page whose SW does nothing in its `fetch` handler, so if it looks like I'm deviating from that, or if that's the wrong model for me to have, let me know. :)

explainer.md
jyasskin: Done.

jyasskin:
I think this is pretty much stable now, so I'll merge it. Comments and bugs are still welcome.

PR #179, opened 2018-04-12
Allow certificate re-use for TLS.

jyasskin:
See discussion in
https://lists.w3.org/Archives/Public/ietf-http-wg/2018AprJun/0041.html.

[Preview](https://jyasskin.github.io/webpackage/remove-rsa/draft-yasskin-http-origin-signed-responses.html), [Diff](https://tools.ietf.org/rfcdiff?url1=https://wicg.github.io/webpackage/draft-yasskin-http-origin-signed-responses.txt&url2=https://jyasskin.github.io/webpackage/remove-rsa/draft-yasskin-http-origin-signed-responses.txt)

### DIFF
--- a/draft-yasskin-http-origin-signed-responses.md
+++ b/draft-yasskin-http-origin-signed-responses.md
@@ -905,11 +905,9 @@ able to make even one unauthorized signature.
 Certificates with this extension MUST be revoked if an unauthorized entity is
 able to make even one unauthorized signature.
 
-Conforming CAs MUST mark this extension as critical, and clients MUST NOT accept
-certificates with this extension in TLS connections (Section 4.4.2.2 of
-{{!I-D.ietf-tls-tls13}}).  This simplifies security analysis of this protocol
-and avoids encouraging server operators to put exchange-signing keys on servers
-exposed directly to the internet.
+Conforming CAs MUST NOT mark this extension as critical, and the presence of
+this extension MUST NOT cause clients to reject a certificate in TLS connections
+(Section 4.4.2.2 of {{!I-D.ietf-tls-tls13}}).
 
 # Transferring a signed exchange {#transfer}
 
@@ -1818,6 +1816,8 @@ draft-04
 * Remove support for integrity protection using the Digest header field.
 * Limit the record size in the mi-sha256 encoding.
 * Forbid RSA keys, and only require clients to support secp256r1 keys.
+* Allow certificates with the CanSignHttpExchanges extension to be used as TLS
+  certificates.
 
 draft-03
 


draft-yasskin-http-origin-signed-responses.md
```diff
@@ -501,26 +501,14 @@ to retrieve an updated OCSP from the original server.
...
-         Set `signing-alg` to the result of applying this function to the type of
-         `main-certificate`'s public key. If the function is undefined on this
-         input, return "invalid".
+      1. If `publicKey` is an RSA key, return "invalid".
+      1. If `publicKey` is a key using the secp256r1 elliptic curve, set
+         `signing-alg` to ecdsa_secp256r1_sha256 as defined in Section 4.2.3 of
+         {{!I-D.ietf-tls-tls13}}.
+      1. Otherwise, either return "invalid" or set `signing-alg` to a non-legacy
+         signing algorithm defined by TLS 1.3 or later
+         ({{!I-D.ietf-tls-tls13}}). This choice MUST depend only on
```
nyaxt: Would it be possible to pin the TLS version/draft in the signedexchange spec?

jyasskin: Do you mean to the -28 draft or to exactly TLS 1.3? The RFC is ["Submitted to IESG for Publication"](https://datatracker.ietf.org/doc/draft-ietf-tls-tls13/), so it's unlikely that -28 will change significantly before it's published as an RFC.

It's possible to pin to 1.3, but [Ilari's post](https://lists.w3.org/Archives/Public/ietf-http-wg/2018AprJun/0050.html) seems to say that he wants this RFC to allow for future expansion in the set of algorithms without needing to revise this RFC. Other security folks might speak up and disagree, of course.

Note that the WHATWG-side spec will be a living spec that'll follow the current version of TLS and specify exactly which algorithms browsers are using, even if this spec is less precise.

sleevi:
So, I am not convinced that this is a positive change we should make. While I can understand the proposal to restrict the signature algorithms, the reuse with TLS is something that I believe is a net-negative for the practical deployment scenario. While I realize that some providers, particularly those BYOC, may have difficulty managing two certificates, the separation requirement was mitigating both operational and cryptographic concerns.

That is, in a forced-separation model, there is no direct advantage or incentive to promoting online signing, while that is fundamentally necessary for TLS certificates. The same discussion can be seen regarding the key protections for Secondary Certificates. This is critically important when considering "key compromise" scenarios. While the current separation does not absolutely guarantee any operational security, it ensures the design does not incentivize insecurity. This is the same consideration as when thinking about forward-dating signatures, and just as in that case, there may be future operational or protocol restrictions (e.g. needing to cover an OCSP response), one can imagine that Web Packages may have future requirements similar to code signing certificates in the Microsoft ecosystem - namely, can only be issued to hardware-backed keys (as an example)

At this time, I do not think allowing the reuse with TLS is good for clients or for servers, despite the challenges some have raised.

cramforce:
@sleevi I think you are missing at least one part of the ecosystem incentive that I know you would usually care about:

1. Allowing reuse with TLS is good for decentralization because it allows parties to reuse their cert to participate in package generation with minimal added effort.
2. Forbidding reuse incentivizes outsourcing of package generation to specialized and naturally more centralized entities who generate the special certificates on behalf of the origin and generate packages for them.

sleevi:
@cramforce No, I'm not missing those from the considerations. I think the cross-protocol and key compromise risks are significantly greater through the introduction of Web Packaging (or Secondary Certificates), and thus deserve serious mitigations if clients are to consider implementing and shipping them. I do not think it is a positive step forward to ship without such mitigations.

Your first argument is not consistent with the efforts that Google or the Chrome team have encouraged or the statements made regarding the automation of certificates. An alternative phrasing is that Web Packaging's practical deployment encourages greater automation through certificate issuance - which we already see widespread. Thus, the concerns about getting a second certificate are merely a misdirect and not applicable.

Regarding your second argument, this is true already for CDNs, which can offer greater key protection for nodes than those afforded to clients. This is fundamentally no different than existing solutions such as Cloudflare, Amazon, or Google, all of which take steps to secure client keys from compromise. My concern would be if the use of a particular CDN or provider was functionally required or incentivized, but that is separate from an argument against CDNs.

In light of the security risks that clients would be faced with, I do think this change is good, and I worry that it would make Web Packaging more difficult to reason about and accept from a security perspective. But I wonder whether we should be continuing this conversation on the list instead - I'm not sure @jyasskin's desired workmode, but since the conversation started there, it's probably best to continue it there.

davidben:
Additionally this PR forgets to restore those 32 spaces in the context string.

Regardless of which way this goes, as I mentioned when removing those spaces came up, I think removing them in the first place was a bad idea. It was only a matter of time before someone tries to remove the restriction and then forgets to fix the context string to match.

davidben:
Note that, by requiring P-256 keys, most sites will already have to manage multiple certificates anyway. There are enough RSA-only TLS clients out there that, in practice, most folks need to have an RSA certificate. ECDSA certificates are primarily deployed in *addition* to RSA certificates, to reduce server resource usage (ECDSA signing is much faster than RSA signing).

jyasskin:
Thanks for having the main discussion on the list; I think more people can participate there.

@davidben Your bug is now fixed; thanks for pointing it out.

jyasskin:
Most of this change was merged in #181, and Ryan's objections in https://lists.w3.org/Archives/Public/ietf-http-wg/2018AprJun/0070.html seem to have convinced everyone to stop pushing for certificate re-use, so I'm closing this PR.

PR #181, opened 2018-04-18
Forbid RSA and only require secp256r1.

jyasskin:
See discussion in
https://lists.w3.org/Archives/Public/ietf-http-wg/2018AprJun/0041.html.

This is #179 without the change to allow exchange-signing certificates to be used for TLS.

[Preview](https://jyasskin.github.io/webpackage/remove-rsa-no-tls/draft-yasskin-http-origin-signed-responses.html), [Diff](https://tools.ietf.org/rfcdiff?url1=https://wicg.github.io/webpackage/draft-yasskin-http-origin-signed-responses.txt&url2=https://jyasskin.github.io/webpackage/remove-rsa-no-tls/draft-yasskin-http-origin-signed-responses.txt)

### DIFF
--- a/draft-yasskin-http-origin-signed-responses.md
+++ b/draft-yasskin-http-origin-signed-responses.md
@@ -500,26 +500,14 @@ to retrieve an updated OCSP from the original server.
          this returns "invalid", return "invalid".
       1. Let `main-certificate` be the first certificate in `certificate-chain`.
       1. Set `publicKey` to `main-certificate`'s public key.
-      1. The client MUST define a partial function from public key types to
-         signing algorithms, and this function must at the minimum include the
-         following mappings:
-
-         RSA, 2048 bits:
-         : rsa_pss_rsae_sha256 or rsa_pss_pss_sha256, as defined in Section
-           4.2.3 of {{!I-D.ietf-tls-tls13}}, depending on which of the
-           rsaEncryption OID or RSASSA-PSS OID {{!RFC8017}} is used.
-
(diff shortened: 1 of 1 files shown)


draft-yasskin-http-origin-signed-responses.md
```diff
@@ -794,15 +777,14 @@ Accept-Signature: mi/mi-sha256
 states that the client will accept signatures with payload integrity assured by
 the `MI` header and `mi-sha256` content encoding and implies that the client
 will accept integrity assured by the `Digest: SHA-256` header and signatures
-from 2048-bit RSA keys and ECDSA keys on the secp256r1 and secp384r1 curves.
+from ECDSA keys on the secp256r1 curve.
 
 ~~~http
-Accept-Signature: -rsa/2048, rsa/4096
+Accept-Signature: -ecdsa/secp256r1, ecdsa/secp384r1
```
davidben: Nit: You don't want the ecdsa/secp384r1 here, right?

jyasskin: It was an intentional part of the example, showing how an odd client, that changed which curve is supported, would announce that fact to servers. Let me know if you think I'm wrong to have an example like that.

(I'm not certain my `Accept-Signature` thing is a good idea overall, but that's a discussion for another PR.)

davidben: Oh! Sorry, I got confused and read the text above as describing this example rather than the text below. Ignore me.

PR #182, opened 2018-04-18
Remove the "Digest" integrity header and bound MI record size.

### LINKED ISSUE #138: Remove Digest in favor of MI
For DoS reasons, the low-level portions of the networking stack must never buffer memory unboundedly. That means the Digest option is not usable, unless we move the hash check to the higher-level portions, but integrity is handled in the low-level portions. This means the low-level portions must return data with a "this is unchecked, please check it later". This is a classic mistake in designing a crypto system. All in all, this means that it is unlikely that any implementation of Digest could pass muster in Chrome.

Additionally, adding random redundant options is generally a bad idea. There's no reason to offer Digest when MI suffices. Please remove Digest from the specification completely.

jyasskin:
Digests require buffering the whole response body before determining that it's
valid, which unnecessarily increases the complexity in the verifier.

This change also puts a bound on the MI record size, also to limit the needed buffering.

@davidben, have I explained the right reasons for doing this?

Fixes #138

[Preview](https://jyasskin.github.io/webpackage/remove-digest/draft-yasskin-http-origin-signed-responses.html), [Diff](https://tools.ietf.org/rfcdiff?url1=https://wicg.github.io/webpackage/draft-yasskin-http-origin-signed-responses.txt&url2=https://jyasskin.github.io/webpackage/remove-digest/draft-yasskin-http-origin-signed-responses.txt)

### DIFF
--- a/draft-yasskin-http-origin-signed-responses.md
+++ b/draft-yasskin-http-origin-signed-responses.md
@@ -1714,6 +1714,12 @@ Simpler implementations are, all things equal, less likely to include bugs. This
 section describes decisions that were made in the rest of the specification to
 reduce complexity.
 
+### Limited choices
+
+In general, we're trying to eliminate unnecessary choices in the specification.
+For example, instead of requiring clients to support two methods for verifying
+payload integrity, we only require one.
+
 ### Bounded-buffering integrity checking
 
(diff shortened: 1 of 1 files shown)


jyasskin:
I've now mentioned that motivation too.

Issue #191, opened 2018-05-21
Request for fixed length magic string: "sxg1-"


nyaxt:
In the current spec draft, the magic string is allowed to be arbitrary length for implementation drafts.

> Note: RFC EDITOR PLEASE DELETE THIS NOTE; The implementation of the final RFC MUST use this file signature, but implementations of drafts MUST NOT use it and MUST use another implementation-specific string beginning with “sxg1-“ and ending with a 0 byte instead.

This makes its streaming parsing non-trivial, since we need to scan for `"\0"` to appear, rather than reading fixed length record.

Two proposals:
- Allocate fixed length (8bytes?) for the magic string field, and pad the rest of the bytes `"\0"` 
  - Will file a PR for this
- Separate version field?

nyaxt:
We would like to simplify the decoder which would live in the browser process as much as possible. I will draft a change that basically does:
```
"sxg1" magic + 3byte ver + 3byte siglen + 3byte hdrlen + ...
```

> I think an implementation can parse the current format by reading as many bytes as the longest magic string it supports, and then memcmp'ing against each of the magic strings including their trailing '\0'.

Yes, but it would require concatenating the overread portion to the start of {signature,headers} buffer.

davidben:
> Yes, but it would require concatenating the overread portion to the start of {signature,headers} buffer

It is also possible that the signature/headers buffer may be smaller than the overread and you go into the body itself.

That said, you likely should be doing this anyway because issuing small reads at a time is not very efficient. (I don't remember being asked to review any CLs to implement the new format... I assume  they're just still in progress?) If you look at Chromium's HTTP/1.1 implementation, we read data in large chunks at a time and, if we got more or less than the full HTTP headers, we just handle that case.

nyaxt:
I haven't written any code yet. (Mainly landing golang lib impl right now)

The current impl for the old format basically issues 3 reads at the minimum:
- SignedExchangeHandler: Read fixed header's header === {magic, headerLen}
- SignedExchangeHandler: Read CBOR header of headerLen
- MercleIntegritySourceStream: Read payload
  - SignedExchangeHandler moves ownership of the `source_` stream to MercleIntegritySourceStream

We can basically keep this 3+ read scheme if the new "header's header" is const, but needs to do buffering (as @davidben mentioned) to handle variable len magic.

[...]

PR #192, opened 2018-05-21
Pad magic string to 8 bytes

nyaxt:
Addresses #191

### DIFF
--- a/draft-yasskin-http-origin-signed-responses.md
+++ b/draft-yasskin-http-origin-signed-responses.md
@@ -1057,14 +1057,14 @@ and header fields, and a response payload.
 
 This content type consists of the concatenation of the following items:
 
-1. The ASCII characters "sxg1" followed by a 0 byte, to serve as a file
-   signature. This is redundant with the MIME type, and receipients that receive
-   both MUST check that they match and stop parsing if they don't.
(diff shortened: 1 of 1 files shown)


draft-yasskin-http-origin-signed-responses.md
```diff
@@ -1057,14 +1057,14 @@ and header fields, and a response payload.
...
+1. The ASCII characters "sxg1" followed by 0 byte repeated 4 times, to serve as
+   a file signature. This is redundant with the MIME type, and receipients that
+   receive both MUST check that they match and stop parsing if they don't.
 
    Note: RFC EDITOR PLEASE DELETE THIS NOTE; The implementation of the final RFC
    MUST use this file signature, but implementations of drafts MUST NOT use it
-   and MUST use another implementation-specific string beginning with "sxg1-" and
-   ending with a 0 byte instead.
+   and MUST use another implementation-specific string beginning with "sxg1-"
+   instead, and have it padded with 0 bytes to preserve total length of 8 bytes.
```
davidben: Especially since this already only leaves 3 bytes anyway, probably better to simply write:
> and MUST use another 8-byte implementation-specific string beginning with "sxg1-".

I.e. no sense in talking about padding schemes and such. You already only have 3 bytes anyway.

jyasskin:
Per discussion in #191, I think this isn't needed anymore.

PR #193, opened 2018-05-22
Update key strings in serializeSignedMessage to lower-case

irori:
This is a follow-up to #190.

#178 also changed the key strings in the canonical CBOR map.


nyaxt:
lgtm. Thanks for the catch

PR #194, opened 2018-05-22
Fix the HTTP/1.1 exchanges to use Host headers instead of full URLs in the GET line.

jyasskin:
Thanks to @triblondon for noticing in
https://github.com/w3ctag/design-reviews/issues/235#issuecomment-379424440.

[Preview](https://jyasskin.github.io/webpackage/fix-http1-samples/draft-yasskin-http-origin-signed-responses.html), [Diff](https://tools.ietf.org/rfcdiff?url1=https://wicg.github.io/webpackage/draft-yasskin-http-origin-signed-responses.txt&url2=https://jyasskin.github.io/webpackage/fix-http1-samples/draft-yasskin-http-origin-signed-responses.txt)


nyaxt:
How about `:path` + `:host` to deduplicate host?

jyasskin:
It's more that the `Host` header isn't ever going to actually appear in the encoding. https://wicg.github.io/webpackage/draft-yasskin-http-origin-signed-responses.html#same-origin-response doesn't allow it, and https://wicg.github.io/webpackage/draft-yasskin-http-origin-signed-responses.html#application-signed-exchange lets the publisher pick their headers, and there's no reason for a publisher to include `Host` redundantly with the `:url`.

Do you think it's ok to show the misleading mapping, or should I switch it to an HTTP/2 representation that doesn't have the Host header in the first place, or something else?

nyaxt:
I'd like to check my understanding first.
Am I correct that:
- The modified `3.2.1` example is only focused on how the http text format maps to CBOR representation
  - It is out-of-scope of the section that the CBOR representation in the signature message may have filtered it out.
  - It is possible but should be very rare for a signature message have `Host` request header in practice, since:
     - application/signed-exchange packagers can choose to only include `:url` 
     - When transferring the exchange via H2, most servers would have no reason to include it in the list of Signed-Headers

nyaxt:
(Question from @horo-t)
Do you have any recommendation on how browsers should react if it actually received a signed exchange with conflicting `:url` and `Host` in its request headers?

jyasskin:
That all sounds correct. How does it sound to say that a signed exchange with conflicting `:url` and `Host` will never [be sufficiently similar to the browser's request](https://github.com/WICG/webpackage/blob/master/explainer.md#navigations-and-subresources-redirect)?

nyaxt:
SGTM

jyasskin:
I felt bad about including the misleading `Host` headers, so I've now excluded them explicitly.

nyaxt:
still lgtm. I like the latest change better.

PR #195, opened 2018-05-23
Use the signature context string to distinguish draft versions.

jyasskin:
Thanks @sleevi for noticing the problem.

[Preview](https://jyasskin.github.io/webpackage/include-version-in-sig-context/draft-yasskin-http-origin-signed-responses.html), [Diff](https://tools.ietf.org/rfcdiff?url1=https://wicg.github.io/webpackage/draft-yasskin-http-origin-signed-responses.txt&url2=https://jyasskin.github.io/webpackage/include-version-in-sig-context/draft-yasskin-http-origin-signed-responses.txt)

### DIFF
--- a/draft-yasskin-http-origin-signed-responses.md
+++ b/draft-yasskin-http-origin-signed-responses.md
@@ -1821,6 +1821,7 @@ draft-04
 * Require absolute URLs.
 * Make all identifiers in headers lower-case, as required by Structured Headers.
 * Switch back to the TLS 1.3 signature format.
+* Include the version and draft number in the signature context string.
 * Remove support for integrity protection using the Digest header field.
 * Limit the record size in the mi-sha256 encoding.
 * Forbid RSA keys, and only require clients to support secp256r1 keys.
--- a/draft-yasskin-http-origin-signed-responses.md
+++ b/draft-yasskin-http-origin-signed-responses.md
@@ -518,7 +518,13 @@ to retrieve an updated OCSP from the original server.
    anyone uses the same key in a TLS certificate and an exchange-signing
    certificate.
    1. A string that consists of octet 32 (0x20) repeated 64 times.
-   1. A context string: the ASCII encoding of “HTTP Exchange”.
+   1. A context string: the ASCII encoding of "HTTP Exchange".
+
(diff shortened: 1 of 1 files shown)


sleevi:
Happy to explain. This is about mitigation version rollback attacks. Right now, the version of the envelope is outside of the signed data. This creates the opportunity for an attacker to alter the version number, causing the signed data to be interpreted differently.

In Chromium, it came up in exploring the b0 to b1 transition, in which b0 ignores the OCSP response, but b1 requires it. An attacker could take a signed b1 response, mark it as b0, and bypass those security checks.

Similar to how X.509 certificates include the signature algorithm within the inner (signed) context, in addition to the outer envelope, or how TLS includes the version number as part of the session transcript, the goal is to ensure that the externally-negotiated properties (such as version) are also bound within the internal, signed content, to prevent any confusion issues. These are functionally the same as cross-protocol attacks, it just happens that its different versions of the same protocol (but functionally different protocols).

The general method of mitigating these cross-protocol attacks is using the context string, to ensure that distinct signatures are created for distinct protocols. This change is necessary, but I'm not entirely sure it's sufficient - if the Web Packaging protocol evolves over time / additional versions, the context string of those subsequent versions will still need to migrate.

Thus, it may make sense to explicitly include both versioning and draft numbers, of which the latter is more applicable here, and formalize that as an inherent part of the protocol. That's the sufficient part :)

davidben:
> I'm curious about the motivation

All signature payloads with the same key be distinguishable, otherwise an attacker can cause the verifier to misinterpret a signature made in one protocol as something else in a different protocol. This includes differences between different versions of the same protocol.

As a toy example, consider if one version signed request headers followed by response headers, and then another version switched the order. Without baking the version into the signature, I can trick the receiver into accepting a signed exchange different from what the origin actually signed.

sleevi:
@jyasskin That sounds good, as it makes sure the expectation is setup in the base spec and so we remember to rev on any semantically-relevant change (overall) and on drafts :)

jyasskin:
Done.

twifkak:
Thanks for the explanations!

nyaxt:
Would you add a ChangeLog entry too?

jyasskin:
> Would you add a ChangeLog entry too? — @nyaxt 

Whoops, thanks, done.

PR #199, opened 2018-05-29
Update context string to follow #195

nyaxt:


irori:
I think we should use a draft-specific string as noted in the spec, such as `"HTTP Exchange 1 b1"`.

go/signedexchange/signer.go
```diff
@@ -15,6 +15,10 @@ import (
 	"github.com/WICG/webpackage/go/signedexchange/cbor"
 )
 
+// The "context string" in Step 7.2 of
+// https://wicg.github.io/webpackage/draft-yasskin-http-origin-signed-responses.html#signature-validity
+const ContextString = "HTTP Exchange 1"
```
hajimehoshi: Following the convention of Go, comments should start with `ContextString is...`.

hajimehoshi: Do we need to export this const? If not, this should be `contextString`

nyaxt: Done

PR #208, opened 2018-06-11
cmd: introduce dump-signedexchange tool

hajimehoshi:
This PR introduces dump-exchange tool to pretty-print an signedexchange file.


nyaxt:
Would you split the PR to only commit go/signedexchange/cbor/decoder* for now?
Please update the dump-signedexchange tool so that it can actually decode the file generated by gen-signedexchange

hajimehoshi:
Sure, let me close this.

PR #210, opened 2018-06-12
Make MI encoding optional

nyaxt:
WebBundling format doesn't require the payload to be MI encoded. This PR splits the encoding of the payload, so that it is possible to create an exchange without MI encoding.

### DIFF
--- a/go/signedexchange/signedexchange.go
+++ b/go/signedexchange/signedexchange.go
@@ -2,6 +2,7 @@ package signedexchange
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
 	"io"
 	"net/http"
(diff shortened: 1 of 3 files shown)


go/signedexchange/signedexchange.go
```diff
@@ -29,7 +30,7 @@ type Exchange struct {
 
 var HeaderMagicBytes = []byte("sxg1-b1\x00")
 
-func NewExchange(uri *url.URL, requestHeaders http.Header, status int, responseHeaders http.Header, payload []byte, miRecordSize int) (*Exchange, error) {
+func NewExchange(uri *url.URL, requestHeaders http.Header, status int, responseHeaders http.Header, payload []byte) (*Exchange, error) {
```
hajimehoshi: This function no longer needs to return error?

PR #211, opened 2018-06-12
Introduce gen-bundle tool to emit "Bundled HTTP Exchanges" [1]

nyaxt:
[1] https://wicg.github.io/webpackage/draft-yasskin-dispatch-bundled-exchanges.html

### DIFF
--- a/go/bundle/bundle.go
+++ b/go/bundle/bundle.go
@@ -1,0 +1,237 @@
+package bundle
+
+import (
+	"bytes"
+	"encoding/binary"
+	"fmt"
+	"io"
+
+	"github.com/WICG/webpackage/go/signedexchange"
+	"github.com/WICG/webpackage/go/signedexchange/cbor"
+)
+
+var HeaderMagicBytes = []byte{0x84, 0x48, 0xf0, 0x9f, 0x8c, 0x90, 0xf0, 0x9f, 0x93, 0xa6}
+
+const FooterLength = 9
+
+type Input struct {
+	Exchanges []*signedexchange.Exchange
+}
+
+// staging area for writing index section
+type indexSection struct {
+	mes   []*cbor.MapEntryEncoder
+	bytes []byte
+}
+
+func (is *indexSection) addExchange(e *signedexchange.Exchange, offset, length int) error {
+	me := cbor.GenerateMapEntry(func(keyE *cbor.Encoder, valueE *cbor.Encoder) {
+		if err := e.EncodeRequestWithHeaders(keyE); err != nil {
+			panic(err) // fixme
+		}
+		if err := valueE.EncodeArrayHeader(2); err != nil {
+			panic(err)
+		}
+		if err := valueE.EncodeUInt(uint64(offset)); err != nil {
+			panic(err)
+		}
+		if err := valueE.EncodeUInt(uint64(length)); err != nil {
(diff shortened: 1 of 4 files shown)


go/bundle/bundle.go
```diff
@@ -0,0 +1,237 @@
...
+	"bytes"
+	"encoding/binary"
+	"fmt"
+	"io"
+
+	"github.com/WICG/webpackage/go/signedexchange"
+	"github.com/WICG/webpackage/go/signedexchange/cbor"
+)
+
+var HeaderMagicBytes = []byte{0x84, 0x48, 0xf0, 0x9f, 0x8c, 0x90, 0xf0, 0x9f, 0x93, 0xa6}
```
hajimehoshi: `headerMagicBytes` to be unexported?

If this is used only in `WriteBundle`, I'd move there.

go/bundle/bundle.go
```diff
@@ -0,0 +1,237 @@
...
+
+const FooterLength = 9
```
hajimehoshi: `footerLength` in `writeFooter`

go/bundle/bundle.go
```diff
@@ -0,0 +1,237 @@
...
+
+type Input struct {
+	Exchanges []*signedexchange.Exchange
```
hajimehoshi: Do you have a plan to add members to the input? If no, just using slice would be enough.

go/signedexchange/signedexchange.go
```diff
@@ -51,6 +51,10 @@ func NewExchange(uri *url.URL, requestHeaders http.Header, status int, responseH
 	}, nil
 }
 
+func (e *Exchange) Payload() []byte {
```
hajimehoshi: Hmm, this sounds unsafe since payload can be modified on the caller side. At least we should have comments.

go/bundle/countingwriter.go
```diff
@@ -0,0 +1,53 @@
...
+	for {
+		var nr int
+		nr, err = r.Read(buf)
+		if err != nil {
+			return
+		}
+
+		var nw int
+		nw, err = cw.w.Write(buf[:nr])
+		if err != nil {
```
hajimehoshi: Where is error gone?

go/bundle/bundle.go
nyaxt: The magic bytes are useful by itself. The clients can see if a given binary is a bundle or not.

go/bundle/bundle.go
nyaxt: Yes. The implementation only encodes "index","responses" sections atm, but there are other sections (such as manifest)

go/bundle/countingwriter.go
nyaxt: The func uses named return vars

hajimehoshi: Ack

go/bundle/bundle.go
hajimehoshi: I'm against exposing a `[]byte`.

> The clients can see if a given binary is a bundle or not.

Would it be possible to add a function to do that?

hajimehoshi: Discussed offline, and we compromised this since there is no good substitution.

PR #213, opened 2018-06-13
Rename bundle.{Input => Bundle}, and bundle.{WriteBundle => Write}

nyaxt:
To prepare for bundle.Read


go/bundle/bundle.go
```diff
@@ -195,13 +195,13 @@ func writeFooter(w io.Writer, offset int) error {
 	return nil
 }
 
-func WriteBundle(w io.Writer, i *Input) error {
+func Write(w io.Writer, b *Bundle) error {
```
hajimehoshi: optional: How about implementing `WriteTo`?

https://golang.org/pkg/io/#WriterTo

nyaxt: Done

PR #218, opened 2018-06-15
Add note about `ignoredSections`

nyaxt:
There doesn't seem to be a section that "says not to process other sections" as of now. Can we omit the steps until we have one?

### DIFF
--- a/draft-yasskin-wpack-bundled-exchanges.md
+++ b/draft-yasskin-wpack-bundled-exchanges.md
@@ -268,6 +268,10 @@ steps, taking the `stream` as input.
    `knownSections` says not to process other sections, add those sections' names
    to `ignoredSections`.
 
+   Note: The `ignoredSections` enables sections that supercede other sections
+   to be introduced in the future. Implementations that don't implement any such
+   sections are free to omit the relevant steps.
+
 1. Let `metadata` be an empty map ({{INFRA}}).
 
 1. For each `"name"`/\[`offset`, `length`] triple in `sectionOffsets`:


jyasskin:
This is here so that a future specification can define a section that replaces another one without needing to re-release this specification. This step definitely doesn't need to appear in an implementation until it implements a section that says to ignore another one, but to remove it from the specification, I'd want an argument that this kind of section is fundamentally a bad idea.

nyaxt:
Got it.

I was a bit confused when I was implementing the steps line-by-line. Would it be able to add a note how this is going to be used in future?

jyasskin:
If you were confused, it probably deserves a note. Do you want to try to word it, or should I try next week?

nyaxt:
Tried something. PTAL

draft-yasskin-wpack-bundled-exchanges.md
```diff
@@ -268,6 +268,11 @@ steps, taking the `stream` as input.
    `knownSections` says not to process other sections, add those sections' names
    to `ignoredSections`.
 
+   Note: The `ignoredSections` enables sections that supercedes other sections
```
jyasskin: s/supercedes/supercede/

draft-yasskin-wpack-bundled-exchanges.md
```diff
@@ -268,6 +268,11 @@ steps, taking the `stream` as input.
...
+   to be introduced in the future. Since there are no such sections in the
```
jyasskin: How about

> Implementations that don't implement any such sections are free to omit the relevant steps.

since even after we add such a section to the registry, implementations that don't implement it will still be able to skip the `ignoredSections` steps.

I expect to get some pushback based on the fact that implementations are [always allowed to differ from the specified steps as long as they get the same answer](https://infra.spec.whatwg.org/#algorithms).

PR #226, opened 2018-06-20
In both random-access maps, store the index and contents in the same order.

### LINKED ISSUE #212: Structured responses section for streaming loading
# Goal
- Easier streaming consumption of large bundles.
- Easier non-streaming consumption of the small bundles.

# Background
In the current bundle spec, the structure of the responses section is allowed to be very flexible and not required to be a CBOR construct.

## Streaming loading algorithm for current format
When loading a bundle in a streamed fashion, currently the parser will:
1. Consume the index section
1. Sort the exchanges by their response range offset
1. Consume the other sections up till the responses section
1. Discard the bytes until it reaches the first exchange in the sorted list
1. Buffer the bytes until it reaches the end of the the first exchange
1. The first exchange becomes ready for its use:
  1. Parse its responses header CBOR map
  1. Read its payload
1. Append bytes to the buffer until it reaches the end of the next item.
1. The second exchange becomes ready for its use.
1. ...

Since there may be overlap in the response exchange, we need to keep around the original responses section in the buffer (which probably will reside on memory), since they may theoretically be used more than 1 time in a different CBOR interpretation. It is non-trivial to compute the "done" part of the responses section which are not referenced in later exchanges to partially free the original responses bytes buffer.

## Non-streaming consumption of the small bundles
If the entire bundle is guaranteed to be a valid CBOR, it will become very easy to consume small packages which will easily fit in memory. If efficiency is not an issue, you can feed the entire package to CBOR parser, and consume the exchanges without implementing the full parsing algorithm.

For example, @irori created [a live web demo of unbundler](
https://bundle-load-test.surge.sh/) which assumes that the bundler output a CBOR array as response section.

# Proposals
- Specify the structure of the responses section
  - A valid CBOR array? 
- An index section entry references the response via array index, not offset
  - Enforces the start of the response is aligned to the start of the item in the responses CBOR array
  - Can save an alignment assert in the parser
- Require the responses section to be always after the index section

jyasskin:
[Preview](https://jyasskin.github.io/webpackage/ordered-responses/draft-yasskin-wpack-bundled-exchanges.html), [Diff](https://tools.ietf.org/rfcdiff?url1=https://wicg.github.io/webpackage/draft-yasskin-wpack-bundled-exchanges.txt&url2=https://jyasskin.github.io/webpackage/ordered-responses/draft-yasskin-wpack-bundled-exchanges.txt)

This lets us store only lengths in the index, and compute offsets from
them, which in turn enforces that the contents don't overlap and don't
have gaps, making the content array well-formed CBOR.

This applies to both the section index and the request index.

Fixes #212.

Let me know if you see any places that the parser still isn't checking that the bundle is well-formed CBOR.

### DIFF
--- a/draft-yasskin-wpack-bundled-exchanges.md
+++ b/draft-yasskin-wpack-bundled-exchanges.md
@@ -196,23 +196,19 @@ Future specifications can define new sections with extra data, and if necessary,
 these sections can be marked "critical" ({{critical-section}}) to prevent older
 parsers from using the rest of the bundle incorrectly.
 
-The bundle is roughly a CBOR item ({{?I-D.ietf-cbor-7049bis}}) with the
-following CDDL ({{?I-D.ietf-cbor-cddl}}) schema, but bundle parsers are required
-to successfully parse some byte strings that aren't valid CBOR. For example,
-sections might have padding between them, or even overlap, as long as the
-embedded relative offsets cause the parsing algorithms in this specification to
-return data.
+The bundle is a CBOR item ({{?I-D.ietf-cbor-7049bis}}) with the following CDDL
+({{?I-D.ietf-cbor-cddl}}) schema:
 
 ~~~~~ cddl
 webbundle = [
   ; 🌐📦 in UTF-8.
   magic: h'F0 9F 8C 90 F0 9F 93 A6',
-  section-offsets: bytes .cbor {* ($section-name .within tstr) =>
-                                  [ offset: uint, length: uint] },
+  section-lengths: bytes .cbor [* (section-name, length: uint) ],
   sections: [* $section ],
   length: bytes .size 8,  ; Big-endian number of bytes in the bundle.
 ]
 
+section-name = $section-name .within tstr
 $section-name /= "index" / "manifest" / "critical" / "responses"
 
 $section /= index / manifest / critical / responses
@@ -224,13 +220,12 @@ responses = [*response]
 ## Load a bundle's metadata {#load-metadata}
 
 A bundle holds a series of sections, which can be accessed randomly using the
-information in the `section-offset` CBOR item:
+information in the `section-lengths` CBOR item, which holds a list of
+alternating section names and section lengths:
 
 ~~~~~ cddl
-section-offsets = {* tstr => [ offset: uint, length: uint] },
+section-lengths = [* (section-name, length: uint) ],
 ~~~~~
-
-Offsets in this item are relative to the *end* of the section-offset item.
 
 To implement {{semantics-load-metadata}}, the parser MUST run the following
 steps, taking the `stream` as input.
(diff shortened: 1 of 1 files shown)


nyaxt:
I love this change. Looks great on a first pass. I'll take a closer look later today.

draft-yasskin-wpack-bundled-exchanges.md
```diff
@@ -336,13 +370,16 @@ the parser MUST do the following:
       * header list is `headers`, and
       * client is null.
 
-   1. Let `streamOffset` be `sectionsStart +
-      sectionOffsets["responses"].offset + offset`. That is, offsets in the
-      index are relative to the start of the "responses" section.
-   1. If `offset + length` is greater than
+   1. Let `streamOffset` be `sectionOffsets["responses"].offset +
+      currentOffset`. That is, offsets in the index are relative to the start of
+      the "responses" section.
```
nyaxt: Should we remove the sentence "That is, offsets in the index are relative to the start of the "responses" section."?

draft-yasskin-wpack-bundled-exchanges.md
```diff
@@ -295,28 +312,45 @@ steps, taking the `stream` as input.
...
+To parse the index section, given its `sectionContents`, the `sectionOffsets`
+map, and the `metadata` map to fill in, the parser MUST do the following:
 
 1. Let `index` be the result of parsing `sectionContents` as a CBOR item
    matching the `index` rule in the above CDDL ({{parse-cbor}}). If `index` is
    an error, return an error.
 
+1. Let `currentOffset` be a positive integer.
+
+1. Check that the responses array has the right number of items, and compute `currentOffset`:
```
nyaxt: Should we scope this step (+its substeps) to parsing the responses array header?

draft-yasskin-wpack-bundled-exchanges.md
```diff
@@ -295,28 +312,45 @@ steps, taking the `stream` as input.
...
+      fails, return an error.
+
+   1. Let (`responsesType`, `numResponses`) be the result of parsing the type
+      and argument of a CBOR item from the stream ({{parse-type-argument}}). If
+      this returns an error, return that error.
+
+   1. If `responsesType` is not `4` (a CBOR array) or `numResponses` is not half
+      of the length of `index`, return an error.
+
+   1. Set `currentOffset` to the current offset within `stream` minus
```
nyaxt: Then, we can have this say:
Let `currentOffset` be the ...

without Step 2.

draft-yasskin-wpack-bundled-exchanges.md
jyasskin: Sure.

draft-yasskin-wpack-bundled-exchanges.md
jyasskin: Whoops, done.

PR #228, opened 2018-06-21
Limit the allowed schemes for URLs.

### LINKED ISSUE #222: Clarify what schemes should be allowed for URL fields in the signature
Current Signed Exchange spec draft [says](https://wicg.github.io/webpackage/draft-yasskin-http-origin-signed-responses.html#rfc.section.3.1) `cert-url` and `validity-url` should be [an absolute-URL string](https://url.spec.whatwg.org/#absolute-url-string), but that could potentially allow multiple schemes than http/https (e.g. file and ws), while it doesn't seem to include data: URL.  It looks we'd better clarify what schemes should be allowed there.

jyasskin:
And clarify the definition of "absolute URL".

Fixes #222.

@kinu 

[Preview](https://jyasskin.github.io/webpackage/filter-schemes/draft-yasskin-http-origin-signed-responses.html), [Diff](https://tools.ietf.org/rfcdiff?url1=https://wicg.github.io/webpackage/draft-yasskin-http-origin-signed-responses.txt&url2=https://jyasskin.github.io/webpackage/filter-schemes/draft-yasskin-http-origin-signed-responses.txt)

### DIFF
--- a/draft-yasskin-http-origin-signed-responses.md
+++ b/draft-yasskin-http-origin-signed-responses.md
@@ -124,7 +124,15 @@ will provide a way to group signed exchanges into bundles that can be
 transmitted and stored together, but single signed exchanges are useful enough
 to standardize on their own.
 
-# Terminology
+# Terminology {#terminology}
+
+Absolute URL
+: A string for which the [URL
+ parser](https://url.spec.whatwg.org/#concept-url-parser) ({{URL}}), when run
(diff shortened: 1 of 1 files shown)


kinu:
Thanks! lgtm.

jyasskin:
@twifkak I should have double-checked with you before merging this. Let me know if I've gotten the right set of schemes in each location.

twifkak:
No worries; this LGTM.

PR #230, opened 2018-06-22
Introduce dump-bundle tool and follow #226 format

nyaxt:


go/bundle/cmd/dump-bundle/main.go
```diff
@@ -0,0 +1,51 @@
...
+	b, err := ReadBundleFromFile(*flagInput)
+	if err != nil {
+		return err
+	}
+
+	dopts := signedexchange.DumpOptions{
+		DumpContentText: *flagDumpContentText,
+	}
+	for _, e := range b.Exchanges {
+		err := e.Dump(os.Stdout, dopts)
```
hajimehoshi: `if err := ... ; err {`

go/signedexchange/signedexchange.go
```diff
@@ -51,12 +51,69 @@ func NewExchange(uri *url.URL, requestHeaders http.Header, status int, responseH
...
 // The caller should not modify the returned content.
 func (e *Exchange) Payload() []byte {
 	return e.payload
 }
 
+type DumpOptions struct {
+	DumpContentText bool
+}
+
+func (e *Exchange) Dump(w io.Writer, opts DumpOptions) error {
```
hajimehoshi: Why do we need to introduce a struct for a option? Isn't just an argument enough so far? My expectation is `Dump`'s signature is changeable though `Dump` is exported.

go/bundle/cmd/dump-bundle/main.go
nyaxt: Done

go/bundle/bundle.go
```diff
@@ -10,24 +10,120 @@ import (
...
 
 var HeaderMagicBytes = []byte{0x84, 0x48, 0xf0, 0x9f, 0x8c, 0x90, 0xf0, 0x9f, 0x93, 0xa6}
 
+type Request struct {
+	*url.URL
+	http.Header
+}
+
+type Response struct {
+	Status string
```
hajimehoshi: As HTTP status is represented as an integer in `net/http` package, let's use an integer here.

PR #232, opened 2018-06-23
Start the -impl-01 draft.

### LINKED ISSUE #186: Increment ver num on "implementation checkpoint"
Chromium would like to start adding support for next "implementation checkpoint".
Can we increment ver num on https://wicg.github.io/webpackage/draft-yasskin-httpbis-origin-signed-exchanges-impl.html#rfc.section.5.3 ?

jyasskin:
I think this catches everything that we're expecting to be different or missing in the Chrome 69 implementation vs the main spec. Let me know if I've missed anything.

Fixes #186.

### DIFF
--- a/draft-yasskin-httpbis-origin-signed-exchanges-impl.md
+++ b/draft-yasskin-httpbis-origin-signed-exchanges-impl.md
@@ -481,13 +481,11 @@ to retrieve an updated OCSP from the original server.
    anyone uses the same key in a TLS certificate and an exchange-signing
    certificate.
    1. A string that consists of octet 32 (0x20) repeated 64 times.
-   1. A context string: the ASCII encoding of "HTTP Exchange 1".
+   1. A context string: the ASCII encoding of "HTTP Exchange 1 b1".
 
-      Note: RFC EDITOR PLEASE DELETE THIS NOTE; The implementation of the final
-      RFC MUST use this context string, but implementations of drafts MUST NOT
-      use it and MUST use another draft-specific string beginning with "HTTP
-      Exchange 1 ” instead. This ensures that signers can predict how their
-      signatures will be used.
+      Note: As this is a snapshot of a draft of
+      {{?I-D.yasskin-http-origin-signed-responses}}, it uses a distinct context
+      string.
    1. A single 0 byte which serves as a separator.
    1. The bytes of the canonical CBOR serialization ({{canonical-cbor}}) of a
       CBOR map mapping:
@@ -965,14 +963,13 @@ and header fields, and a response payload.
 
 This content type consists of the concatenation of the following items:
 
-1. The ASCII characters "sxg1" followed by a 0 byte, to serve as a file
+1. The ASCII characters "sxg1-b1" followed by a 0 byte, to serve as a file
    signature. This is redundant with the MIME type, and recipients that receive
    both MUST check that they match and stop parsing if they don't.
 
-   Note: RFC EDITOR PLEASE DELETE THIS NOTE; The implementation of the final RFC
-   MUST use this file signature, but implementations of drafts MUST NOT use it
-   and MUST use another implementation-specific string beginning with "sxg1-" and
-   ending with a 0 byte instead.
+   Note: As this is a snapshot of a draft of
+   {{?I-D.yasskin-http-origin-signed-responses}}, it uses a distinct file
+   signature.
 1. 3 bytes storing a big-endian integer `sigLength`. If this is larger
    than 16kB, parsing MUST fail.
 1. 3 bytes storing a big-endian integer `headerLength`. If this is larger than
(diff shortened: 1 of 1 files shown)


nyaxt:
LGTM.

Should we put tentative values on TBD size limits?

draft-yasskin-httpbis-origin-signed-exchanges-impl.md
```diff
@@ -135,59 +149,63 @@ this validity information for some period of time.
 
 ## The Signature Header ## {#signature-header}
 
-The `Signature` header field conveys a single signature for an exchange,
-accompanied by information about how to determine the authority of and
+The `Signature` header field conveys a list of signatures for an exchange, each
+one accompanied by information about how to determine the authority of and
```
nyaxt: For M69, we currently plan to support only single signature for an exchange.

draft-yasskin-httpbis-origin-signed-exchanges-impl.md
```diff
@@ -135,59 +149,63 @@ this validity information for some period of time.
...
-{{I-D.ietf-httpbis-header-structure-02}}), and the list MUST contain exactly one
-element.
-
-Each parameterised label MUST have parameters named "sig", "integrity",
-"validityUrl", "date", and "expires". Each parameterised label MUST also have
-"certUrl" and "certSha256" parameters. This
-specification gives no meaning to the label itself, which can be used as a
-human-readable identifier for the signature (see {{parameterised-binary}}). The
-present parameters MUST have the following values:
+{{!I-D.ietf-httpbis-header-structure}}. Its value MUST be a parameterised list
```
nyaxt: [tangentially related]
While the header-structure spec's 04->06 diff doesn't affect signed exchange much, we were wondering if the header spec would also apply to "Content-Type" header someday, since it disallows identifier being an `item`.
With the change, `v=b1` is invalid and it has to be `v="b1"`.

When parsing signed exchange, Chrome Canary currently allows both `v=b1` and `v="b1"` so we are good here.
However, we might want to change the Accept header to say `v="b1"`

draft-yasskin-httpbis-origin-signed-exchanges-impl.md
```diff
@@ -369,86 +428,93 @@ Potentially-valid results include:
...
-         `main-certificate`'s public key. If the function is undefined on this
-         input, return "invalid".
+      1. If `publicKey` is an RSA key, return "invalid".
+      1. If `publicKey` is a key using the secp256r1 elliptic curve, set
+         `signing-alg` to ecdsa_secp256r1_sha256 as defined in Section 4.2.3 of
+         {{!I-D.ietf-tls-tls13}}.
+      1. Otherwise, either return "invalid" or set `signing-alg` to a non-legacy
+         signing algorithm defined by TLS 1.3 or later
+         ({{!I-D.ietf-tls-tls13}}). This choice MUST depend only on
+         `publicKey`'s type and not on any other context.
```
twifkak: 1. Does this mean the choice cannot depend on the signature_algorithms_cert extension?
2. Can the choice be "always return invalid"? (That appears to be what [chromium](https://cs.chromium.org/chromium/src/content/browser/web_package/signed_exchange_signature_verifier.cc?l=183&rcl=b7527b9711e28030e03ad438ba323af7142ce43b) does IIUC.)

(Pardon my naivety!)

draft-yasskin-httpbis-origin-signed-exchanges-impl.md
```diff
@@ -514,26 +579,39 @@ irrelevant fields omitted for ease of reading).
...
+  date=1511478660; expires=1511824260
 ~~~
 
-At 2017-11-27 11:02 UTC, `sig1` has expired, so the client needs to fetch
-`https://example.com/resource.validity.1511157180` (the `validityUrl` of `sig1`)
-to update that signatures. This URL might contain:
+At 2017-11-27 11:02 UTC, `sig1` and `sig2` have expired, but `thirdpartysig`
+doesn't exipire until 23:11 that night, so the client needs to fetch
+`https://example.com/resource.validity.1511157180` (the `validity-url` of `sig1`
+and `sig2`) to update those signatures. This URL might contain:
```
twifkak: s/to/if it wishes to/? (i.e. A client may use `thirdpartysig` without updating the others if it validates and meets its needs?)

draft-yasskin-httpbis-origin-signed-exchanges-impl.md
```diff
@@ -544,23 +622,137 @@ to update that signatures. This URL might contain:
...
+identifiers and parameters MUST be ignored because new identifiers and new
+parameters on existing identifiers may be defined by future specifications.
+
+### Integrity identifiers ### {#accept-signature-integrity}
+
+Identifiers starting with "mi/" indicate that the client supports the `MI`
+header field ({{!I-D.thomson-http-mice}}) with the parameter from the HTTP MI
+Parameter Registry registry named in lower-case by the rest of the identifier.
+For example, "mi/mi-blake2" indicates support for Merkle integrity with the
+as-yet-unspecified mi-blake2 parameter, and "-digest/mi-sha256" indicates
```
twifkak: s/digest/mi/

draft-yasskin-httpbis-origin-signed-exchanges-impl.md
```diff
@@ -637,71 +889,147 @@ described here.
...
+   Note that the use of the payload body here means that a `Transfer-Encoding`
+   header field inside the `application/signed-exchange` header block has no
+   effect. A `Transfer-Encoding` header field on the outer HTTP response that
+   transfers this resource still has its normal effect.
+
+### Cross-origin trust in application/signed-exchange {#co-trust-app-signed-exchange}
+
+To determine whether to trust a cross-origin exchange stored in an
+`application/signed-exchange` resource, pass the `Signature` header field from
+the non-signed header section and an exchange consisting of the headers from the
```
twifkak: Should this be updated to remove the reference to the "non-signed header section", or perhaps the above "`sigLength` bytes" item clarified to be named that?

draft-yasskin-httpbis-origin-signed-exchanges-impl.md
```diff
@@ -637,71 +889,147 @@ described here.
...
 
-The remainder of the resource is the exchange's payload, encoded with the
-`mi-sha256-draft2` content encoding. The `mi-sha256-draft2` content encoding is
-the `mi-sha256` encoding defined in draft 02 of ({{!I-D.thomson-http-mice}}). If
-the `mi-sha256-draft2` record length (the first 8 bytes of the payload) is
-greater than 16kB, or if any of the integrity proofs fail validation, parsing
-MUST fail.
+~~~
+sxg1\0<3-byte length of the following header
+value>sig1; sig=*...; integrity="mi"; ...<3-byte length of the encoding of the
```
twifkak: IIUC, the 3-byte header length should precede the signature value.

draft-yasskin-httpbis-origin-signed-exchanges-impl.md
```diff
@@ -775,7 +1143,7 @@ Additional information:
 
   Deprecated alias names for this type:  N/A
 
-  Magic number(s):  82 A?
+  Magic number(s):  73 78 67 31 00
```
twifkak: Should there be an implementation draft note for this, too?

twifkak:
LGTM

draft-yasskin-httpbis-origin-signed-exchanges-impl.md
twifkak: Out of curiosity, is there a specific "other context" that this language intends to prevent?

jyasskin: For example, lots of signature formats, like [JWS](https://tools.ietf.org/html/rfc7515#section-5.2) include an algorithm name in the signature in addition to a public key identifier. This allows some silly things like https://www.sjoerdlangkemper.nl/2016/09/28/attacking-jwt-authentication/.

PR #235, opened 2018-06-27
Fixes #234: mi-sha256 should be base64 encoded

### LINKED ISSUE #234: Bad mi-sha256 signature in example
## I expect

- mi-sha256 signatures to be base64 encoded

## instead

- in examples, they are hexdigest

ioggstream:
The examples show a mi-sha256 hexdigest instead of bas64 encoding,

see https://github.com/martinthomson/http-mice/blob/master/draft-thomson-http-mice.md

### DIFF
--- a/draft-yasskin-http-origin-signed-responses.md
+++ b/draft-yasskin-http-origin-signed-responses.md
@@ -331,7 +331,7 @@ Accept: */*
 
 HTTP/1.1 200
 Content-Type: text/html
-MI: mi-sha256=20addcf7368837f616d549f035bf6784ea6d4bf4817a3736cd2fc7a763897fe3
+MI: mi-sha256=dcRDgR2GM35DluAV13PzgnG6-pvQwPywfFvAu1UeFrs==
 Signed-Headers: "content-type", "mi"
 
 <!doctype html>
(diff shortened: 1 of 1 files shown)


draft-yasskin-http-origin-signed-responses.md
```diff
@@ -331,7 +331,7 @@ Accept: */*
 
 HTTP/1.1 200
 Content-Type: text/html
-MI: mi-sha256=20addcf7368837f616d549f035bf6784ea6d4bf4817a3736cd2fc7a763897fe3
+MI: mi-sha256=dcRDgR2GM35DluAV13PzgnG6-pvQwPywfFvAu1UeFrs==
```
nyaxt: I was wondering if this should be
```
MI: mi-sha256=*dcRDgR2GM35DluAV13PzgnG6-pvQwPywfFvAu1UeFrs==*
```

but it looks like https://martinthomson.github.io/http-mice/draft-thomson-http-mice.html#header refers to RFC7230 and RFC7231, not https://tools.ietf.org/html/draft-ietf-httpbis-header-structure-06 yet.

ioggstream: @nyaxt right, MICE doesn't use structured headers.

OT: where can I ask questions/discuss on other parts of this proposal?

draft-yasskin-http-origin-signed-responses.md
```diff
@@ -331,7 +331,7 @@ Accept: */*
 
 HTTP/1.1 200
 Content-Type: text/html
-MI: mi-sha256=20addcf7368837f616d549f035bf6784ea6d4bf4817a3736cd2fc7a763897fe3
```
irori: https://martinthomson.github.io/http-mice/draft-thomson-http-mice.html#mi-header-field-parameters says that the value is base64url encoding [[RFC7515](https://tools.ietf.org/html/rfc7515)], which does not have padding '=' characters.

ioggstream: @irori thx++ :D

draft-yasskin-http-origin-signed-responses.md
jyasskin: OT: Filing issues here is great. :)

PR #246, opened 2018-07-06
Several small fixes noticed by twifkak

jyasskin:
* Certificate chains have an array item before the first certificate.
* The first certificate's OCSP response was described as optional when
  it's not.
* The first certificate's SCT was *not* described as optional, but it
  can appear in the certificate itself or the OCSP response.
* "digest/mi-sha256" doesn't make sense.
* There's no longer a non-signed headers section in SXG files.
* An example had lengths in the wrong order.

[Preview](https://jyasskin.github.io/webpackage/devin-cleanups/draft-yasskin-http-origin-signed-responses.html), [Diff](https://tools.ietf.org/rfcdiff?url1=https://wicg.github.io/webpackage/draft-yasskin-http-origin-signed-responses.txt&url2=https://jyasskin.github.io/webpackage/devin-cleanups/draft-yasskin-http-origin-signed-responses.txt)

### DIFF
--- a/draft-yasskin-http-origin-signed-responses.md
+++ b/draft-yasskin-http-origin-signed-responses.md
@@ -375,19 +375,19 @@ cert-chain = [
 ]
 ~~~
 
-The first item in the CBOR array is treated as the end-entity certificate, and
-the client will attempt to build a path ({{?RFC5280}}) to it from a trusted root
-using the other certificates in the chain.
+The first map (second item) in the CBOR array is treated as the end-entity
+certificate, and the client will attempt to build a path ({{?RFC5280}}) to it
+from a trusted root using the other certificates in the chain.
 
 1. Each `cert` value MUST be a DER-encoded X.509v3 certificate ({{!RFC5280}}).
    Other key/value pairs in the same array item define properties of this
    certificate.
-1. The first certificate's `ocsp` value if any MUST be a complete, DER-encoded
-   OCSP response for that certificate (using the ASN.1 type `OCSPResponse`
-   defined in {{!RFC6960}}). Subsequent certificates MUST NOT have an `ocsp`
-   value.
-1. Each certificate's `sct` value MUST be a `SignedCertificateTimestampList` for
-   that certificate as defined by Section 3.3 of {{!RFC6962}}.
+1. The first certificate's `ocsp` value MUST be a complete, DER-encoded OCSP
+   response for that certificate (using the ASN.1 type `OCSPResponse` defined in
+   {{!RFC6960}}). Subsequent certificates MUST NOT have an `ocsp` value.
+1. Each certificate's `sct` value if any MUST be a
+   `SignedCertificateTimestampList` for that certificate as defined by Section
+   3.3 of {{!RFC6962}}.
 
 Loading a `cert-url` takes a `forceFetch` flag. The client MUST:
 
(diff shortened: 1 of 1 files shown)


draft-yasskin-http-origin-signed-responses.md
```diff
@@ -375,19 +375,19 @@ cert-chain = [
...
 1. Each `cert` value MUST be a DER-encoded X.509v3 certificate ({{!RFC5280}}).
    Other key/value pairs in the same array item define properties of this
    certificate.
-1. The first certificate's `ocsp` value if any MUST be a complete, DER-encoded
-   OCSP response for that certificate (using the ASN.1 type `OCSPResponse`
-   defined in {{!RFC6960}}). Subsequent certificates MUST NOT have an `ocsp`
-   value.
-1. Each certificate's `sct` value MUST be a `SignedCertificateTimestampList` for
-   that certificate as defined by Section 3.3 of {{!RFC6962}}.
+1. The first certificate's `ocsp` value MUST be a complete, DER-encoded OCSP
```
jyasskin: @davidben I think this matches what we actually want (OCSP is required on 1st cert & forbidden on later ones; SCT is optional everywhere), but I want to double-check with you in case I'm forgetting something. https://github.com/WICG/webpackage/pull/122/files#r169830552 has our previous discussion on the topic.

jyasskin: @sleevi, David deferred to you on this. I'm finding myself a little uncomfortable having a different rule for SCTs on intermediate certs vs OCSP responses on intermediates, but I don't have a strong preference for which policy we have. I suspect it'll be easier to relax a rule that they're all forbidden on intermediates than to tighten a rule that they're allowed. Even if they're allowed in the format, of course, Chrome wouldn't need to process them.

sleevi: I'm not sure I understand the question. Are you asking whether it's ok to MUST NOT SCTs in subsequent certs, similar to OCSP? Or whether it's OK to relax the MUST NOT for OCSP? Or something different?

jyasskin: The question is basically, "Is the new text right, and if not, what should I say instead?"

sleevi: So, for the normative changes:
* OCSP now required, at least for the first cert (if any -> MUST)
* CT now optional (MUST -> if any)

If so, that's correct.

As to any asymmetry between the two:
OCSP on every cert is similar to OCSP Multi-stapling ( https://tools.ietf.org/html/rfc6961 ) in <= TLS 1.2 or the TLS 1.3 Certificate frame ( https://tools.ietf.org/html/draft-ietf-tls-tls13-28#section-4.4.2 ), for which the Extensions structure allows an OCSP response or an SCT for each individual certificate presented in the handshake. In those cases, both are optional (e.g. a TLS1.2 server could only support status_request extension), but allowed for every cert.

This is slightly different - OCSP is required by policy for the first cert, and forbidden for subsequent certs. I could see the MUST NOT be reduced to a SHOULD NOT in the cases where a browser's whitelist might not handle it (e.g. technically constrained sub-CA certs), but, meh.

So... Yes, the change and motivation look good. I could also see explicitly forbidding SCTs on subsequent certs to parallel the OCSP case, for the same reasons and motivations.

jyasskin: Ok, thanks. I'll leave the new text as I have it since you don't say it's important for the two to be symmetric.

PR #247, opened 2018-07-06
Make the CanSignHttpExchanges extension non-critical.

### LINKED ISSUE #238: Critical-ness of CanSignHttpExchanges extension
This came up in https://crrev.com/c/1111485.

https://wicg.github.io/webpackage/draft-yasskin-http-origin-signed-responses.html#cross-origin-cert-req

> Conforming CAs MUST mark this extension as critical, and clients MUST NOT accept certificates with this extension in TLS connections (Section 4.4.2.2 of [I-D.ietf-tls-tls13]). 

In Chrome, certificate path building and verification is performed using underlying operating systems' API. So, if the CanSignHttpExchanges extension is marked as critical, the OS API will reject the cert, until the OS supports CanSignHttpExchanges extension.

Can we revisit the critical-ness of this extension? What's the security implication of not marking it as critical?

cc: @sleevi @jyasskin

jyasskin:
But warn servers away from using it in TLS servers.

Fixes #238.

@sleevi and any CA folks who see this, do you think I'm right to say "Conforming CAs MUST NOT mark this extension as critical.", or should I leave them without a requirement here?

Spec: [Preview](https://jyasskin.github.io/webpackage/non-critical-extension/draft-yasskin-http-origin-signed-responses.html), [Diff](https://tools.ietf.org/rfcdiff?url1=https://wicg.github.io/webpackage/draft-yasskin-http-origin-signed-responses.txt&url2=https://jyasskin.github.io/webpackage/non-critical-extension/draft-yasskin-http-origin-signed-responses.txt)

Impl draft: [Preview](https://jyasskin.github.io/webpackage/non-critical-extension/draft-yasskin-httpbis-origin-signed-exchanges-impl.html), [Diff](https://tools.ietf.org/rfcdiff?url1=https://wicg.github.io/webpackage/draft-yasskin-httpbis-origin-signed-exchanges-impl.txt&url2=https://jyasskin.github.io/webpackage/non-critical-extension/draft-yasskin-httpbis-origin-signed-exchanges-impl.txt)

### DIFF
--- a/draft-yasskin-http-origin-signed-responses.md
+++ b/draft-yasskin-http-origin-signed-responses.md
@@ -897,10 +897,14 @@ able to make even one unauthorized signature.
 Certificates with this extension MUST be revoked if an unauthorized entity is
 able to make even one unauthorized signature.
 
-Conforming CAs MUST mark this extension as critical, and clients MUST NOT accept
-certificates with this extension in TLS connections (Section 4.4.2.2 of
-{{!I-D.ietf-tls-tls13}}).  This simplifies security analysis of this protocol
-and avoids encouraging server operators to put exchange-signing keys on servers
+Conforming CAs MUST NOT mark this extension as critical. This allows new clients
+with old operating systems to handle the extension.
+
+Clients MUST NOT accept certificates with this extension in TLS connections
+(Section 4.4.2.2 of {{!I-D.ietf-tls-tls13}}).  This discourages server operators
+from using the same certificate to sign both TLS connections and signed
+exchanges, which in turn simplifies security analysis of this protocol and
+avoids encouraging server operators to put exchange-signing keys on servers
 exposed directly to the internet.
 
 RFC EDITOR PLEASE DELETE THE REST OF THE PARAGRAPHS IN THIS SECTION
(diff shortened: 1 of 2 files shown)


draft-yasskin-http-origin-signed-responses.md
```diff
@@ -897,10 +897,14 @@ able to make even one unauthorized signature.
 Certificates with this extension MUST be revoked if an unauthorized entity is
 able to make even one unauthorized signature.
 
-Conforming CAs MUST mark this extension as critical, and clients MUST NOT accept
-certificates with this extension in TLS connections (Section 4.4.2.2 of
-{{!I-D.ietf-tls-tls13}}).  This simplifies security analysis of this protocol
-and avoids encouraging server operators to put exchange-signing keys on servers
+Conforming CAs MUST NOT mark this extension as critical. This allows new clients
+with old operating systems to handle the extension.
```
sleevi: I don't think we'd want to include the second sentence in the spec. That is, it presumes that new clients use the OS verifier, or that they even have an OS verifier, which... isn't going to hold :)

draft-yasskin-http-origin-signed-responses.md
```diff
@@ -897,10 +897,14 @@ able to make even one unauthorized signature.
...
+
+Clients MUST NOT accept certificates with this extension in TLS connections
+(Section 4.4.2.2 of {{!I-D.ietf-tls-tls13}}).  This discourages server operators
+from using the same certificate to sign both TLS connections and signed
+exchanges, which in turn simplifies security analysis of this protocol and
+avoids encouraging server operators to put exchange-signing keys on servers
```
sleevi: I think the motivation might be better detailed in security considerations.

The reason this is a bit tricky is that by virtue of non-critical, clients that don't recognize this extension WILL accept this extension in TLS connections. So we don't have a hard security boundary. Because of this motivation, it seems like it makes sense to be discussed in the security section, calling out that this is not a security boundary itself, but simply an ecosystem health thing.

PR #254, opened 2018-07-13
gen-certurl: Fetch OCSP response from network if -ocsp is omitted

irori:


nyaxt:
- Would you add test for this? (separate PR is fine)
- Can we dedupe pem parsing? (separate PR is fine)

irori:
PTAL

Test added. Let me do pem cleanup in a separate PR.

PR #261, opened 2018-07-18
Increase the header size limit to 512kB.

jyasskin:
This is twice Chromium's limit for the request or response headers,
since both are included in the signed exchange headers block.


nyaxt:
Thanks. This matches M69 now.

PR #264, opened 2018-07-19
signedexchange: Rename some functions

hajimehoshi:
Change the function to a method to follow the same way as net/http.Request.Write.

### DIFF
--- a/go/signedexchange/signedexchange.go
+++ b/go/signedexchange/signedexchange.go
@@ -231,7 +231,7 @@ func (e *Exchange) decodeExchangeHeaders(dec *cbor.Decoder) error {
 }
 
 // draft-yasskin-http-origin-signed-responses.html#application-http-exchange
-func WriteExchangeFile(w io.Writer, e *Exchange) error {
+func (e *Exchange) Write(w io.Writer) error {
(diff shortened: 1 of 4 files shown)


go/signedexchange/signedexchange.go
```diff
@@ -243,7 +243,7 @@ func (e *Exchange) decodeExchangeHeaders(dec *cbor.Decoder) error {
 }
 
 // draft-yasskin-http-origin-signed-responses.html#application-http-exchange
-func WriteExchangeFile(w io.Writer, e *Exchange) error {
+func (e *Exchange) Write(w io.Writer) error {
```
nyaxt: @twifkak wdyt?

Please also rename `ReadExchangeFile` if we are going to proceed.

twifkak: LGTM. I'm not well-acquainted with idiomatic Go style, so I'll defer to your judgment there. 

Thanks for the heads up! ampproject/amppackager still depends on nyaxt/webpackage, so there's no breakage by this change, but I'll keep it in mind for when I fix that.

> ampproject/amppackager still depends on nyaxt/webpackage, so there's no breakage by this change, but I'll keep it in mind for when I fix that.

Sure!

Issue #270, opened 2018-07-23
 "manifest" section URL should first look if the resource is included in the bundle


nyaxt:
https://wicg.github.io/webpackage/draft-yasskin-wpack-bundled-exchanges.html#manifest-section

PR #271, opened 2018-07-25
Always embed manifest in the bundle

nyaxt:
Addresses #270

After giving more thought, I couldn't think of a use-case where we want to refer to a external manifest.

On the other hand, I also want to make the manifest URL optional, for use-cases where a bundle contains a part of a page resources (for example, common assets across multiple webpages, a ES module tree, etc.)

### DIFF
--- a/draft-yasskin-wpack-bundled-exchanges.md
+++ b/draft-yasskin-wpack-bundled-exchanges.md
@@ -383,7 +383,10 @@ map, and the `metadata` map to fill in, the parser MUST do the following:
 ### Parsing the manifest section {#manifest-section}
 
 The "manifest" section records a single URL identifying the manifest of the
-bundle. The bundle can contain multiple resources at this URL, and the client is
+bundle. The URL MUST refer to the one or more response(s) contained in the
+bundle itself.
+
+The bundle can contain multiple resources at this URL, and the client is
 expected to content-negotiate for the best one. For example, a client might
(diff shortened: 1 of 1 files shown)


draft-yasskin-wpack-bundled-exchanges.md
```diff
@@ -383,7 +383,10 @@ map, and the `metadata` map to fill in, the parser MUST do the following:
 ### Parsing the manifest section {#manifest-section}
 
 The "manifest" section records a single URL identifying the manifest of the
-bundle. The bundle can contain multiple resources at this URL, and the client is
+bundle. The URL must refer to the one or more response(s) contained in the
```
jyasskin: "MUST" should be capitalized per https://tools.ietf.org/html/rfc8174.

jyasskin:
I made the manifest mandatory so that a start_url would be available for when someone navigated to a bundle. Once we move the fallback URL out of the manifest, it serves that purpose, and I think it makes sense to make the manifest optional.

draft-yasskin-wpack-bundled-exchanges.md
nyaxt: Done.

nyaxt:
> Once we move the fallback URL out of the manifest, it serves that purpose, and I think it makes sense to make the manifest optional.

SGTM

PR #278, opened 2018-07-30
gen-bundle: Drop pseudo headers in HAR

irori:
Chromium generates HAR that includes HTTP/2 pseudo headers in the
`request.headers` field. Drop them so that they don't conflict with
Bundled Exchanges' pseudo headers.


go/bundle/cmd/gen-bundle/main.go
```diff
@@ -43,6 +43,10 @@ func ReadHarFromFile(path string) (*hargo.Har, error) {
 func nvpToHeader(nvps []hargo.NVP, isStatefulHeader func(string) bool) (http.Header, error) {
 	h := make(http.Header)
 	for _, nvp := range nvps {
+		// Drop HTTP/2 pseudo headers.
+		if len(nvp.Name) > 0 && nvp.Name[0] == ':' {
```
nyaxt: if strings.HasPrefix(nvp.Name, ":")

irori: Done

Issue #276, opened 2018-07-30
Make it easy to reuse the encoded header cbor when verifying signature


nyaxt:
We might want to adjust the signature message format so that it is easy to reuse the encoded headers when verifying the signature.
The `headers` required in Section 3.5 step 7.4.5 is a copy of the cbor appearing earlier. However, the fact that it appears inline in the map makes it hard for many cbor encoder implementation to reconstruct the cbor representation of the map without decode(encode(headers)), which is non-secure.

More discussion on the [crbug](https://bugs.chromium.org/p/chromium/issues/detail?id=863499)

jyasskin:
http://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/jyasskin/webpackage/loading-spec/loading.bs#signed-message has an in-progress specification of how to build the signed message without re-encoding. I have no problem with making that easier.

nyaxt:
I'm happy with your proposal.

Just to enumerate alternatives:
- Move `headers` out of the map #279
    - slightly easier to implement, since we don't have to actually concat everything but construct hash func state but `headers` first, then progress the state with `headers` and then emit the final result
- Don't use CBOR to encode {date, expires, headers, cert-sha256, validity-url}
    - 64bit big-endian for date, expires
    - bytes dump for cert-sha256, validity-url

IIUC, the sole purpose of the `message` is to be used as input to signing algorithm, so we may not have to worry about its parsability so much and just focus on ease of its generation.

jyasskin:
The CBOR map was originally designed to be clear about version changes, but that's now handled by the versioned context string. Given that, a pure concatenation of the input (making sure everything's length-prefixed or fixed length) may be better.

PR #279, opened 2018-07-31
Move headers out of the CBOR map when constructing signed message bytes.

nyaxt:
To address #276


jyasskin:
I'd like to get @davidben's thoughts on any rework we do here, with reference to the [draft loading spec](http://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/jyasskin/webpackage/loading-spec/loading.bs#signed-message) that specifies how to check this.

He's out until mid-next-week, at which point I might be on leave, but if he's happy with a change while I'm out, go ahead and make it.

nyaxt:
I don't have a strong preference on how we should fix this.
@davidben Please also take a look at options being discussed at https://github.com/WICG/webpackage/issues/276 . We are happy to hear other serialization method ideas too.

nyaxt:
We will go forward with #287 instead.

PR #281, opened 2018-08-06
Integrate b1 signed exchanges with Fetch

jyasskin:
[Preview](https://jyasskin.github.io/webpackage/loading-spec/loading.html)

### DIFF
--- a/loading.bs
+++ b/loading.bs
@@ -273,6 +273,20 @@ An exchange is a [=struct=] with the following items:
 
 </ul>
 
+<h3 dfn-type=dfn>Read buffer</h3>
+
+A read buffer is a [=struct=] with the following items:
+
+<ul dfn-for="read buffer">
+
+* <dfn>stream</dfn>: A {{ReadableStream}}.
+* <dfn>reader</dfn>: A {{ReadableStreamDefaultReader}} whose stream is [=read
+    buffer/stream=].
+* <dfn>bytes</dfn>: A [=byte sequence=] holding the bytes that have been read
+    from [=read buffer/stream=] but not returned to a calling algorithm yet.
+
+</ul>
+
 <h3 dfn-type=dfn>Augmented Certificate</h3>
 
 An augmented certificate is a [=tuple=] with the following items:
@@ -748,20 +762,6 @@ A [=request=] |browserRequest| <dfn>matches the stored request</dfn>
 Issue(whatwg/fetch#730): The algorithms in this section create and operate over
 ECMAScript objects like {{ReadableStream}} despite not having a Realm to attach
 them to.
-
-A <dfn>read buffer</dfn> is a [=struct=] with the following items:
-
-<ul dfn-for="read buffer">
-
-* <dfn>stream</dfn>: A {{ReadableStream}}.
-
-* <dfn>reader</dfn>: A {{ReadableStreamDefaultReader}} whose stream is [=read
-    buffer/stream=].
-
-* <dfn>bytes</dfn>: A [=byte sequence=] holding the bytes that have been read
-    from [=read buffer/stream=] but not returned to a calling algorithm yet.
-
-</ul>
 
 <h4 algorithm id="create-read-buffer">Create a read buffer</h4>
 
--- a/loading.bs
+++ b/loading.bs
@@ -644,27 +644,27 @@ The [=certificate chain=] |chain| <dfn for="certificate chain" lt="has a trusted
 leaf|does not have a trusted leaf">has a trusted leaf</dfn> for an [=origin=]
 |origin| if the following steps return `trusted`:
 
-1. Let |authority| be |origin|'s [=origin/host=].
 1. Let |leaf| be |chain|'s [=certificate chain/leaf=].
-1. Attempt to build a trustworthy path from |leaf| to a trusted root using
-    [[!RFC5280]] and any other conventions used in making TLS ([[!RFC8446]])
-    connections. If no such path can be built, return `untrusted`.
+1. Attempt to build a trustworthy path from |leaf|'s [=augmented
+    certificate/certificate=] to a trusted root with
+
+    * |chain|,
+    * |leaf|'s [=augmented certificate/OCSP response=],
+    * any SCTs from
+        * |leaf|'s [=augmented certificate/SCT=].
+        * An OCSP extension in |leaf|'s [=augmented certificate/OCSP response=].
+        * An X.509 extension in |leaf|'s [=augmented certificate/certificate=].
+
+    as input, using [[!RFC5280]] and any other conventions used in making TLS
+    ([[!RFC8446]]) connections. The UA MUST support Certificate Transparency
+    ([[RFC6962]]) for this check. The UA MUST check that it has evidence the
+    |leaf|'s [=augmented certificate/certificate=] was not revoked more than 7
+    days ago (for example using the |leaf|'s [=augmented certificate/OCSP
+    response=]). If no such path can be built, return `untrusted`.
 1. If |leaf|'s [=augmented certificate/certificate=] is not trusted for
-    |authority|, return `untrusted`.
+    |origin|'s [=origin/host=], return `untrusted`.
 1. If |leaf|'s [=certificate/extensions=] don't map a [=CanSignHttpExchanges=]
     OID to the ASN.1/DER encoding of NULL (0x05 0x00), return `untrusted`.
-1. If |leaf| has no [=augmented certificate/OCSP response=] or this response has
-    a lifetime of more than 7 days or isn't valid for |leaf|'s [=augmented
-    certificate/certificate=], return `untrusted`.
-
-    Note: This does not check for revocation of intermediate certificates, and
-    clients SHOULD implement another mechanism for that.
-
-1. If |leaf| doesn't have enough SCTs from trusted logs return `untrusted`,
-    where SCTs can be gathered from:
-    * |leaf|'s [=augmented certificate/SCT=].
-    * An OCSP extension in |leaf|'s [=augmented certificate/OCSP response=].
-    * An X.509 extension in |leaf|'s [=augmented certificate/certificate=].
 1. Return `trusted`.
 
 <h3 algorithm id="parse-cbor-headers">Parsing CBOR headers</h3>
--- a/loading.bs
+++ b/loading.bs
@@ -271,6 +271,13 @@ add the following steps:
 
         then set |response| to |httpRequest|'s [=request/stashed exchange=]'s
         [=exchange/response=].
+    1. If |response| is null and |httpRequest|'s [=request/initiator=] is
+        "`prefetch`" or "`preload`", return a [=network error=].
+
+        Note: This ensures that prefetching a signed exchange from one origin
+        won't accidentally do a network request from another origin, which could
+        [compromise the user's
+        privacy](https://wicg.github.io/webpackage/draft-yasskin-webpackage-use-cases.html#private-prefetch).
 
 Note: Applying the signed exchange's response here has the effect of letting a
 newer HTTP cache entry override a signed exchange's content, and of not storing
--- a/loading.bs
+++ b/loading.bs
@@ -412,6 +412,12 @@ following steps:
     either "`deprecated`" or "`modern`".
 
     Note: See <a spec="fetch">HTTP-network fetch</a> for details of this choice.
+1. If |parsedExchange|'s [=exchange/response=]'s [=response/status=] is a
+    [=redirect status=] or the [=signed exchange version=] of |parsedExchange|'s
+    [=exchange/response=] is not undefined, return a failure.
+
+    Note: This might simplify the UA's implementation, since it doesn't have to
+    handle nested signed exchanges.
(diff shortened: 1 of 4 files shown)


loading.bs
```diff
@@ -0,0 +1,949 @@
...
+signatures, and then if all is well, redirects to its request URL
+(`https://publisher.example.org/foo`) with a "stashed" exchange attached to the
+request. The redirect applies all the usual processing, and then when it would
+normally check for an HTTP cache hit, it also checks whether the stashed request
+matches the redirected request and which of the stashed exchange or HTTP cache
+contents is newer. If the stashed exchange matches and is newer, the UA returns
+the stashed response.
+
+A Service Worker for `https://distributor.example.org/` gets to handle the
+original request. A Service Worker for `https://publisher.example.org/` does not
```
jyasskin: @kinu Have y'all looked into the Service Worker side? I believe Fetch's https://fetch.spec.whatwg.org/#concept-http-fetch step 4.2

> If request’s redirect mode is "follow", then set request’s service-workers mode to "none".

prevents all redirects from hitting a service worker, so even if distributor.example.org just served a 302 to publisher.example.org, publisher.example.org's SW wouldn't get to handle the navigation. Am I misreading anything?

loading.bs
```diff
@@ -0,0 +1,949 @@
...
+    skew, those clients will reject the signature.
+1. If the UA's estimate of the current time is after |signature|'s [=exchange
+    signature/expiration time=], return "untrusted".
+
+    Note: We use the client's best guess of the current time to check the
+    [=exchange signature/expiration time=] so that attackers trying to get an
+    exchange trusted for longer, are constrained to modify the client's clock
+    and can't also attack its estimate of its skew.
+1. Let |message| be the [=b1 signed message=] for |signature| and |headerBytes|.
+1. Let |publicKey| be the [=certificate/public key=] of |parsedSignature|'s
```
jyasskin: @sleevi @davidben The fact that the [certificate](http://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/jyasskin/webpackage/loading-spec/loading.bs#augmented-certificate-certificate) is stored un-parsed means that the spec winds up saying to parse out a public key twice, which feels like a security risk. How much can systems actually parse these structures ahead of time, vs when they're validating origin-trust?

loading.bs
```diff
@@ -0,0 +1,949 @@
...
+
+    Note: The final [[draft-yasskin-http-origin-signed-responses]] will use a
+    version of `` `1` ``, but this specification tracks what's actually
+    implemented in browsers, which still uses draft versions.
+
+<h3 algorithm id="mp-http-network-or-cache-fetch">Monkeypatch HTTP-network-or-cache fetch</h3>
+
+In <a spec="fetch">HTTP-network-or-cache fetch</a>, after
+
+> 5.19. If |httpRequest|’s [=request/cache mode=] is neither "`no-store`" nor
```
jyasskin: Right now, I'm ignoring all of the [cache modes](https://fetch.spec.whatwg.org/#concept-request-cache-mode) in deciding to accept a signed exchange's content. They may be useful in controlling how to fetch the `.sxg` itself, and applying them to the content would probably just have the effect of forcing that to the network, which would defeat the point.

Does that sound right to y'all, or do you see any of the modes I should apply here?

loading.bs
```diff
@@ -0,0 +1,949 @@
...
+|exchange| if the following steps return "trusted":
+
+1. Let |requestUrl| be |exchange|'s [=exchange/request=]'s [=request/url=].
+1. If |signature|'s [=exchange signature/validityUrl=]'s [=url/origin=] is not
+    [=same origin=] with |requestUrl|'s [=url/origin=], return "untrusted".
+1. If |exchange|'s [=exchange/request=]'s [=request/header list=] includes a
+    [=stateful request header=] or |exchange|'s [=exchange/response=]'s
+    [=response/header list=] includes a [=stateful response header=], return
+    "untrusted".
+1. If |signature|'s [=exchange signature/expiration time=] is more than 604800
```
jyasskin: I check the length of the validity window in cross-origin trust rather than while parsing the signature so that other uses of signatures can define their own maximum validities. e.g. maybe the Play store wants to sign that it's audited an app and have that be valid for more than a week.

loading.bs
```diff
@@ -0,0 +1,949 @@
...
+1. If |leaf|'s [=certificate/extensions=] don't map a [=CanSignHttpExchanges=]
+    OID to the ASN.1/DER encoding of NULL (0x05 0x00), return `untrusted`.
+1. If |leaf| has no [=augmented certificate/OCSP response=] or this response has
+    a lifetime of more than 7 days or isn't valid for |leaf|'s [=augmented
+    certificate/certificate=], return `untrusted`.
+
+    Note: This does not check for revocation of intermediate certificates, and
+    clients SHOULD implement another mechanism for that.
+
+1. If |leaf| doesn't have enough SCTs from trusted logs return `untrusted`,
```
jyasskin: @sleevi Is this a reasonable way to phrase the requirement?

loading.bs
```diff
@@ -0,0 +1,949 @@
...
+    |storedRequest|'s [=request/url=], return "mismatch".
+
+1. Issue: Flesh this out, along the lines of
+    https://docs.google.com/document/d/1JdJkdY7cK2rD9JXbK75sSn069c0rhF9UnLQACBZ1OtY/edit.
+
+
+1. Return "match".
+
+
+## Stream algorithms ## {#stream-algs}
```
jyasskin: @domenic This [whole section](http://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/jyasskin/webpackage/loading-spec/loading.bs#stream-algs) could migrate to either Fetch or Streams if you think it'll be useful to other specs.

loading.bs
```diff
@@ -0,0 +1,949 @@
...
+    format to avoid cross- protocol attacks if anyone uses the same key in a TLS
+    certificate and an exchange-signing certificate.
+1. `` `HTTP Exchange 1 b1` `` to serve as a context string.
+
+    Note: Each draft of
+    [[draft-yasskin-httpbis-origin-signed-exchanges-impl-01]] and the final RFC
+    for [[draft-yasskin-http-origin-signed-responses]] will use distinct context
+    strings.
+1. A single 0x00 byte which serves as a separator.
+1. The concatenation of the following [=byte sequences=], which is the
```
jyasskin: @davidben How do you feel about this message algorithm? I think we can simplify it for b2, since I only made it this complicated to avoid confusion between multiple versions of the message, and that's now handled by the `HTTP Exchange 1 b1` context string, but it'll inform whether we can go to Origin Trial with b1 or have to go to b2 first.

loading.bs
kinu: For navigation the default redirect mode is "manual", therefore every time a request's redirected corresponding SW has a chance to see it.

loading.bs
```diff
@@ -0,0 +1,949 @@
...
+## Overview ## {#overview}
+
+When fetching a resource (`https://distributor.example.org/foo.sxg`) with the
+`application/signed-exchange` [=MIME type=], the UA parses it, checks its
+signatures, and then if all is well, redirects to its request URL
+(`https://publisher.example.org/foo`) with a "stashed" exchange attached to the
+request. The redirect applies all the usual processing, and then when it would
+normally check for an HTTP cache hit, it also checks whether the stashed request
+matches the redirected request and which of the stashed exchange or HTTP cache
+contents is newer. If the stashed exchange matches and is newer, the UA returns
```
kinu: This means that whenever UA loads things from a Signed Exchange UA also needs to check HTTP cache. Can you clarify what situation we are trying to deal with by this? If we don't cache the stashed exchange it feels it's just fine, or a site should probably set a short expiration date?

loading.bs
jyasskin: Aha! Thanks! 

I think both @cramforce's idea to use the `preloadResponse` and asking people to call `fetch()` with cache=="only-if-cached" will work, but to use "only-if-cached" we'll have to tell people to modify the original request or its clone, instead of creating a new request with the same URL. A new request with the same URL won't have the stashed exchange attached, and wiring those up will require an extra global map in the SW itself, which could also get confused by interleaved requests to different signed exchanges holding the same publication URL.

loading.bs
jyasskin: Imagine that Vashti visits https://app.com/ regularly, and app.com also publishes a signed exchange valid for a week to allow offline distribution. Now app.com discovers a vulnerability and patches it, but the vulnerable signed exchange is still out there.

If Vashti visits the site after the update and gets the new version into her cache, even stale, then even if an attacker manages to serve her the vulnerable signed exchange, she won't use it.

This isn't an absolute defense against anything, but it lowers the probability of a successful attack on frequent users of a site.

loading.bs
```diff
@@ -0,0 +1,965 @@
...
+<h3 algorithm id="validating-signature">Validating a signature</h3>
+
+An [=exchange signature=] |signature| <dfn for="exchange signature" lt="is
+valid|is not valid">is valid</dfn> for a [=byte sequence=] |headerBytes| if the
+following steps return `valid`:
+
+1. Let |clockSkew| be the uncertainty in the UA's estimate of the current time
+    caused by clock skew on the client. The UA MAY set this to 0 or use a more
+    sophisticated estimate.
+1. If the UA's estimate of the current time is more than |clockSkew| before
```
jyasskin: @davidben @sleevi How many hives does this bit give you?

sleevi: I'm not going to have bandwidth to review this this week, in part because I need to go get treated for hives. It quite concerns me, and I can tell it's going to take time and focus that I don't yet have to adequately review :/

loading.bs
sleevi: Can you explain why it feels like a security risk? I'm not sure I fully understand the question. Extracting an SPKI from a certificate is something easy on virtually all systems - the question is whether or not it's "trusted" data yet.

loading.bs
sleevi: No. There's no such thing as "trusted logs", and this also isn't how clients evaluate whether a certificate complies with a given CT policy.

I think much of this algorithm can be replaced by waving into the blackbox and saying that the OCSP response, the SCTs, and the certificates are inputs, for the host, and determine the response. I suspect you may want an added requirement on the OCSP responses freshness, or that the validating client MUST support Certificate Transparency (or is it optional for UAs that don't support CT but still support SXG?)

loading.bs
sleevi: I presume they're going to be using the same parser - is there reason to believe they may be different?

loading.bs
sleevi: You realize this means that only Chrome will implement SXGs for the foreseeable future? Coupling to CT in this way is going to either explicitly preclude certain UAs, or encourage them to adopt patterns that they've been reticent to do for TLS, precisely because they know it will harm the CT ecosystem. We should sync up on that. I realize the intersection between CT relates to SXG, but I think we need to be careful about MUSTing certain policy bits, compared to highlighting the tradeoffs.

loading.bs
jyasskin: No particular reason.

loading.bs
jyasskin: This is now a SHOULD with a link to a new Security Considerations section.

jyasskin:
Hooking into [HTTP fetch](https://jyasskin.github.io/webpackage/loading-spec/loading.html#mp-http-fetch) gives me an easy way to inject a redirect, but it means that signed exchanges loaded from the filesystem, from blobs, etc. don't work. In the long run, I think we'll want them to work, so we'll want to either move the attachment point up to [main fetch](https://fetch.spec.whatwg.org/#concept-main-fetch), or manually go through some of the redirect steps in the other branches of [scheme fetch](https://fetch.spec.whatwg.org/#concept-scheme-fetch).

I still think this PR is the right place to start: it'll be easier to see what that extension changes with this as a baseline.

nyaxt:
Given that other @jyasskin PRs build on top of this PR, I'd like to proceed with this and then revise the bits later. @kinu Are you cool with this?

loading.bs
kinu: Chrome's current impl is not following this, which is the biggest concern, but that part probably needs a separate discussion anyways to proceed. (I'll file an impl bug and take a closer look if this needs to block our further milestones)

loading.bs
kinu: This sg to me.

kinu:
Okay, let's merge it and consider separate PRs if we find any modifications are needed.

loading.bs
```diff
@@ -0,0 +1,992 @@
...
+
+<h3 algorithm id="request-matching">Request matching</h3>
+
+A [=request=] |browserRequest| <dfn>matches the stored request</dfn>
+|storedRequest| if the following steps return "match":
+
+1. If |browserRequest|'s [=request/url=] is not [=url/equal=] to
+    |storedRequest|'s [=request/url=], return "mismatch".
+
+1. Issue: Flesh this out, along the lines of
```
rwlbuis: Note that the way this is rendered in Safari looks a bit weird, I only see step 1. and 3. with a big ISSUE section in between. But I guess once Step 2 has been fleshed out, it will look ok.

PR #282, opened 2018-08-07
gen-bundle: Add checks to prevent creating invalid bundles

irori:
Drop the HAR entry if:
- Request method is not GET.
- Status code is not valid. Chromium puts status code 0 if the request
  failed with a network error.

### DIFF
--- a/go/bundle/cmd/gen-bundle/main.go
+++ b/go/bundle/cmd/gen-bundle/main.go
@@ -101,6 +101,15 @@ func fromHar(harPath string) error {
 			return fmt.Errorf("Failed to extract body from response content for the request %q. err: %v", e.Request.URL, err)
 		}
 
+		if e.Request.Method != "GET" {
+			log.Printf("Dropping the entry: non-GET request method (%s)", e.Request.Method)
+			continue
+		}
+		if http.StatusText(e.Response.Status) == "" {
+			log.Printf("Dropping the entry: unknown response status (%d)", e.Response.Status)
+			continue
+		}
+
 		e := &bundle.Exchange{
 			Request: bundle.Request{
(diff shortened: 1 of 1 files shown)


go/bundle/cmd/gen-bundle/main.go
```diff
@@ -101,6 +101,15 @@ func fromHar(harPath string) error {
 			return fmt.Errorf("Failed to extract body from response content for the request %q. err: %v", e.Request.URL, err)
 		}
 
+		if e.Request.Method != "GET" {
+			log.Printf("Dropping the entry: non-GET request method (%s)", e.Request.Method)
+			continue
+		}
+		if http.StatusText(e.Response.Status) == "" {
```
nyaxt: I'm not sure on this change. Would you provide more context?

nyaxt: Ah, how about simple check `Status != 0`?

nyaxt: or >=100 <=999

irori: Changed to check that status code is within [100, 999].

go/bundle/cmd/gen-bundle/main.go
```diff
@@ -101,6 +101,15 @@ func fromHar(harPath string) error {
 			return fmt.Errorf("Failed to extract body from response content for the request %q. err: %v", e.Request.URL, err)
 		}
 
+		if e.Request.Method != "GET" {
```
hajimehoshi: optional: `http.MethodGet`

irori: Done.

PR #285, opened 2018-08-08
Define request matching in terms of draft-ietf-httpbis-variants.

jyasskin:
See https://docs.google.com/document/d/1JdJkdY7cK2rD9JXbK75sSn069c0rhF9UnLQACBZ1OtY/edit for more analysis and options, but building on @mnot's Variants seems like a good plan here.

If we go this way, it probably makes sense to move away from storing arbitrary request headers in both signed exchanges and bundles, and toward storing something like the [Variants + Variant-Key headers](https://httpwg.org/http-extensions/draft-ietf-httpbis-variants.html#variants) directly.

***
<a href="https://pr-preview.s3.amazonaws.com/jyasskin/webpackage/pull/285.html" title="Last updated on Dec 7, 2018, 11:35 PM GMT (d1d5c8c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/webpackage/285/f603e91...jyasskin:d1d5c8c.html" title="Last updated on Dec 7, 2018, 11:35 PM GMT (d1d5c8c)">Diff</a>

### DIFF
--- a/loading.bs
+++ b/loading.bs
@@ -843,7 +843,6 @@ A [=request=] |browserRequest| <dfn>matches the stored request</dfn>
         defined by |name|'s content negotiation mechanism.
     1. Let |browserValue| be the result of [=header list/getting=] |name| from
         |browserRequest|'s [=request/header list=].
-    1. If |browserValue| is null, return "mismatch".
     1. If |variantMatches| returns an empty list when given a request header of
         |browserValue| and an available-values list consisting of
         «|storedValue|», return "mismatch".
--- a/loading.bs
+++ b/loading.bs
@@ -15,6 +15,15 @@ Assume Explicit For: yes
 </pre>
 <pre class='biblio'>
 {
+    "draft-ietf-httpbis-variants": {
+        "authors": [
+            "Mark Nottingham"
+        ],
+        "href": "https://httpwg.org/http-extensions/draft-ietf-httpbis-variants.html",
+        "title": "HTTP Representation Variants",
+        "status": "WD",
+        "publisher": "IETF"
+    },
     "draft-thomson-http-mice": {
         "authors": [
             "Martin Thomson"
@@ -89,6 +98,11 @@ spec: RFC7231; urlPrefix: https://tools.ietf.org/html/rfc7231#
         text: Date header field; url: section-7.1.1.2
 spec: RFC8446; urlPrefix: https://tools.ietf.org/html/draft-ietf-tls-tls13-28#
     text: ecdsa_secp256r1_sha256; type: dfn; url: section-4.2.3
+spec: draft-ietf-httpbis-variants; urlPrefix: https://httpwg.org/http-extensions/draft-ietf-httpbis-variants.html#
+    type: dfn
+        text: content negotiation mechanism specified to be usable with Variants; url: define
+        text: Variants cache behavior; url: cache
+        text: Variants header field; url: variants
 spec: draft-thomson-http-mice; urlPrefix: https://tools.ietf.org/html/draft-thomson-http-mice-02#
     type: dfn
         text: mi-sha256 parameter; url: section-3.1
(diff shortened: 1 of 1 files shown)


loading.bs
```diff
@@ -818,13 +831,23 @@ A [=request=] |browserRequest| <dfn>matches the stored request</dfn>
 1. If |browserRequest|'s [=request/url=] is not [=url/equal=] to
     |storedRequest|'s [=request/url=], return "mismatch".
 
-1. Issue: Flesh this out, along the lines of
-    https://docs.google.com/document/d/1JdJkdY7cK2rD9JXbK75sSn069c0rhF9UnLQACBZ1OtY/edit.
-
+1. [=list/For each=] header (|name|, |storedValue|) in |storedRequest|'s
+    [=request/header list=]:
+    1. If |name| doesn't name a [=content negotiation mechanism specified to be
```
twifkak: Maybe reword as e.g. [=header field specified to be content-negotiable with Variants=].

loading.bs
```diff
@@ -818,13 +831,23 @@ A [=request=] |browserRequest| <dfn>matches the stored request</dfn>
...
+        usable with Variants=], return "mismatch".
```
twifkak: Also, it seems a bit up in the air as to where the UA implementor should look for specifications of Variant content negotiation mechanisms. Appendix A is one place, but presumably new or experimental HTTP headers will define them as part of their own spec? Or maybe it's OK to leave this a bit up in the air for now?

loading.bs
```diff
@@ -818,13 +831,23 @@ A [=request=] |browserRequest| <dfn>matches the stored request</dfn>
...
+    1. Otherwise, let |variantMatches| be the algorithm for selecting a result
+        defined by |name|'s content negotiation mechanism.
+    1. If |browserRequest|'s [=request/header list=] [=header list/does not
+        contain=] |name|, return "mismatch".
```
twifkak: I think this is wrong, but I may be misreading this. I would think that if the cached response is `Accept-Language: foo` and the browser doesn't specify an `Accept-Language`, this would match, per the [specified content-negotiation mechanism](https://httpwg.org/http-extensions/draft-ietf-httpbis-variants.html#content-language).

loading.bs
```diff
@@ -818,13 +831,23 @@ A [=request=] |browserRequest| <dfn>matches the stored request</dfn>
...
+    1. Let |browserValue| be the [=header/combined value=] of |name| in
+        |browserRequest|'s [=request/header list=].
+    1. If |browserValue| is empty, return "mismatch".
```
twifkak: Ditto.

loading.bs
jyasskin: It was out of date, now that Fetch has removed "combined value".

Variants doesn't seem to handle the case where a request header is just missing: https://github.com/httpwg/http-extensions/issues/731.

loading.bs
twifkak: Yeah, works for me.

loading.bs
twifkak: Ah, okay, sounds good. I guess the effect of this is that SXG generators should choose the "greatest common subset" of variable headers that are supported by the UAs they're targeting. As long as UAs tend to add support for headers over time rather than remove, this should be fine.

loading.bs
twifkak: I think support for null values of negotiable request headers is a necessity, though. Otherwise, if 3 different UAs support 3 different sets of content negotiation algorithms, then it would be impossible to generate an SXG that could be used by all of them. The union of all supported headers would fail because of the first item above, and the intersection of all supported headers would fail because of this item.

Even given a single UA, it would make it difficult, as one would have to tailor the header list to the particular version of the UA.

(I hope I'm not misunderstanding; I haven't spent a lot of time thinking about this, so I may be.)

jyasskin: I agree. Now I just delegate to the content-negotiation algorithms, which I'm fixing in https://github.com/httpwg/http-extensions/pull/735.

PR #287, opened 2018-08-14
Operate over header bytes instead of logical exchanges.

### LINKED ISSUE #276: Make it easy to reuse the encoded header cbor when verifying signature

jyasskin:
Fixes #276 and makes it clearer that implementations shouldn't reserialize things per https://crbug.com/863499.

[Preview](https://jyasskin.github.io/webpackage/fix-276-clearly-reuse-header-bytes/draft-yasskin-http-origin-signed-responses.html), [Diff](https://tools.ietf.org/rfcdiff?url1=https://wicg.github.io/webpackage/draft-yasskin-http-origin-signed-responses.txt&url2=https://jyasskin.github.io/webpackage/fix-276-clearly-reuse-header-bytes/draft-yasskin-http-origin-signed-responses.txt)

The change to the signed-message format can only take effect in b2, of course.

### DIFF
--- a/draft-yasskin-http-origin-signed-responses.md
+++ b/draft-yasskin-http-origin-signed-responses.md
@@ -302,26 +302,26 @@ the headers as this serialization. Instead, this section defines a CBOR
 representation that can be embedded into other CBOR, canonically serialized
 ({{canonical-cbor}}), and then signed.
 
-The CBOR representation of an exchange `exchange`'s headers is the CBOR
-({{!RFC7049}}) array with the following content:
+The CBOR representation of a set of request and response metadata and headers is
+the CBOR ({{!RFC7049}}) array with the following content:
 
 1. The map mapping:
-   * The byte string ':method' to the byte string containing `exchange`'s
-     request's method.
-   * The byte string ':url' to the byte string containing `exchange`'s request's
+   * The byte string ':method' to the byte string containing the request's
+     method.
+   * The byte string ':url' to the byte string containing the request's
      effective request URI, which MUST be an absolute URL ({{terminology}}) with
      a scheme of "https".
-   * For each request header field in `exchange` except for the `Host` header
-     field, the header field's lowercase name as a byte string to the header
-     field's value as a byte string.
+   * For each request header field except for the `Host` header field, the
+     header field's lowercase name as a byte string to the header field's value
+     as a byte string.
 
      Note: `Host` is excluded because it is already part of the effective
      request URI.
 1. The map mapping:
-   * the byte string ':status' to the byte string containing `exchange`'s
-     response's 3-digit status code, and
-   * for each response header field in `exchange`, the header field's lowercase
-     name as a byte string to the header field's value as a byte string.
+   * The byte string ':status' to the byte string containing the response's
+     3-digit status code, and
+   * For each response header field, the header field's lowercase name as a byte
+     string to the header field's value as a byte string.
 
 ### Example ### {#example-cbor-representation}
 
@@ -439,7 +439,15 @@ signatures. Otherwise, each member of this list represents a signature with
 parameters.
 
 The client MUST use the following algorithm to determine whether each signature
-with parameters is invalid or potentially-valid for an `exchange`.
+with parameters is invalid or potentially-valid for an exchange's
+
+* `headers`, a byte sequence holding the canonical serialization
+  ({{canonical-cbor}}) of the CBOR representation ({{cbor-representation}}) of
+  the exchange's request and response metadata and headers, and
+* `payload`, a stream of bytes constituting the exchange's payload body (Section
+  3.3 of {{!RFC7230}}). Note that the payload body is the message body with any
+  transfer encodings removed.
+
 Potentially-valid results include:
 
 * The signed headers of the exchange so that higher-level protocols can avoid
@@ -453,9 +461,6 @@ actually invalid due to an expired OCSP response MAY retry with `forceFetch` set
 to retrieve an updated OCSP from the original server.
 {:#force-fetch}
 
-1. Let `payload` be the payload body (Section 3.3 of {{!RFC7230}}) of
-   `exchange`. Note that the payload body is the message body with any transfer
-   encodings removed.
 1. Let:
    * `signature` be the signature (binary content in the parameterised
      identifier's "sig" parameter).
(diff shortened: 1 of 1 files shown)


draft-yasskin-http-origin-signed-responses.md
```diff
@@ -941,35 +956,37 @@ Because different clients send different request header fields, and intermediate
...
 validates the `Signature` header field for an exchange represented as a normal
-HTTP request/response pair, it MUST pass only the subset of header fields
-defined by {{significant-headers}} to the validation procedure
-({{signature-validity}}).
+HTTP request/response pair, it MUST pass the serialized headers defined by
+{{serialized-headers}} to the validation procedure ({{signature-validity}})
+along with the `Signature` header field and the response's payload.
 
 If the client relies on signature validity for any aspect of its behavior, it
 MUST ignore any header fields that it didn't pass to the validation procedure.
```
nyaxt: Since we removed `{{significant-headers}}` concept, we might want to be pedantic here:

If the client relies on signature validity for any aspect of its behavior, it MUST ignore any header fields whose name are not listed in the `Signed-Headers` header field.

kinu:
Took a look, this looks good and am fine with either cherry-picking the format change only (nyaxt@ have a concern as the change included more) or landing this as is.

nyaxt:
IIUC, the commit has 2 goals:
- Change the signature message format to not use CBOR
- Explicitly pass CBOR header bytes to `{{ signature-validity }}` algorithm

I'm not 100% confident from 30min review if all `exchange` concept change and `{{ signature-validity }}` procedure arguments change are consistent across all spec text.
I think I'm fairly confident about the cross-origin path but I feel like there is subtlety with same-origin path.

Since we are getting close to branch point, we discussed this with @kinu and decided to proceed with the PR and then follow up with fix-ups if needed. I was wondering if we could cherry-pick just the message format change for now but the lack of active implementation on the same-origin path atm may justify it a bit. That being said, I'd appreciate any help on reviewing cross-origin path.

draft-yasskin-http-origin-signed-responses.md
```diff
@@ -518,17 +505,14 @@ to retrieve an updated OCSP from the original server.
...
-      1. The text string "headers" to the CBOR representation
-         ({{cbor-representation}}) of `exchange`'s headers.
+   1. If `cert-sha256` is set, a byte holding the value 32 followed by the 32
+      bytes of the value of `cert-sha256`. Otherwise a 0 byte.
+   1. The 8-byte big-endian encoding of the length in bytes of `validity-url`,
+      followed by the bytes of `validity-url`.
+   1. The 8-byte big-endian encoding of `date`.
+   1. The 8-byte big-endian encoding of `expires`.
+   1. The 8-byte big-endian encoding of the length in bytes of `headers`,
+      followed by the bytes of `headers`.
```
davidben: I think it's fine for this part to be CBOR if you all want it to be CBOR.

This format's fine too, though if you do that, I suggest using BoringSSL's `CBB` functions in the implementation (we don't currently have `CBB_add_u64{,_length_prefixed}` since TLS doesn't use those, but they'd be easy to add). TLS also has a struct syntax you could borrow. (An 8 byte length prefix is also kind of impressively large for both URL and headers. :-) I don't think any implementation would tolerate either field to exceed, say, a 3 byte length. But 8 bytes is cool too if you prefer.)

To the CBOR point, I think I didn't express the concern clearly. My bad. It's not that you shouldn't trust the CBOR encoder itself. Rather, let's look at the pipeline of conversions we've got here.

We convert the byte representation of the signed exchange, to a CBOR AST (CBORValue), to some parsed signed exchange struct (SignedExchangeEnvelope), to the net/loading stack's representation (HttpResponseHeaders). Along the way, to verify the signature, we convert the SignedExchangeEnvelope *back* to the CBOR AST, then *back* to bytes, and *then* verify it.

That is:

```
bytes -> CBORValue -> SignedExchangeEnvelope -> HttpResponseHeaders -...-> web stack
                                 |
                                 v
                             CBORValue
                                 |
                                 v
                               bytes
                                 |
                                 v
                         (verify signature)
```

We want the "verify signature" step to cover what was ultimately sent to the rest of the stack. That's the point of this whole process. We wish to know that everything downstream of the verification only sees content attested by the signature.

Suppose there's a logic bug in one of the right arrows. Maybe on certain malformed or weird input, we output something that can be massaged into an XSS. That's bad, but the signature mitigates a lot of potential attack vectors. The origin isn't going to sign the bad input, so as long as the verify step covered everything, an attacker can't use that bug to inject bad input into the origin. *The requirement on right arrows is the depend only on their inputs.*

However, suppose there is a logic bug in one of the down arrows. Maybe we forgot to copy some new field from the SignedExchangeEnvelope into the CBORValue. That's a problem. An attacker can put the new field in, the signature would check out, and all the right arrows would think this new field was covered, when it really wasn't. *The requirement on down arrows is they preserve all information.*

Note that, of these two requirements, one is much more straightforward than the other. It's hard to accidentally check a random global variable somewhere. :-) It's much easier to accidentally forget to encode something.

Thus the goal is to minimize the complexity in the verify step. The trick I like most is to find the simplest representation and verify that. An array of bytes is a good candidate. No one is going to extend that. All the complex structure is carried within, and the verify step is blind to it all.

Now, you can't always manage this. In the case of this logic, you are not just bytes. It's bytes plus some metadata decoded out of headers (cert-sha256, validity-url, etc). And, yeah, that means you need to go encode that metadata in some way. Maybe CBOR, maybe DER, or maybe something ad hoc. And, yeah, that's more down arrows. It would be more robust if you didn't have to do that, but there are also benefits to having the signature metadata outside of the payload, so the tradeoff is probably to leave those as is. Hopefully we don't have to extend those very often and the right arrows probably discard that information pretty quickly.

The rest of the payload, on the other hand, seems preferable to keep as bytes. Maybe someone will try to put trailers in there someday, I dunno.

draft-yasskin-http-origin-signed-responses.md
```diff
@@ -941,35 +956,37 @@ Because different clients send different request header fields, and intermediate
...
 * The method (Section 4 of {{!RFC7231}}) and effective request URI (Section 5.5
   of {{!RFC7230}}) of the request.
 * The response status code (Section 6 of {{!RFC7231}}) and the response header
   fields whose names are listed in that exchange's `Signed-Headers` header field
-  ({{signed-headers}}), in the order they appear in that header field. If a
-  response header field name from `Signed-Headers` does not appear in the
-  exchange's response header fields, the exchange has no significant headers.
+  ({{signed-headers}}). If a response header field name from `Signed-Headers`
+  does not appear in the exchange's response header fields, the exchange has no
+  serialized headers.
```
twifkak: I get a bit confused by the language "no serialized headers". It sounds like this means :method, :url, and :status would be omitted from {{cbor-representation}}, but I don't think that's what's intended.

draft-yasskin-http-origin-signed-responses.md
```diff
@@ -941,35 +956,37 @@ Because different clients send different request header fields, and intermediate
...
 
 If the exchange's `Signed-Headers` header field is not present, doesn't parse as
 a Structured Header ({{!I-D.ietf-httpbis-header-structure}}) or doesn't follow
 the constraints on its value described in {{signed-headers}}, the exchange has
-no significant headers.
+no serialized headers.
```
twifkak: (Ditto.)

draft-yasskin-http-origin-signed-responses.md
nyaxt: I'll do a separate PR for this.

PR #288, opened 2018-08-14
Put a fallback URL at the beginning of signed exchanges

### LINKED ISSUE #242: Fallback URLs for forward compatibility
Chromium would like to be able to ship these formats and then continue evolving them as they make their way through the standards process. QUIC attempts something like this by defining ["invariants"](https://tools.ietf.org/html/draft-ietf-quic-invariants-01), "properties of the QUIC transport protocol that are expected to remain unchanged over time as new versions of the protocol are developed." To try the same thing in these specifications, it might make sense to put a fallback URL in a fixed place in both the signed exchange and bundles formats.

For signed exchanges, this would be the request URL that's currently encoded as the `:url` field in the [request headers](https://tools.ietf.org/html/draft-yasskin-http-origin-signed-responses-04#section-3.2). For bundles, it would be the [`start_url`](https://www.w3.org/TR/appmanifest/#start_url-member) that's expected to be in the bundle's [manifest](https://tools.ietf.org/html/draft-yasskin-wpack-bundled-exchanges-00#section-3.2.2). In both cases, the value would be moved instead of copied to the new location, to avoid the possibility of inconsistencies.

When the client receives an unknown version of the format (either too old or too new), it would redirect to the fallback URL. Offline users would be out of luck, but online ones would still be able to get the content.

This could also be accomplished using content negotiation on the `Accept` header, but that's a [joint that might rust shut](https://www.imperialviolet.org/2016/05/16/agility.html), so we should try to make things work even if it's rusty, if possible.

Thoughts?

jyasskin:
This lets the browser redirect there if it doesn't recognize the version of the signed exchange.

This sits on top of #287 and #281 and fixes #242.

One question here is whether to include the method with the fallback URL. We can't do anything with a non-GET method right now, and after many years of HTTP, the only acceptable methods are GET and HEAD anyway, so I think we should consider dropping the method entirely, although not in this PR.

signed-responses: [Preview](https://jyasskin.github.io/webpackage/fallback-url/draft-yasskin-http-origin-signed-responses.html), [Diff](https://tools.ietf.org/rfcdiff?url1=https://jyasskin.github.io/webpackage/fix-276-clearly-reuse-header-bytes/draft-yasskin-http-origin-signed-responses.txt&url2=https://jyasskin.github.io/webpackage/fallback-url/draft-yasskin-http-origin-signed-responses.txt)

Loading: [Preview](https://jyasskin.github.io/webpackage/fallback-url/loading.html#mp-http-fetch)

### DIFF
--- a/draft-yasskin-http-origin-signed-responses.md
+++ b/draft-yasskin-http-origin-signed-responses.md
@@ -308,15 +308,12 @@ the CBOR ({{!RFC7049}}) array with the following content:
 1. The map mapping:
    * The byte string ':method' to the byte string containing the request's
      method.
-   * The byte string ':url' to the byte string containing the request's
-     effective request URI, which MUST be an absolute URL ({{terminology}}) with
-     a scheme of "https".
    * For each request header field except for the `Host` header field, the
      header field's lowercase name as a byte string to the header field's value
      as a byte string.
 
-     Note: `Host` is excluded because it is already part of the effective
-     request URI.
+     Note: `Host` is excluded because it is part of the effective request URI,
+     which is represented outside of this map.
 1. The map mapping:
    * The byte string ':status' to the byte string containing the response's
      3-digit status code, and
@@ -441,6 +438,8 @@ parameters.
 The client MUST use the following algorithm to determine whether each signature
 with parameters is invalid or potentially-valid for an exchange's
 
+* `requestUrl`, a byte sequence that can be parsed into the exchange's effective
+  request URI (Section 5.5 of {{!RFC7230}}),
 * `headers`, a byte sequence holding the canonical serialization
   ({{canonical-cbor}}) of the CBOR representation ({{cbor-representation}}) of
   the exchange's request and response metadata and headers, and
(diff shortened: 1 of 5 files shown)


draft-yasskin-http-origin-signed-responses.md
```diff
@@ -302,26 +302,23 @@ the headers as this serialization. Instead, this section defines a CBOR
...
 
-The CBOR representation of an exchange `exchange`'s headers is the CBOR
-({{!RFC7049}}) array with the following content:
+The CBOR representation of a set of request and response metadata and headers is
+the CBOR ({{!RFC7049}}) array with the following content:
 
 1. The map mapping:
-   * The byte string ':method' to the byte string containing `exchange`'s
-     request's method.
-   * The byte string ':url' to the byte string containing `exchange`'s request's
```
hajimehoshi: I was wondering if we'd need to fix the blow example, that includes ':url' in cbor representation.

nyaxt: Good catch. I'll file a separate PR for this

nyaxt: Filed #290

PR #291, opened 2018-08-21
Add signedexchange/version

hajimehoshi:
This PR introduces `Version` type to represent spec versions. This is needed to be a new package since this is referenced from multiple packages not only `signedexchange` but also `mice`.


nyaxt:
Lgtm

PR #301, opened 2018-09-04
dump-signedexchange: support b2 format

irori:
Also updates the spec comments in ReadExchange().

### DIFF
--- a/go/signedexchange/signedexchange.go
+++ b/go/signedexchange/signedexchange.go
@@ -187,9 +187,8 @@ func (e *Exchange) decodeRequest(dec *cbor.Decoder, ver version.Version) error {
 					return err
 				}
 				continue
-			} else {
-				return fmt.Errorf("signedexchange: found a deprecated request key %q", keyURL)
 			}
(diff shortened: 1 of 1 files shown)


go/signedexchange/signedexchange.go
```diff
@@ -180,11 +181,15 @@ func (e *Exchange) decodeRequest(dec *cbor.Decoder) error {
...
-			if err != nil {
-				return err
+			if ver == version.Version1b1 {
+				e.RequestURI, err = url.Parse(string(value))
+				if err != nil {
+					return err
+				}
+				continue
+			} else {
+				return fmt.Errorf("signedexchange: found a deprecated request key %q", keyURL)
```
hajimehoshi: `else` is not needed here?

```go
if ver == version.Version1b1 {
  // ...
  continue
}
return fmt.Errorf("signedexchange: found a deprecated request key %q", keyURL)
```

irori: done.

PR #318, opened 2018-10-24
gen-signedexchange: Write SXG to stdout when "-o -" is given

irori:
This is useful when using gen-signedexchange in a command pipeline.

### DIFF
--- a/go/signedexchange/cmd/gen-signedexchange/main.go
+++ b/go/signedexchange/cmd/gen-signedexchange/main.go
@@ -44,7 +44,7 @@ var (
 
 	flagDumpSignatureMessage = flag.String("dumpSignatureMessage", "", "Dump signature message bytes to a file for debugging.")
 	flagDumpHeadersCbor      = flag.String("dumpHeadersCbor", "", "Dump metadata and headers encoded as a canonical CBOR to a file for debugging.")
-	flagOutput               = flag.String("o", "out.sxg", "Signed exchange output file")
+	flagOutput               = flag.String("o", "out.sxg", "Signed exchange output file. If value is '-', sxg is written to stdout.")
 
 	flagRequestHeader  = headerArgs{}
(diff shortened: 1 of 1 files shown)


go/signedexchange/cmd/gen-signedexchange/main.go
```diff
@@ -44,7 +44,7 @@ var (
 
 	flagDumpSignatureMessage = flag.String("dumpSignatureMessage", "", "Dump signature message bytes to a file for debugging.")
 	flagDumpHeadersCbor      = flag.String("dumpHeadersCbor", "", "Dump metadata and headers encoded as a canonical CBOR to a file for debugging.")
-	flagOutput               = flag.String("o", "out.sxg", "Signed exchange output file")
+	flagOutput               = flag.String("o", "out.sxg", "Signed exchange output file. If value is '-', sxg is written to stdout.")
```
hajimehoshi: Hmm, how about changing the default value an empty string, which means stdout?

nyaxt: +1 irori's concern. Can we isatty?
One possible impl of this is https://github.com/mattn/go-isatty but feel free to use alternatives

hajimehoshi: +1 to use isatty. You are suggesting that if -o is omitted and isatty is true, this application should show warnings?

irori: Hmmm, isatty feels too fancy to me.

Can we go with the original approach ("-o -" means stdout)? That's a common convention specified in the [POSIX Utility Syntax Guidelines](http://pubs.opengroup.org/onlinepubs/009695399/basedefs/xbd_chap12.html#tag_12_02) (Guideline 13) so it wouldn't surprise users.

Issue #319, opened 2018-10-30
dump-signedexchange: Implement signature verification


irori:
Add a reference implementation of [signature validity](https://wicg.github.io/webpackage/draft-yasskin-http-origin-signed-responses.html#signature-validity) check in go/signedexchange, and use it from `dump-signedexchange`.

It would help implementors of other signed exchange generators. Also, it will enable us to add roundtrip (sign-and-verify) tests.

nyaxt:
SGTM

PR #323, opened 2018-11-01
Add Structured Headers parser

irori:
This is part of #319 and will be used to parse `signature` header.

### DIFF
--- a/go/signedexchange/structuredheader/parser.go
+++ b/go/signedexchange/structuredheader/parser.go
@@ -129,7 +129,7 @@ func (p *parser) parseItem() (interface{}, error) {
 	// OWS for items. https://github.com/httpwg/http-extensions/issues/703
 
 	if p.isEmpty() {
-		return nil, errors.New("Item expected, got EOS")
+		return nil, errors.New("item expected, got EOS")
 	}
 	c := p.input[0]
 	if c == '-' || isDigit(c) {
@@ -145,16 +145,16 @@ func (p *parser) parseItem() (interface{}, error) {
 	if isLCAlpha(c) {
 		return p.parseIdentifier()
 	}
-	return nil, fmt.Errorf("Item expected, got '%c'", c)
+	return nil, fmt.Errorf("item expected, got '%c'", c)
 }
 
 // http://httpwg.org/http-extensions/draft-ietf-httpbis-header-structure.html#parse-number
 func (p *parser) parseNumber() (int64, error) {
 	if p.isEmpty() {
-		return 0, errors.New("Number expected, got EOS")
+		return 0, errors.New("number expected, got EOS")
 	}
 	if p.input[0] != '-' && !isDigit(p.input[0]) {
-		return 0, fmt.Errorf("Number expected, got '%c'", p.input[0])
+		return 0, fmt.Errorf("number expected, got '%c'", p.input[0])
 	}
 	// TODO: Support Floats.
 	i := 1
@@ -172,7 +172,7 @@ func (p *parser) parseNumber() (int64, error) {
 // http://httpwg.org/http-extensions/draft-ietf-httpbis-header-structure.html#parse-string
 func (p *parser) parseString() (string, error) {
 	if p.isEmpty() {
-		return "", errors.New("String expected, got EOS")
+		return "", errors.New("string expected, got EOS")
 	}
 	if !p.consumeChar('"') {
 		return "", fmt.Errorf("'\"' expected, got '%c'", p.input[0])
@@ -204,10 +204,10 @@ func (p *parser) parseString() (string, error) {
 // http://httpwg.org/http-extensions/draft-ietf-httpbis-header-structure.html#parse-identifier
 func (p *parser) parseIdentifier() (Identifier, error) {
 	if p.isEmpty() {
-		return "", errors.New("Identifier expected, got EOS")
+		return "", errors.New("identifier expected, got EOS")
 	}
 	if !isLCAlpha(p.input[0]) {
-		return "", fmt.Errorf("Identifier expected, got '%c'", p.input[0])
+		return "", fmt.Errorf("identifier expected, got '%c'", p.input[0])
 	}
 	i := 0
 	for i < len(p.input) && isIdent(p.input[i]) {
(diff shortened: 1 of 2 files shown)


go/signedexchange/structuredheader/parser.go
```diff
@@ -0,0 +1,259 @@
...
+	"fmt"
+	"strconv"
+	"strings"
+)
+
+// http://httpwg.org/http-extensions/draft-ietf-httpbis-header-structure.html#identifier
+type Identifier string
+
+// http://httpwg.org/http-extensions/draft-ietf-httpbis-header-structure.html#param
+type ParameterisedList []*ParameterisedIdentifier
```
nyaxt: Simply `[]ParameterizedIdentifier`?

go/signedexchange/structuredheader/parser.go
```diff
@@ -0,0 +1,259 @@
...
+//	[]byte     for Byte Sequences
+//	nil        for parameters with no value
+type Parameters map[Identifier]interface{}
+
+type parser struct {
+	input string
+}
+
+func (p *parser) discardLeadingOWS() {
+	p.input = strings.TrimLeft(p.input, " \t")
```
hajimehoshi: I thought OWS is defined as a space char or a tab char that might be followed by whitespace characters: https://tools.ietf.org/html/rfc7230#section-3.2.3 In this implementation, only tab characters on left side would be removed.

hajimehoshi: Sorry this was my misunderstanding

go/signedexchange/structuredheader/parser.go
```diff
@@ -0,0 +1,259 @@
...
+}
+
+func (p *parser) isEmpty() bool {
+	return len(p.input) == 0
+}
+
+func (p *parser) getChar() byte {
```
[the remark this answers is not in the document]
hajimehoshi: As we discussed offline, I'm ok to go with `byte`. Rather, should the parser cause errors when non-ASCII chars were found?

go/signedexchange/structuredheader/parser.go
```diff
@@ -0,0 +1,259 @@
...
+		if err != nil {
+			return nil, err
+		}
+		items = append(items, item)
+		p.discardLeadingOWS()
+		if p.isEmpty() {
+			return items, nil
+		}
+		if !p.consumeChar(',') {
+			return nil, fmt.Errorf("COMMA expacted, got '%c'", input[0])
```
hajimehoshi: nits: error message should start with small caps in convention (there are a lot of messages don't follow this rule in this project though...)

https://github.com/golang/go/wiki/Errors

How about starting with `structuredheader:`?

go/signedexchange/structuredheader/parser_test.go
```diff
@@ -0,0 +1,260 @@
+package structuredheader
```
hajimehoshi: Let's make it a different package `structured_header`. We'd need to export some functions.

go/signedexchange/structuredheader/parser.go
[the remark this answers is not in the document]
irori: ASCII-ness is not explicitly checked, but non-ASCII characters will cause error in `perse*` functions where the character is consumed. For example, parseString() fails if the character isn't in the range [` `, `~`].

go/signedexchange/structuredheader/parser_test.go
irori: I thought this is the preferred way?

https://golang.org/pkg/testing/#pkg-overview

> To write a new test suite, create a file whose name ends _test.go that contains the TestXxx functions as described here. Put the file in the same package as the one being tested. The file will be excluded from regular package builds but will be included when the “go test” command is run. For more detail, run “go help test” and “go help testflag”.

go/signedexchange/structuredheader/parser.go
irori: Done.

go/signedexchange/structuredheader/parser.go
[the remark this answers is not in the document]
irori: Done.

go/signedexchange/structuredheader/parser.go
```diff
@@ -0,0 +1,259 @@
...
+//
+//	int64      for Numbers
+//	string     for Strings
+//	Identifier for Identifiers
+//	[]byte     for Byte Sequences
+//	nil        for parameters with no value
+type Parameters map[Identifier]interface{}
+
+type parser struct {
+	input string
```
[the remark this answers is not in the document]
irori: I'd like to keep using string, as functions like `strings.TrimLeft` can be used directly.

go/signedexchange/structuredheader/parser_test.go
hajimehoshi: We discussed offline and concluded it would not be worth making this a different package by exposing a lot of functions. (There is a technique to expose functions only for testing, but it is a little messy). Let's keep as this is. Thanks.

go/signedexchange/structuredheader/parser.go
```diff
@@ -0,0 +1,259 @@
...
+	if len%4 != 0 {
+		// Allow unpadded encoding.
+		enc = base64.RawStdEncoding
+	}
+	data, err := enc.DecodeString(s)
+	if err != nil {
+		return nil, fmt.Errorf("structuredheader: couldn't decode base64 %q: %v", s, err)
+	}
+	if !p.consumeChar('*') {
+		panic("cannot happen")
```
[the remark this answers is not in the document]
irori: Line 228 guarantees that here's `*`.

mnot:
FWIW, we've been collecting common tests in:
  https://github.com/httpwg/structured-header-tests

They haven't been updated in a while, so they might be out of sync with the current spec, but I plan on spending some time on them soon.

Contributions most welcome, of course. Having a common test suite for all implementations should help intro considerably.

irori:
Thanks. Current implementation lacks several features (toplevel is only Parameterised List, no float and boolean supports). I'd like to add these when I have time, and then we'll be able to use the test suite.

mnot:
Understood. Even if you just cherry-pick the applicable tests in the meantime, that would be great, as it would help assure there isn't too much drift. If there's anything we can do to help (e.g., separating into files differently, labelling test types, etc.) just say so.

PR #325, opened 2018-11-01
Do not use signedexchange/version in mice package

irori:
So that the package can be used independently from Signed Exchange.

### DIFF
--- a/go/signedexchange/mice/mice.go
+++ b/go/signedexchange/mice/mice.go
@@ -1,3 +1,5 @@
+// Package mice implements Merkle Integrity Content Encoding
+// (https://martinthomson.github.io/http-mice/draft-thomson-http-mice.html).
 package mice
 
 import (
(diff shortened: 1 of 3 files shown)


go/signedexchange/mice/mice.go
```diff
@@ -1,42 +1,72 @@
...
-// Encode encodes the given content buf to MICE (Merkle Integrity Content Encoding)
-// format.
-//
-// Encode returns MI header field parameter string and error if one exists for version 1b1, and
-// returns Digest header value parameter string and error if one exists for version 1b2.
-//
-// Spec: https://tools.ietf.org/html/draft-thomson-http-mice-02 for version 1b1, and
-// https://tools.ietf.org/html/draft-thomson-http-mice-03 for version 1b2
-func Encode(w io.Writer, buf []byte, recordSize int, ver version.Version) (string, error) {
+// ContentEncoding() returns content encoding name of the Encoding.
```
hajimehoshi: nit: No `()`

go/signedexchange/mice/mice.go
```diff
@@ -1,42 +1,72 @@
...
+func (enc Encoding) ContentEncoding() string {
+	return string(enc)
+}
+
+// DigestHeaderName() returns the name of HTTP header that carries integrity proofs.
```
hajimehoshi: ditto

PR #326, opened 2018-11-02
Reject uncachable status code

nyaxt:
Chromium CL: https://chromium-review.googlesource.com/c/chromium/src/+/1313712

***
<a href="https://pr-preview.s3.amazonaws.com/nyaxt/webpackage/pull/326.html" title="Last updated on Nov 20, 2018, 5:59 AM GMT (95b849a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/webpackage/326/54c7a94...nyaxt:95b849a.html" title="Last updated on Nov 20, 2018, 5:59 AM GMT (95b849a)">Diff</a>

### DIFF
--- a/draft-yasskin-http-origin-signed-responses.md
+++ b/draft-yasskin-http-origin-signed-responses.md
@@ -836,12 +836,12 @@ signature returns "valid", return "valid". Otherwise, return "invalid".
    this returned "invalid" or didn't return a certificate chain, return
    "invalid".
 1. Let `exchange` be the exchange metadata and headers parsed out of `headers`.
-1. If `exchange`'s request method is not safe (Section 4.2.1 of {{!RFC7231}}) or
-   not cacheable (Section 4.2.3 of {{!RFC7231}}), return "invalid".
-1. If `exchange`'s response is not complete, (Section 3.1 of {{!RFC7231}}),
+1. If `exchange`'s request method is not safe (Section 4.2.1 of {{!RFC7231}}),
    return "invalid".
-1. If `exchange`'s response is not cachable by a public cache (Section 3 of
-   {{!RFC7231}}), return "invalid".
+1. If `exchange`'s response is not complete (Section 3.1 of {{!RFC7234}}),
+   return "invalid".
+1. If Section 3 of {{!RFC7234}} forbids a shared cache from storing `exchange`'s
+   response, return "invalid".
 1. If `exchange`'s headers contain a stateful header field, as defined in
    {{stateful-headers}}, return "invalid".
 1. Let `authority` be the host component of `requestUrl`.
@@ -1934,7 +1934,7 @@ draft-05
 * Update to mice-03 including the Digest header.
 * Refer to draft-yasskin-httpbis-origin-signed-exchanges-impl for draft version
   numbers.
-* Require `exchange`'s response to be complete and cachable by a public cache.
+* Require `exchange`'s response to be complete and cachable by a shared cache.
 
 draft-04
 
--- a/draft-yasskin-http-origin-signed-responses.md
+++ b/draft-yasskin-http-origin-signed-responses.md
@@ -838,6 +838,10 @@ signature returns "valid", return "valid". Otherwise, return "invalid".
 1. Let `exchange` be the exchange metadata and headers parsed out of `headers`.
 1. If `exchange`'s request method is not safe (Section 4.2.1 of {{!RFC7231}}) or
    not cacheable (Section 4.2.3 of {{!RFC7231}}), return "invalid".
+1. If `exchange`'s response is not complete, (Section 3.1 of {{!RFC7231}}),
+   return "invalid".
+1. If `exchange`'s response is not cachable by a public cache (Section 3 of
+   {{!RFC7231}}), return "invalid".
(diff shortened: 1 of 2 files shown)


nyaxt:
@kinu wdyt?

nyaxt:
I think there is a reasonable set of status codes which we should allow, but not sure where the boundary is.

- **1xx Informational**: IIUC, these are partial responses which doesn't make sense to be captured.
- **non-cachable responses**: Disallowing those seem to align with the request method restrictions. (strawman)

@kinu made a point that redirect is disallowed in the loading spec, so there is also discussion of where to enforce the restriction

kinu:
I'm less sure about the non-cacheable responses, for example for CaheStorage we don't check any of these but specifically throw an error if it's 206 (partial response). And if we take a deeper look getting 203 (non-authoritative) seems a little weird, handling of 206 (partial content) responses would need to be discussed in the loading spec (as well as handling of range requests), and all 3xx are forbidden in the loading spec.  Overall I feel it might make better sense to have the part only in the loading spec, at least until we have a clearer thought of what should be valid.

kinu:
Well or maybe we can just start with 200 only. Looking into other status code most of them don't really make sense. (I prefer forbidding it in the loading spec though) /cc @horo-t

mnot:
A status code that is uncacheable by default can still be cached if it has explicit freshness information. I think you need to reference the entire [response cacheability algorithm](https://httpwg.org/specs/rfc7234.html#response.cacheability).

nyaxt:
@kinu @horo-t @mnot Thank you for your input!
I'll update the PR shortly to only allow 200 here to be most restrictive, but I'm open to loosen the restriction to the response cacheablility algorithm if we have a use case.

nyaxt:
@kinu Would you mind sharing the background why you'd prefer this to be loading spec?

I'm mixed since we already have `:method` cacheability validation in the sxg format spec. While I think having it outside format spec makes sense if we consider non-webbrowser consumers in the future (which are not covered by `loading.md` iiuc).

Another strawman would be to have:
- format spec say only cachable responses are signature valid
- loading spec to allow only 200

draft-yasskin-http-origin-signed-responses.md
```diff
@@ -838,6 +838,10 @@ signature returns "valid", return "valid". Otherwise, return "invalid".
 1. Let `exchange` be the exchange metadata and headers parsed out of `headers`.
 1. If `exchange`'s request method is not safe (Section 4.2.1 of {{!RFC7231}}) or
    not cacheable (Section 4.2.3 of {{!RFC7231}}), return "invalid".
```
jyasskin: Let's replace the "or not cacheable" bit with the reference to https://httpwg.org/specs/rfc7234.html#rfc.section.3, since that also talks about methods.

draft-yasskin-http-origin-signed-responses.md
```diff
@@ -838,6 +838,10 @@ signature returns "valid", return "valid". Otherwise, return "invalid".
...
+1. If `exchange`'s response is not complete, (Section 3.1 of {{!RFC7231}}),
```
jyasskin:
```suggestion
1. If `exchange`'s response is not complete, (Section 3.1 of {{!RFC7234}}),
```

draft-yasskin-http-origin-signed-responses.md
```diff
@@ -838,6 +838,10 @@ signature returns "valid", return "valid". Otherwise, return "invalid".
...
+   return "invalid".
+1. If `exchange`'s response is not cachable by a public cache (Section 3 of
```
jyasskin: I believe "shared cache" is the right term, according to https://httpwg.org/specs/rfc7234.html#rfc.section.1. "public" is a response directive, not a type of cache.

I'd also rather not use the term "cacheable" here since https://httpwg.org/specs/rfc7234.html#rfc.section.3 doesn't define it as a predicate on responses. Instead perhaps:

```suggestion
1. If Section 3 of {{RFC7234}} forbids a shared cache from storing `exchange`'s response, return "invalid".
```

loading.bs
```diff
@@ -733,7 +733,7 @@ returns a failure or an [=exchange=] via the following steps:
     * |headers|[0] contains a `` `host` `` key.
     * |headers|[0][`` `:method` ``] is not `` `GET` ``.
     * |headers|[1] contains any keys starting with `` `:` `` that aren't `` `:status` ``.
-    * |headers|[1][`` `:status` ``] doesn't consist of exactly 3 digits.
+    * |headers|[1][`` `:status` ``] is not `` 200 ``.
```
jyasskin: I think I agree with having this tighter restriction in the loading spec only, at least for now.

draft-yasskin-http-origin-signed-responses.md
nyaxt: Done.

draft-yasskin-http-origin-signed-responses.md
nyaxt: Done.

draft-yasskin-http-origin-signed-responses.md
nyaxt: Done

PR #333, opened 2018-11-07
dump-signedexchange: verify exchange's signature

irori:
This adds -verify option to dump-signedexhcange, which runs
the verification algorithm [1] on the given signed-exchange.

By default certificate chain is fetched from signature's cert-url,
but that can be overridden by -cert flag.

Currently it only checks signature validity [2], and does not
perform certificate verification (step 7 of [2]).

[1] draft-yasskin-http-origin-signed-responses.html#cross-origin-trust
[2] draft-yasskin-http-origin-signed-responses.html#signature-validity

### DIFF
--- a/go/signedexchange/cmd/dump-signedexchange/main.go
+++ b/go/signedexchange/cmd/dump-signedexchange/main.go
@@ -3,8 +3,10 @@ package main
 import (
 	"flag"
 	"fmt"
+	"io/ioutil"
 	"log"
 	"os"
+	"time"
 
 	"github.com/WICG/webpackage/go/signedexchange"
 )
@@ -12,6 +14,8 @@ import (
 var (
 	flagInput     = flag.String("i", "", "Signed-exchange input file")
 	flagSignature = flag.Bool("signature", false, "Print only signature value")
+	flagVerify    = flag.Bool("verify", false, "Perform signature verification")
+	flagCert      = flag.String("cert", "", "Certificate CBOR file. If specified, used instead of fetching from signature's cert-url")
 )
 
 func run() error {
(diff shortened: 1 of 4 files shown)


irori:
This is the main patch of #319.

go/signedexchange/signedexchange_test.go
```diff
@@ -222,3 +225,60 @@ func TestSignedExchangeBannedCertUrlScheme(t *testing.T) {
...
+	const ver = version.Version1b2
+	e, err := NewExchange(ver, u, nil, 200, header, []byte(payload))
+	if err != nil {
+		t.Fatal(err)
+	}
+	if err := e.MiEncodePayload(16); err != nil {
+		t.Fatal(err)
+	}
+
+	now := time.Date(2018, 1, 31, 17, 13, 20, 0, time.UTC)
```
nyaxt: Would you add a comment what this is?

irori: Oh, it was just copied from an existing test case, but it doesn't have to be fixed timestamp.
Changed to `time.Now()`.

hajimehoshi: Would a constant time be better for testing in general?

irori: Yeah maybe. (I was Chromium build sheriff on Mon and Tue, and saw tests failing because of the end of DST...)

Changed to a (better-named) fixed timestamp.

go/signedexchange/cmd/dump-signedexchange/main.go
```diff
@@ -32,13 +36,45 @@ func run() error {
...
+	if *flagVerify {
+		verify(e)
+	}
+
+	e.PrettyPrintPayload(os.Stdout)
+
 	return nil
 }
 
+func verify(e *signedexchange.Exchange) {
```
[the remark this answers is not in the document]
irori: Done.

Since this is a dump tool, verification errors are not considered "hard error", so printing them to stdout and continue processing is fine I think.

go/signedexchange/verifier.go
```diff
@@ -0,0 +1,240 @@
...
+	}
+	decoded, err := ioutil.ReadAll(dec)
+	if err != nil {
+		return nil, err
+	}
+	return decoded, nil
+}
+
+func isSameOrigin(u1, u2 *url.URL) bool {
+	return u1.Scheme == u2.Scheme && u1.Host == u2.Host
```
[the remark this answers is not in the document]
irori: URL.Host contains port number.
https://golang.org/pkg/net/url/#URL

go/signedexchange/verifier.go
```diff
@@ -0,0 +1,240 @@
...
+	}
+	if sig.Expires, ok = params["expires"].(int64); !ok {
+		return nil, errors.New("verify: no valid 'expires' value")
+	}
+	return sig, nil
+}
+
+// CertFetcher takes certificat URL and returns certificate bytes in
+// application/cert-chain+cbor format.
+type CertFetcher = func(url string) ([]byte, error)
```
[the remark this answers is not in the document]
irori: I don't have strong opinion either. Can we go with a function pointer for now, since it requires fewer lines?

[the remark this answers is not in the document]
hajimehoshi: Sure, go with a function pointer :-)

go/signedexchange/cmd/dump-signedexchange/main.go
```diff
@@ -32,10 +36,43 @@ func run() error {
...
-	} else {
-		e.PrettyPrint(os.Stdout)
+		return nil
 	}
 
+	e.PrettyPrintHeaders(os.Stdout)
+
+	if *flagVerify {
+		if err := verify(e); err != nil {
+			return err
```
hajimehoshi: `fmt.Println(err)`? Now the error is handled by `log.Fatal`, which I think is different from what we want.

irori: `verify` returns error not on verification errors, but when a bad command-line flag is given (e.g. -cert file doesn't exist). In such cases, probably we should exit with non-zero status.

PR #339, opened 2018-11-14
Reject uncached headers

### LINKED ISSUE #336: Transfer-Encoding in inner-exchange's response header.
I've noticed the `Transfer-Encoding` header usage in SXG inner response headers.
We should specify how it will work.

- Option 1) Add to banned headers list.
- Option 2) Clarify in the loading spec that...
  - Option 2.a) Only compression schemes are allowed (Ban only `Transfer-Encoding: chunked`)
  - Option 2.b) They are ignored.

Note that Chromium ToT currently ~~ignores them. Which means, the response bytes are treated as they are raw bytes and doesn't go through decompressor/unchunking.~~ supports them

### LINKED ISSUE #340: Consider banning Strict-Transport-Security and Public-Key-Pins headers
In https://github.com/WICG/webpackage/issues/336#issuecomment-439560219, @sleevi suggests that we add the contents of [`kSecurityStateHeaders`](https://cs.chromium.org/chromium/src/net/http/http_response_headers.cc?l=77) to https://wicg.github.io/webpackage/draft-yasskin-http-origin-signed-responses.html#stateful-headers.

@sleevi, you're worried about attackers targeting sites that use a set of HSTS domains as an authentication cookie?

HTTP will define some set of headers that shared caches don't store, and I'd like to just blindly ban that list in signed exchanges, but I do see an argument to be a bit stricter in signed exchanges, since with signed exchanges the attacker doesn't have to be the first to visit the cache.

On the other hand, to be vulnerable to this attack, a site has to be using HSTS in a user-hostile way.

nyaxt:
Fixes #336 and fixes #340.

***
<a href="https://pr-preview.s3.amazonaws.com/nyaxt/webpackage/pull/339.html" title="Last updated on Jan 10, 2019, 5:26 PM UTC (3e85268)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/webpackage/339/80005f0...nyaxt:3e85268.html" title="Last updated on Jan 10, 2019, 5:26 PM UTC (3e85268)">Diff</a>

### DIFF
--- a/draft-yasskin-http-origin-signed-responses.md
+++ b/draft-yasskin-http-origin-signed-responses.md
@@ -1968,7 +1968,7 @@ draft-05
   metadata and headers.
 * Explicitly check the response payload's integrity instead of assuming the
   client did it elsewhere in processing the response.
-* Reject Transfer-Encoding response header.
+* Reject uncached header fields.
 
 draft-04
 
--- a/draft-yasskin-http-origin-signed-responses.md
+++ b/draft-yasskin-http-origin-signed-responses.md
@@ -834,6 +834,8 @@ signature returns "valid", return "valid". Otherwise, return "invalid".
    return "invalid".
 1. If `response`'s headers contain a stateful header field, as defined in
    {{stateful-headers}}, return "invalid".
+1. If `exchange`'s response headers contain `Transfer-Encoding` header field,
+   return "invalid".
 1. Let `authority` be the host component of `requestUrl`.
 1. Validate the `certificate-chain` using the following substeps. If any of them
    fail, re-run {{signature-validity}} once over the signature with the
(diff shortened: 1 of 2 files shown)


nyaxt:
We would like to rely on https://github.com/httpwg/http-core/pull/181 instead. However we might want to use this for the implementation snapshot depending on the time frame.

jyasskin:
I'm not hopeful that httpwg/http-core#181 will get in by the end of this week, so I'm probably going to add the rest of the hop-by-hop headers and other non-cached headers here and get this merged.

jyasskin:
Please take a look. This will fix #336 and fix #340.

twifkak:
Sorry for the delay - why is the list of hop-by-hop headers here (and in Chromium) different from the one listed at https://tools.ietf.org/html/rfc2616#section-13.5.1?

PR #341, opened 2018-11-20
verifier.go: Add checks for request method and stateful headers

irori:
This implements Step 3-5 of the "Cross-origin trust" algorithm
([Section 4 of draft-yasskin-http-origin-signed-responses.html](https://wicg.github.io/webpackage/draft-yasskin-http-origin-signed-responses.html#rfc.section.4)).

### DIFF
--- a/go/signedexchange/verifier.go
+++ b/go/signedexchange/verifier.go
@@ -98,14 +98,15 @@ func (e *Exchange) Verify(verificationTime time.Time, certFetcher CertFetcher, l
 	// "...and run the following algorithm for each signature, stopping at the
 	// first one that returns "valid". If any signature returns "valid", return
 	// "valid". Otherwise, return "invalid"."
+SignatureLoop:
 	for _, item := range signatures {
 		signature, err := extractSignatureFields(item)
 		if err != nil {
 			l.Printf("Invalid signature: %v", err)
 			continue
 		}
-		// Step 1: "If the signature's "validity-url" parameter is not same-origin
-		// with requestUrl, return "invalid"."
+		// Step 1: "If the signature's "validity-url" parameter is not
+		//         same-origin with requestUrl, return "invalid"."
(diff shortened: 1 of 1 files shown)


go/signedexchange/verifier.go
```diff
@@ -115,16 +116,47 @@ func (e *Exchange) Verify(verificationTime time.Time, certFetcher CertFetcher, l
...
+		}
+
+		// Step 5: "If exchange's headers contain a stateful header field, as
+		//         defined in Section 4.1, return "invalid"."
+		for k := range e.RequestHeaders {
+			if IsStatefulRequestHeader(k) {
+				l.Printf("Exchange has stateful request header %q.", k)
+				continue SignatureLoop
+			}
+		}
```
nyaxt: Would you factor out a helper method so we wont have label jump `continue`s?

go/signedexchange/verifier.go
```diff
@@ -115,16 +116,47 @@ func (e *Exchange) Verify(verificationTime time.Time, certFetcher CertFetcher, l
...
 		_, err = verifySignature(e, verificationTime, certFetcher, signature)
 		if err != nil {
 			l.Printf("Verification of sinature %q failed: %v", signature.Label, err)
 			continue
 		}
-		// TODO: Implement Step 3-7.
+
+		// Step 3: "Let exchange be the exchange metadata and headers parsed out
+		//         of headers."
+		// e contains the exchange metadata and headers.
```
nyaxt: L132 seems broken?

go/signedexchange/verifier.go
irori: Done.

go/signedexchange/verifier.go
irori: Not actually, 'e' refers to a local variable.
Added `` ` `` around `e` hoping to make it easier to understand.

PR #343, opened 2018-11-21
go/signedexchange: More tests for signature verification

irori:


go/signedexchange/signedexchange_test.go
```diff
@@ -227,13 +227,15 @@ func TestSignedExchangeBannedCertUrlScheme(t *testing.T) {
 	}
 }
 
-func TestVerify(t *testing.T) {
+func createTestExchange(t *testing.T) (e *Exchange, s *Signer, certBytes []byte) {
 	u, _ := url.Parse("https://example.com/")
-	header := http.Header{}
-	header.Add("Content-Type", "text/html; charset=utf-8")
+	reqHeaders := http.Header{}
+	reqHeaders.Add(":method", "GET")
```
nyaxt: The current situation around `Exchange.RequestHeaders` is quite tricky. We might want to resolve it first.
`func (*Exchange) encodeRequest()` assumes `":method"` key to not be in the `RequestHeaders` map, while `func (*Exchange) decodeRequest()` will populate the key.

irori: OK, created #344.

irori:
Rebased onto master after #344. PTAL

go/signedexchange/signedexchange_test.go
```diff
@@ -229,7 +229,7 @@ func TestSignedExchangeBannedCertUrlScheme(t *testing.T) {
 	}
 }
 
-func TestVerify(t *testing.T) {
+func createTestExchange(t *testing.T) (e *Exchange, s *Signer, certBytes []byte) {
```
nyaxt: @hajimehoshi Any thoughts on using named return values here?
https://tour.golang.org/basics/7 discourages using it on long func, but I'm open here.

PR #346, opened 2018-11-30
Strictly UTF-8 decode the fallback URL

jyasskin:
Bikeshed warned that I'd never defined |fallbackUrl|, but I missed it
before going on parental leave. Do y'all see any problems with picking this conversion from bytes to a string?

@irori, we ought to test this with a couple bad fallback URLs, say one starting with a BOM, and one with an invalid sequence that the UTF-8 decoder might convert to U+FFFD. Would you be willing to add those tests?

***
<a href="https://pr-preview.s3.amazonaws.com/jyasskin/webpackage/pull/346.html" title="Last updated on Dec 1, 2018, 12:31 AM GMT (121ecb1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/webpackage/346/2eddf31...jyasskin:121ecb1.html" title="Last updated on Dec 1, 2018, 12:31 AM GMT (121ecb1)">Diff</a>

### DIFF
--- a/loading.bs
+++ b/loading.bs
@@ -414,6 +414,9 @@ the result of the following steps:
 1. Let |fallbackUrlBytes| be the result of [=read buffer/reading=]
     |fallbackUrlLength| bytes from |stream|.
 1. If |fallbackUrlBytes| is a failure, return it.
+1. Let |fallbackUrl| be the result of [=UTF-8 decode without BOM or fail=] on
+    |fallbackUrlBytes|.
+1. If |fallbackUrl| is a failure, return it.
 1. If the result of running the [=URL parser=] on |fallbackUrl| is a failure, if
     it has a non-null [=url/fragment=], or if its [=url/scheme=] is something
     other than `"https"`, return a failure.


jyasskin:
I should add that I found https://cs.chromium.org/chromium/src/net/http/http_response_headers.cc?l=920&rcl=6aa20f013414d789e8d72752a6885ef3d990d274 after I wrote this change, indicating that maybe we'd rather `Location` headers be all-ASCII. We can enforce that for SXG URLs if y'all think it's a good idea.

This is also related to https://github.com/whatwg/fetch/issues/843.

annevk:
UTF-8 is generally preferable I think, but it depends a bit on what kind of field this is. Happy to help if you could give some context.

jyasskin:
@annevk This is the field at the beginning of the Signed Exchange format that gives both the URL of the inner resource and a target to redirect to if the Signed Exchange is broken (doesn't parse, the signature doesn't validate, maybe other things).

annevk:
I'd make those UTF-8. The URL parser will turn it all into ASCII, but it seems preferable to be able to serialize in more ways.

jyasskin:
Thanks, @annevk!

irori:
For the record, tests for this are https://github.com/web-platform-tests/wpt/commit/2e19cbe5d5d74821056d4ae2cef435accefaf4ef and https://github.com/web-platform-tests/wpt/commit/e663fa084dcdac43fea5cdd6c69b059b5ad0743f.

PR #348, opened 2018-12-08
Add a security consideration about content sniffing.

### LINKED ISSUE #321: Reflecting the URL leads to content sniffing bugs
When a server returns a web package in the application/signed-exchange format, it'll embed the request URL in the returned binary blob in the fallbackUrl field. Depending on the server implementation, the encoding of bytes of the URL might vary. Some implementation might choose the easiest solution, which is to just reflect whatever representation the client sends.

This is problematic, because this gives attackers an easy injection point within the first few bytes of the file, which can be turned into code running in the victim origin using content sniffing. The Adobe PDF parser for example ignores the unknown prefix bytes if it finds a valid PDF signature within the first 1kb of the file.

The "x-content-type-options: nosniff" header instruct browsers to not do content sniffing, but plugins can choose to (and in many cases do) ignore them. Because of this, the usual suggestion is to host partially user controlled binary files only on sandbox domains (when it's not possible to sufficiently escape the user controlled parts, which is usually not the case).

This demo page is not vulnerable to this, because it requires a special Accept request header: https://azei-package-test.com/ Clients that can't interpret the web package format will never send this header, so they'll never see the signed exchange file format.

One solution would be to use some kind of encoding for all reflected values (e.g. base64) to avoid introducing a valuable injection point in the first place.

jyasskin:
This fixes #321, I think. @molnarg, how do you feel about this?

I'll need to add this to bundles too, but I want to get the SXG text right first.

***
<a href="https://pr-preview.s3.amazonaws.com/jyasskin/webpackage/pull/348.html" title="Last updated on Dec 13, 2018, 12:52 AM GMT (052993b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/webpackage/348/dcb188d...jyasskin:052993b.html" title="Last updated on Dec 13, 2018, 12:52 AM GMT (052993b)">Diff</a>

### DIFF
--- a/draft-yasskin-http-origin-signed-responses.md
+++ b/draft-yasskin-http-origin-signed-responses.md
@@ -1326,9 +1326,8 @@ While modern browsers tend to trust the `Content-Type` header sent with a
 resource, especially when accompanied by `X-Content-Type-Options: nosniff`,
 plugins will sometimes search for executable content buried inside a resource
 and execute it in the context of the origin that served the resource, leading to
-XSS vulnerabilities. One bad plugin in this regard is the Adobe PDF reader,
-which looks for `%PDF` anywhere in the first 1kB and executes the code that
-follows it.
+XSS vulnerabilities. For example, some PDF reader plugins look for `%PDF`
+anywhere in the first 1kB and execute the code that follows it.
 
 The `application/signed-exchange` format ({{application-signed-exchange}})
 includes a URL and request and response headers early in the format, which an
--- a/draft-yasskin-http-origin-signed-responses.md
+++ b/draft-yasskin-http-origin-signed-responses.md
@@ -1320,6 +1320,49 @@ Even once all clients reject these certificates in TLS connections, this will
 still just discourage and not prevent key re-use, since a server operator can
 unwisely request two different certificates with the same private key.
 
+## Content sniffing ## {#seccons-content-sniffing}
+
+While modern browsers tend to trust the `Content-Type` header sent with a
+resource, especially when accompanied by `X-Content-Type-Options: nosniff`,
+plugins will sometimes search for executable content buried inside a resource
+and execute it in the context of the origin that served the resource, leading to
+XSS vulnerabilities. One bad plugin in this regard is the Adobe PDF reader,
+which looks for `%PDF` anywhere in the first 1kB and executes the code that
+follows it.
+
+The `application/signed-exchange` format ({{application-signed-exchange}})
+includes a URL and request and response headers early in the format, which an
+attacker could use to cause these plugins to sniff a bad content type.
+
+To avoid vulnerabilities, servers are advised to only serve an
+`application/signed-exchange` resource (SXG) from a domain if it would also be
+safe for that domain to serve the SXG's content directly, and to follow at least
+one of the following strategies:
+
+1. Only serve signed exchanges from dedicated domains that don't have access to
+   sensitive cookies or user storage.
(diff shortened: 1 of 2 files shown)


annevk:
I need some help understanding the model.

Is the idea that `application/signed-exchange` and `application/webbundle` responses cannot be sniffed as scripts or same-origin quirks mode style sheets? If so, we could add them to <https://fetch.spec.whatwg.org/#should-response-to-request-be-blocked-due-to-mime-type?>. I don't think it matters for other nosniff contexts per https://mimesniff.spec.whatwg.org/#determining-the-computed-mime-type-of-a-resource.

Or is the idea that the responses they encompass (i.e., those in the bundle/archive) cannot be sniffed? If so, we could "dynamically inject" this header when we create those responses ensuring "determine nosniff" always returns true for them downstream.

jyasskin:
@annevk One of the bigger risks is that they might be sniffed as PDFs, Flash, or other plugin-recognized types. The risky file types don't necessarily follow the Fetch spec, so the header isn't guaranteed to work, but we thought it'd improve our chances.

@mikewest may have other answers. He suggested requiring nosniff here. We also want to make internal responses automatically nosniff, but that'll come in a separate change.

annevk:
Ah plugins, that is indeed a poorly specified area. Note that currently we don't have any language that suggests nosniff would work for plugins.

jyasskin:
Yep. I don't personally have evidence that this improves things for any particular plugin; we're just hoping. I don't have strong feelings about whether this is the right thing to do.

annevk:
I think we need to know concretely what this will help with so it can be tested for (and specified if it isn't already). In particular as we'd have to do this for other new formats too, presumably (e.g., Wasm)? (I see the benefits for the contained non-HTTP responses, but as you said that's separate.)

mikewest:
In general, I'd like for us to be strict about new mechanisms we're adding to the platform that might result in executable code. I don't have any concrete examples of places where plugins could misinterpret an SXG response as something unfortunate, but there are a number of examples of plugins' propensity to zealously sniff incoming content into executable code. I'd like to make that less likely.

That said, when I was talking with @jyasskin about this a few days ago, I incorrectly assumed that `nosniff` applied to plugin content. In the meantime, @jyasskin pointed out that https://html.spec.whatwg.org/multipage/iframe-embed-object.html#object-type-detection makes strictly obeying `Content-Type` headers entirely optional for plugin resources, and that we apparently don't do so in Chrome. I think we should change that (though I don't have any data that would suggest that it's safe to do so; I'll add some metrics). If we don't, then I think I agree with @annevk that the Fetch bits of this change wouldn't have any teeth.

> Or is the idea that the responses they encompass (i.e., those in the bundle/archive) cannot be sniffed? If so, we could "dynamically inject" this header when we create those responses ensuring "determine nosniff" always returns true for them downstream.

As above, I'd like for us to strictly enforce MIME type checks for resources contained in SXG responses. Implicitly injecting a nosniff requirement seems like it would indeed be simpler for developers, so I'd be on board with that kind of mechanism. It seems like requiring developers to include the header has a higher chance of increasing usage even in browsers that don't support SXG (as presumably developers would inject the header themselves on both the SXG-encoded resource, and the plain ol' resource), but that might be asking too much.

jyasskin:
I'm going to revert the discussion of `X-Content-Type-Options:nosniff` in this PR, to land the general anti-sniffing advice, and then open a new PR that deals specifically with that header, both for the SXG itself and for its contents.

PR #356, opened 2018-12-20
go/signedexchange: Run signedexchange tests against all supported versions

irori:
This patch turns all existing tests in signedexchange_test.go
into grouped tests parameterized by format version.


irori:
@hajimehoshi WDYT?

PR #359, opened 2019-01-02
Use Fetch's method of extracting a MIME type instead of MIMEsniff.

jyasskin:
Thanks for the pointer, @annevk.

***
<a href="https://pr-preview.s3.amazonaws.com/jyasskin/webpackage/pull/359.html" title="Last updated on Jan 7, 2019, 9:12 PM UTC (51fc87c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/webpackage/359/80005f0...jyasskin:51fc87c.html" title="Last updated on Jan 7, 2019, 9:12 PM UTC (51fc87c)">Diff</a>

### DIFF
--- a/loading.bs
+++ b/loading.bs
@@ -420,7 +420,7 @@ result of the following steps:
 
 1. Let |mimeType| be the result of [=header list/extracting a MIME type=] from
     |response|'s [=response/header list=].
-1. If |mimeType| is undefined, return undefined.
+1. If |mimeType| is a failure, return undefined.
 1. If |mimeType|'s [=MIME type/essence=] is not `"application/signed-exchange"`,
     return undefined.
 1. Let |params| be |mimeType|'s [=MIME type/parameters=]
--- a/loading.bs
+++ b/loading.bs
@@ -642,11 +642,11 @@ returns the result of the following steps:
     : [=request/mode=]
     :: "`cors`"
 1. Let |certResponse| be the result of [=fetching=] |certRequest|.
-1. If |certResponse|'s [=response/status=] is not `200`, or [=header
-    list/extracting a MIME type=] from |certResponse|'s [=response/header list=]
-    does not return `"application/cert-chain+cbor"`, return a failure.
-
-    Note: This forbids the MIME type from including parameters.
+1. If |certResponse|'s [=response/status=] is not `200`, return a failure.
+1. Let |certMimeType| be the result of [=header list/extracting a MIME type=]
+    from |certResponse|'s [=response/header list=].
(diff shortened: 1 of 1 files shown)


loading.bs
```diff
@@ -409,7 +409,8 @@ An exchange signature is a [=struct=] with the following items:
 The <dfn>signed exchange version</dfn> of a [=response=] |response| is the
 result of the following steps:
 
-1. Let |mimeType| be the [=supplied MIME type=] of |response|.
+1. Let |mimeType| be the result of [=header list/extracting a MIME type=] from
+    |response|'s [=response/header list=].
 1. If |mimeType| is undefined, return undefined.
```
annevk: It should be either a MIME type or failure at this point.

loading.bs
```diff
@@ -611,9 +612,9 @@ returns the result of the following steps:
     : [=request/mode=]
     :: "`cors`"
 1. Let |certResponse| be the result of [=fetching=] |certRequest|.
-1. If |certResponse|'s [=response/status=] is not `200`, or the [=supplied MIME
-    type=] of |certResponse| is not `"application/cert-chain+cbor"`, return a
-    failure.
+1. If |certResponse|'s [=response/status=] is not `200`, or [=header
+    list/extracting a MIME type=] from |certResponse|'s [=response/header list=]
+    does not return `"application/cert-chain+cbor"`, return a failure.
```
annevk: I suspect you want to return the essence of the returned MIME type, meaning this will have to be slightly more involved. (It was a little ambiguous already. E.g., do parameters matter?)

loading.bs
jyasskin: Whoops, thanks!

loading.bs
jyasskin: I intended to forbid parameters here, since https://wicg.github.io/webpackage/draft-yasskin-http-origin-signed-responses.html#internet-media-type-applicationcert-chaincbor doesn't define any. I think the current text accomplishes that? If you think it's better to ignore parameters (by using the essence), I'm happy to do that instead.

annevk:
If you indeed want to include parameters you want to omit the double quotes, per discussion in https://github.com/whatwg/mimesniff/issues/52. Typically these kind of checks ignore parameters however, so you might want to add a note to that effect and carefully test for it. (I'm actually not aware of a place where we do a literal check.)

jyasskin:
Ok, if everyone else ignores parameters, we will here too. Done and thanks.

jyasskin:
@nyaxt @kinu FYI, this changes the `application/cert-chain+cbor` MIME type check to allow and ignore parameters.

PR #361, opened 2019-01-03
Restrict the browser's request's method to GET or HEAD.

jyasskin:
Since this is after a 303 redirect, I think the method can only be something else if a service worker messes with things, but we should prevent service workers from getting us into a weird state.

***
<a href="https://pr-preview.s3.amazonaws.com/jyasskin/webpackage/pull/361.html" title="Last updated on Jan 3, 2019, 9:28 PM UTC (e33e03f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/webpackage/361/8f7d57c...jyasskin:e33e03f.html" title="Last updated on Jan 3, 2019, 9:28 PM UTC (e33e03f)">Diff</a>


nyaxt:
cc: @kinu @horo-t

Issue #362, opened 2019-01-07
Implement "Implementation snapshot b3"


nyaxt:
This issue tracks progress of implementation of [the b3 snapshot](#350) in both gen-signedexchange and Chromium.

| Task | gen-se | dump-se (verifier) | Chromium |
|---|---|---|---|
| "Response is complete" | n/a only  | :black_square_button: | :ballot_box_with_check:  |
| #360 shared cache cachable | :black_square_button: | :black_square_button: | :black_square_button: |
| https://github.com/httpwg/http-core/pull/181 Update the list of stateful request header fields | :black_square_button:  | :black_square_button:  | :black_square_button:  |

+ variants support

irori:
Go verifier changes (except Variants): https://github.com/WICG/webpackage/pull/384

PR #363, opened 2019-01-07
verifier.go: Reject incomplete responses

irori:
This is part of #362.


irori:
Response completeness is no longer required. (#368)
Closing.

PR #367, opened 2019-01-07
[b3] go: Verify if the response is cacheable by a shared cache

irori:


irori:
@hajimehoshi PTL, this one is not blocked by spec discussions.

PR #371, opened 2019-01-09
go: Verify() shouldn't replace Payload

irori:
Before this patch, `e.Verify()` replaced `e.Payload` with mice-decoded
payload bytes. It's unintuitive that `Verify()` mutates the `Exchange`, so
this patch lets `Verify()` return the decoded payload in addition to
the boolean result.

### DIFF
--- a/go/signedexchange/verifier.go
+++ b/go/signedexchange/verifier.go
@@ -87,9 +87,9 @@ func DefaultCertFetcher(url string) ([]byte, error) {
 // Signature timestamps are checked against verificationTime.
 // Certificates for signatures are fetched using certFetcher.
 // Errors encountered during verification are logged to l.
-// If successful, it returns true and the decoded payload. otherwise it returns
-// false and nil.
-func (e *Exchange) Verify(verificationTime time.Time, certFetcher CertFetcher, l *log.Logger) (bool, []byte) {
+// On success, returns decoded payload bytes (can be an empty slice). Otherwise,
+// returns nil.
+func (e *Exchange) Verify(verificationTime time.Time, certFetcher CertFetcher, l *log.Logger) []byte {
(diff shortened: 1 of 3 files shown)


go/signedexchange/verifier.go
```diff
@@ -154,66 +154,65 @@ func (e *Exchange) Verify(verificationTime time.Time, certFetcher CertFetcher, l
...
 		// Step 8: "Return "valid"."
-		return true
+		return true, decodedPayload
 	}
-	return false
+	return false, nil
 }
 
 // verifySignature verifies single signature, as described in
 // https://wicg.github.io/webpackage/draft-yasskin-http-origin-signed-responses.html#signature-validity.
```
hajimehoshi: We'd want comments about the returning values.

irori: Done.

go/signedexchange/verifier.go
```diff
@@ -162,9 +164,14 @@ func (e *Exchange) Verify(verificationTime time.Time, certFetcher CertFetcher, l
 		// TODO: Implement Step 6 and 7 (certificate verification).
 
 		// Step 8: "Return "valid"."
-		return true
+
+		// Make sure to return non-nil value, even if the payload is empty.
+		if decodedPayload == nil {
+			decodedPayload = make([]byte, 0)
+		}
+		return decodedPayload
```
hajimehoshi: Ah, I'm really sorry but I missed the case when this function succeeds and the result is empty. Then `Verify` should return a boolean value indicating success. (Using a difference between a nil slice and 0-length slice would be confusing)

Then the signature of this function would be

```go
func (e *Exchange) Verify(verificationTime time.Time, certFetcher CertFetcher, l *log.Logger) ([]byte, bool)
```

irori: No problem, done :)

PR #377, opened 2019-01-18
Define a CAA parameter governing CanSignHttpExchanges cert issuance.

jyasskin:
This gives a stronger indication that the domain owner has opted into the new
security risks than just that a CA issued a certificate with the extension.

@AGWA, how's this look? https://tools.ietf.org/html/draft-ietf-acme-caa-06#section-3 has a MUST, which I've downgraded to a SHOULD here after reading draft-ietf-acme-caa's IANA considerations.

[Preview](https://jyasskin.github.io/webpackage/define-caa/draft-yasskin-http-origin-signed-responses.html#cross-origin-cert-req); [Diff](https://tools.ietf.org/rfcdiff?url1=https://wicg.github.io/webpackage/draft-yasskin-http-origin-signed-responses.txt&url2=https://jyasskin.github.io/webpackage/define-caa/draft-yasskin-http-origin-signed-responses.txt)

### DIFF
--- a/draft-yasskin-http-origin-signed-responses.md
+++ b/draft-yasskin-http-origin-signed-responses.md
@@ -979,6 +979,16 @@ extension. This OID might or might not be used as the final OID for the
 extension, so certificates including it might need to be reissued once the final
 RFC is published.
 
+### Extensions to the CAA Record: cansignhttpexchanges Parameter {#caa-cansignhttpexchanges}
+
+A CAA parameter "cansignhttpexchanges" is defined for the "issue" and
+"issuewild" properties defined by {{!RFC6844}}.  The value of this parameter, if
+specified, SHOULD be "yes".
+
+If the "cansignhttpexchanges" parameter is not present and equal to "yes", the
+CA indicated by the "issue" or "issuewild" property SHOULD NOT issue certificates
+with the CanSignHttpExchanges extension defined in {{cross-origin-cert-req}}.
+
 # Transferring a signed exchange {#transfer}
 
 A signed exchange can be transferred in several ways, of which three are
@@ -1650,6 +1660,13 @@ Restrictions on usage:  N/A
 Author:  See Authors' Addresses section.
 
 Change controller:  IESG
+
+## The cansignhttpexchanges CAA Parameter {#iana-caa-cansignhttpexchanges}
+
+There are no IANA considerations for this parameter. As per the CAA
+specification, the parameter namespace for the CAA "issue" and "issuewild"
+properties has CA-defined semantics. This document merely specifies a
+RECOMMENDED semantic for parameters of the name "cansignhttpexchanges".
 
 --- back
 
--- a/draft-yasskin-http-origin-signed-responses.md
+++ b/draft-yasskin-http-origin-signed-responses.md
@@ -2136,5 +2136,6 @@ draft-02
 
 # Acknowledgements
 
-Thanks to Devin Mullins, Ilari Liusvaara, Justin Schuh, Mark Nottingham, Mike
-Bishop, Ryan Sleevi, and Yoav Weiss for comments that improved this draft.
+Thanks to Andrew Ayer, Devin Mullins, Ilari Liusvaara, Justin Schuh, Mark
+Nottingham, Mike Bishop, Ryan Sleevi, and Yoav Weiss for comments that improved
+this draft.
--- a/draft-yasskin-http-origin-signed-responses.md
+++ b/draft-yasskin-http-origin-signed-responses.md
@@ -985,9 +985,10 @@ A CAA parameter "cansignhttpexchanges" is defined for the "issue" and
 "issuewild" properties defined by {{!RFC6844}}.  The value of this parameter, if
 specified, SHOULD be "yes".
 
-If the "cansignhttpexchanges" parameter is not present and equal to "yes", the
-CA indicated by the "issue" or "issuewild" property SHOULD NOT issue certificates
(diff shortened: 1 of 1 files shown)


draft-yasskin-http-origin-signed-responses.md
```diff
@@ -959,6 +959,16 @@ extension. This OID might or might not be used as the final OID for the
...
 RFC is published.
 
+### Extensions to the CAA Record: cansignhttpexchanges Parameter {#caa-cansignhttpexchanges}
+
+A CAA parameter "cansignhttpexchanges" is defined for the "issue" and
+"issuewild" properties defined by {{!RFC6844}}.  The value of this parameter, if
+specified, SHOULD be "yes".
+
+If the "cansignhttpexchanges" parameter is not present and equal to "yes", the
+CA indicated by the "issue" or "issuewild" property SHOULD NOT issue certificates
```
nyaxt: Am I correct that this only applies to CA who issue certificates via ACME protocol?

sleevi: No. All CAs are required to check for and respect CAA (if present) by the Baseline Requirements.

This change would require that site operators wishing to obtain a CanSignHttpExchanges certificate take affirmative steps to modify their DNS to indicate consent. Combined with the above requirement, this spec change has the effect of declaring how all CAs must obtain consent to issue these certificates, and declaring the means by which site operators can deny consent. It favors 'defenders' (site operators at risk from a CA using a less-than-ideal validation method or being tricked temporarily) by requiring consent by opt-in, rather than opt-out.

nyaxt:
still lgtm

AGWA:
As current worded, it's OK for a CA which isn't listed in any issue/issuwild property to issue certificates with the CanSignHttpExchanges extension.  Instead, I'd suggest:

> CA SHOULD NOT issue certificates with the CanSignHttpExchanges extension defined in {{cross-origin-cert-req}} unless an applicable issue or issuewild property exists for the CA, and the "cansignhttpexchanges" parameter is present on the property and is equal to "yes"

jyasskin:
@AGWA Thanks! Done.

draft-yasskin-http-origin-signed-responses.md
```diff
@@ -959,6 +964,20 @@ extension. This OID might or might not be used as the final OID for the
...
+"issuewild" properties defined by {{!RFC6844}}.  The value of this parameter, if
+specified, SHOULD be "yes".
+
+A CA SHOULD NOT issue certificates with the CanSignHttpExchanges extension
+defined in {{cross-origin-cert-req}} unless an applicable issue or issuewild
+property exists for the CA, and the "cansignhttpexchanges" parameter is present
+on the property and is equal to "yes".
+
+The CA/Browser Forum Baseline Requirements ({{BRs}}) are expected to establish
+requirements around the treatment of this parameter.
```
sleevi: 1) In general, we should avoid SHOULD/SHOULD NOT requirements unless we provide guidance on when there are applicable exceptions. Alternatively, MUST/MUST NOT provides a clearer, unambiguous processing model, modulo what error handling should be (which presumably is MUST NOT issue)

2) The reference to the BRs may end up being circular. The BRs state the rules for which a CA can include arbitrary extensions (7.1.2.4.), so if there are requirements expressed here, they should trigger the rules of 7.1.2.4 (a) (ii) and 7.1.2.4 (b), and thus require following this spec and whatever it describes.

jyasskin: I've moved this to the definition of the CanSignHttpExchanges *extension*, where there's no question that we can MUST CAs. Now I don't need to point at the BRs.

jyasskin:
I'm hoping one of the certificate experts (@AGWA, @sleevi, @clintwilson) will approve this before I merge it.

draft-yasskin-http-origin-signed-responses.md
```diff
@@ -943,6 +943,11 @@ able to make even one unauthorized signature.
 
 Conforming CAs MUST NOT mark this extension as critical.
 
+A conforming CA MUST NOT issue certificates with this extension unless an
+applicable "issue" or "issuewild" CAA property ({{!RFC6844}}) exists for the CA,
+and the "cansignhttpexchanges" parameter ({{caa-cansignhttpexchanges}}) is
+present on the property and is equal to "yes".
+
```
AGWA: I just realized that this wording could be interpreted as allowing a CA to include the extension if just one of the dNSNames in the certificate has `cansignhttpexchanges=yes`. :grimacing: Obviously, the intent is that this parameter must be present for every dNSName.  I'll propose some better wording in a little bit unless someone else beats me to it.

jyasskin: See, this is why I wait. :)

AGWA: How about:

> A conforming CA MUST NOT issue certificates with this extension unless, for each dNSName in the subjectAltName extension of the certificate to be issued:
> 1. An "issue" or "issuewild" CAA property ({{!RFC6844}}) exists that authorizes the CA to issue the certificate; and
> 2. The "cansignhttpexchanges" parameter ({{caa-cansignhttpexchanges}}) is present on the property and is equal to "yes"

jyasskin: Done. RFC6844 manages to avoid saying that a certificate's domains are listed in the dNSName, but if saying it here is a problem, we have plenty of time to revise the wording.

AGWA: I cribbed the "each dNSName in the subjectAltName extension" wording from the BR's section on CAA so it should be solid.

PR #380, opened 2019-01-23
Update the references to header-structure to -09.

jyasskin:
Preview at https://jyasskin.github.io/webpackage/update-structured-headers/draft-yasskin-http-origin-signed-responses.html.

### DIFF
--- a/draft-yasskin-http-origin-signed-responses.md
+++ b/draft-yasskin-http-origin-signed-responses.md
@@ -167,7 +167,7 @@ of the exchange's payload.
 
 The `Signature` header is a Structured Header as defined by
 {{!I-D.ietf-httpbis-header-structure}}. Its value MUST be a parameterised list
-(Section 3.3 of {{!I-D.ietf-httpbis-header-structure}}). Its ABNF is:
+(Section 3.4 of {{!I-D.ietf-httpbis-header-structure}}). Its ABNF is:
 
     Signature = sh-param-list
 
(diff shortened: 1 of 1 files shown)


draft-yasskin-http-origin-signed-responses.md
```diff
@@ -181,13 +181,13 @@ values:
```
twifkak: In the non-diff above: "This specification gives no meaning to the identifier itself". Per httpwg/http-extensions#720, you may want to update this to "the primary identifier itself".

draft-yasskin-http-origin-signed-responses.md
```diff
@@ -705,7 +705,7 @@ all.
 
 The `Accept-Signature` header field is a Structured Header as defined by
 {{!I-D.ietf-httpbis-header-structure}}. Its value MUST be a parameterised list
-(Section 3.3 of {{!I-D.ietf-httpbis-header-structure}}). Its ABNF is:
+(Section 3.4 of {{!I-D.ietf-httpbis-header-structure}}). Its ABNF is:
 
     Accept-Signature = sh-param-list
```
twifkak: Ditto the next two paragraphs. Not sure if you want to stick "primary" in front of all of them, because too verbose. Maybe the first instance per para.

draft-yasskin-http-origin-signed-responses.md
jyasskin: Hm, https://github.com/httpwg/http-extensions/commit/3c96a9b14e82f0356f09ac680a2943bdca9830fc then removes that term in favor of "token". I think I'm going to keep just "identifier" here, since I think it's clear enough that it refers to the parameterised thing, and not any of the parameters, but @mnot FYI in case you want dependent specs to do something else.

Issue #378, opened 2019-01-18
Limit CanSignHttpExchange certificate lifetimes


jyasskin:
In [the blink-dev Intent-to-Ship thread](https://groups.google.com/a/chromium.org/d/msg/blink-dev/gPH_BcOBEtc/WXRzifVlDQAJ), @agwa suggests 

> It may also be a good idea to limit the lifetime of certificates used in SXG, to reduce the amount of time that an illegitimate certificate can be used, in case an attacker exploits a CA vulnerability to obtain future-dated good OCSP responses.

* @agwa [suggests 7-14 days](https://groups.google.com/a/chromium.org/d/msg/blink-dev/gPH_BcOBEtc/eL5q1f-rDQAJ).
* @clintwilson says ["Something around 7-10 days would be a nice target IMO, but even restricting to 30-90 days would be a substantial improvement."](https://groups.google.com/a/chromium.org/d/msg/blink-dev/gPH_BcOBEtc/mfxmsbZ3DQAJ)
* If we go to 7 days, we can probably get rid of any mention of OCSP responses. What's the longest certificate lifetime where folks would be comfortable dropping our requirement for OCSP stapling?
* A very-short lifetime will be a problem if a CA has to temporarily disable issuance, for example after security weaknesses in certain domain validation methods are discovered. 
* Even a 90-day limit probably requires automated certificate rotation, a'la ACME. Shorter certainly does.
* Should we say anything about re-using an old public key in a new certificate?

These requirements may well wind up in the [BRs](https://cabforum.org/baseline-requirements-documents/) instead of this repository's specs, but this issue will track getting them where they need to go.

jyasskin:
We talked to @grittygrease from Cloudflare about shorter lifetimes. My (fallible) notes are:
* 90-day lifetimes seem ok; anything shorter would require some code changes.
* 7-day lifetimes would multiply the number of certificates they need to handle enough that current designs might not scale.
* 7-day lifetime would limit uptime. In practice they see longer downtime periods for certificate issuance than OCSP.
* Clock skew is an issue, but SXG's requirement for stapled OCSP responses hits those issues anyway.
* Reusing key material across successive certificates doesn't help with any of the important problems.
* 7-day certificates could make CT logs significantly bigger, which could hurt that ecosystem.
[...]

PR #383, opened 2019-01-23
Bound validity periods to 90 days.

jyasskin:
This is not necessarily the limit we'll want in the long run, but it
helps ensure that any future changes can go into effect reasonably
quickly.

I won't submit this until I have some indication that the folks who have already started experimenting with signed exchanges can handle 90-day certificates.

@clintwilson, I want to make sure this doesn't cause any problems for you, since I believe Digicert has been issuing certificates with validity periods longer than 90 days. I don't intend to declare that any of your existing certificates are out of bounds.

@AGWA, this isn't intended to stop discussion on #378, just to put _some_ limit in place while we figure out what the right limit is.

Preview: https://jyasskin.github.io/webpackage/90-day-lifetime/draft-yasskin-http-origin-signed-responses.html#cross-origin-cert-req
Diff: https://tools.ietf.org/rfcdiff?url1=https://wicg.github.io/webpackage/draft-yasskin-http-origin-signed-responses.txt&url2=https://jyasskin.github.io/webpackage/90-day-lifetime/draft-yasskin-http-origin-signed-responses.txt

### DIFF
--- a/draft-yasskin-http-origin-signed-responses.md
+++ b/draft-yasskin-http-origin-signed-responses.md
@@ -961,6 +961,9 @@ able to make even one unauthorized signature.
 Certificates with this extension MUST be revoked if an unauthorized entity is
 able to make even one unauthorized signature.
 
+Certificates with this extension MUST have a Validity Period no greater than 90
+days.
+
 Conforming CAs MUST NOT mark this extension as critical.
 
 A conforming CA MUST NOT issue certificates with this extension unless, for each
@@ -2061,6 +2064,7 @@ draft-06
 * Add a security consideration for future-dated OCSP responses and for stolen
   private keys.
 * Define a CAA parameter to opt into certificate issuance.
+* Limit certificate lifetimes to 90 days.
 
 draft-05
 
--- a/draft-yasskin-http-origin-signed-responses.md
+++ b/draft-yasskin-http-origin-signed-responses.md
@@ -990,6 +990,15 @@ extension. This OID might or might not be used as the final OID for the
 extension, so certificates including it might need to be reissued once the final
 RFC is published.
 
+Some certificates have already been issued with this extension and with validity
+periods longer than 90 days. These certificates will not immediately be treated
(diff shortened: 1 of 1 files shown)


kinu:
> I won't submit this until I have some indication that the folks who have already started experimenting with signed exchanges can handle 90-day certificates.

It'd be probably helpful to clarify what should / would happen to the sites that are already experimenting (or ordering) when this change takes effect, and on any other nearish-term changes.

clintwilson:
@shigeki Just a note that ordering for shorter validity will also automatically prorate/decrease the price of the certificate relative to the validity selected. If you need any help with your account or testing, please don't hesitate to reach out.
I think we could also get ACME to work for provisioning if that's beneficial. We'll have to map the inclusion of the extension out of band, so I'd need to come up with a solution there, but otherwise it should be feasible.

shigeki:
@clintwilson Thank for offering your help. We can purchase only one or two years SXG certs via Cyberturst in Japan. 

ACME needs challenge and response so that we can do it to fallbackURL if it is online. Currently fallbackURL is implemented as reqeustURL. That has a risk to renew the existing cert in online .

I think it is too early to limit the lifetime of SXG cert in 90 days for we do not know if it can really operative. 

6.2. Off-path attackers in Security considerations only describes about an attacker who already possesses a valid certificate. We should add the consideration of how much we need to care about misissuance  of CA before discussing SXG cert lifetime. Do we really need to care about a serious CA incident such as DigiNotar in 2011?

jyasskin:
@shigeki Thanks for the feedback and for continuing the conversation via @sisidovski. It looks like you've found a way to get a [<1-year certificate](https://crt.sh/?id=1177695640) that matches the timeline I added here since your comment.

I wrote up some considerations at https://github.com/WICG/webpackage/issues/378#issuecomment-459831551 that I think mean we should go for a 90-day max-lifetime for this year, and the other 2 people who voted, agreed. Given that, I'm going to merge this change.

PR #396, opened 2019-02-08
Don't require CAs to cap lifetime or check CAA until May 1.

jyasskin:
This makes it clear that existing CAs aren't breaking any rules just because I changed the draft.


jyasskin:
It was @clintwilson's spot, to be clear. :)

twifkak:
FYI this updates #377 for #383. (just for the auto-backlinks)

PR #404, opened 2019-03-13
go: Add strucured header writer

irori:
This adds ToString() method to the structured header types, and
uses it from signer.go.

### DIFF
--- a/go/signedexchange/structuredheader/writer.go
+++ b/go/signedexchange/structuredheader/writer.go
@@ -10,7 +10,7 @@ import (
 	"unicode"
 )
 
-func (ll ListOfLists) ToString() (string, error) {
+func (ll ListOfLists) String() (string, error) {
 	var b strings.Builder
 	if err := ll.serialize(&b); err != nil {
 		return "", err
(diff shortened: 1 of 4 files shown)


go/signedexchange/structuredheader/writer.go
```diff
@@ -0,0 +1,174 @@
...
+	"encoding/base64"
+	"errors"
+	"fmt"
+	"sort"
+	"strconv"
+	"strings"
+	"unicode"
+)
+
+func (ll ListOfLists) ToString() (string, error) {
```
hajimehoshi: `String` is more idiomatic in Go.

irori: Doesn't `String()` usually return just a `string`? This returns an `error` too.

hajimehoshi: Right, but using `To` is not common in Go.

There is an actual usecase of `String` that also returns error: https://godoc.org/golang.org/x/text/encoding#Decoder.String

irori: Got it, thanks for finding the example. Done.

PR #410, opened 2019-03-20
Change gen-signedexchange default to b3.

twifkak:


twifkak:
cc @jeffjose

jeffjose:
Can we update b2 -> b3 here too please? https://github.com/WICG/webpackage/blob/master/examples/firebase.json

twifkak:
@jeffjose Yep, that's in this PR.

jeffjose:
Ah yes, missed it the first time around. Thank you, @twifkak!

PR #414, opened 2019-03-27
go/bundle: Add support for the manifest section

irori:
Please take a look when you have time.

### DIFF
--- a/go/bundle/bundle.go
+++ b/go/bundle/bundle.go
@@ -142,7 +142,7 @@ type requestEntryWithOffset struct {
 type section interface {
 	Name() string
 	Len() int
-	Bytes() []byte
+	io.WriterTo
 }
 
 // staging area for writing index section
@@ -228,11 +228,12 @@ func (is *indexSection) Len() int {
 	return len(is.bytes)
 }
 
-func (is *indexSection) Bytes() []byte {
+func (is *indexSection) WriteTo(w io.Writer) (int64, error) {
 	if is.bytes == nil {
 		panic("indexSection must be Finalize()-d before calling Bytes()")
 	}
-	return is.bytes
+	n, err := w.Write(is.bytes)
+	return int64(n), err
 }
 
 // staging area for writing responses section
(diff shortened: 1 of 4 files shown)


go/bundle/bundle.go
```diff
@@ -836,25 +896,31 @@ func (b *Bundle) WriteTo(w io.Writer) (int64, error) {
 		return cw.Written, err
 	}
 
-	sos := []sectionOffset{
-		sectionOffset{"index", uint64(is.Len())},
-		sectionOffset{"responses", uint64(rs.Len())},
+	ms, err := newManifestSection(b.ManifestURL)
```
nyaxt: Can we make Manifest section optional?

irori: It's optional. If `b.ManifestURL` is nil, `newManifestSection` returns nil and the manifest section won't be emitted.

go/bundle/bundle.go
```diff
@@ -138,6 +139,12 @@ type requestEntryWithOffset struct {
 	Offset uint64
 }
 
+type section interface {
+	Name() string
+	Len() int
+	Bytes() []byte
+}
```
hajimehoshi: Can we use `WriteTo` instead of getter property interface?

https://golang.org/pkg/io/#WriterTo

go/bundle/bundle.go
nyaxt: Thanks for explanation. I think we want to make that clear from reading only this part of the code.
`newManifestSection` returning `nil, nil` is a surprising behavior to me.

Some ideas:
- `if url != ""` before newManifestSection
- Alternate namings? `tryCreateManifestSection`?
- Add comment explaining the behavior here?

irori: Thanks for the suggestions. Chenged to check `b.ManifestURL != nil` before newManifestSection.

go/bundle/bundle.go
irori: Done.

hajimehoshi: Ah sorry but I meant to remove section interface and use `io.WriterTo` instead

irori: Hmm. We need to write (name, length) pair and section content separately, so just a `WriterTo` doesn't suffice.

hajimehoshi: Oh I see.

In this case, I prefer to define e.g. `MarshalCbor` function that takes cbor's encoder to having getters, but that's optional.

irori: Let me merge this as-is, and do the refactoring separately.

PR #417, opened 2019-03-28
go: Add basic test for bundle.go

irori:


go/bundle/bundle_test.go
```diff
@@ -0,0 +1,58 @@
+package bundle_test
+
+import (
+	"bytes"
+	"net/http"
+	"net/url"
+	"reflect"
+	"testing"
+
+	. "github.com/WICG/webpackage/go/bundle"
```
nyaxt: can we use bundle.Bundle instead of `.` import?

hajimehoshi: I think using a dot in a test package is common, but I'm fine with both.

yzp0n: OK. Wasn't a strong preference

Issue #422, opened 2019-04-19
Avoiding Built-In Tracking in Signed Packages


johnwilander:
Hi! John Wilander from Apple's WebKit team here.

We are concerned with the privacy implications of User A and user B not getting the same package when they load the same webpage, or to put it another way, **personalized signed packages with cross-site tracking built in.**

### The Actors

* **The user.** This is the human who relies on the user agent to protect their privacy.
* **The user agent.** This is the web browser that tries to protect the user's privacy.
* **AdTech or adtech.example.** This is a website that the user engages with as first-party site but that has a financial interest in 1) knowing what the user does on other websites to augment its rich profile of the user and 2) individual targeting ads at the user on other websites, based on its rich profile of the user.
* **News or news.example.** This is a news website which wants its articles to be served as signed packages with the user agent's URL bar showing news.example.

### The Threat

1. The user does not want AdTech to be able to augment its profile of them while reading articles on news.example.
2. The user does not want AdTech's rich profile of them to influence the content or ads of articles on news.example.

### The Attack

This is how AdTech could foil the user agent's privacy protections with the current signed packages proposal:

News wants to take part in signed package loading but thinks the actual packaging is cumbersome and costly in terms of engineering resources.

AdTech has a financial incentive to help News get going with signed packages because the technology makes AdTech's services better. Because of this incentive, AdTech decides to offer News a more convenient way to do the packaging; it offers to pull unsigned articles directly from News's servers and do packaging for them. News just has to set up a signing service that AdTech can call to get signatures back, or just hand a signing key straight to AdTech. News sees the opportunity to reduce cost and takes the offer.

AdTech also has a financial incentive to identify the user on news.example to augment its profile of the user and to earn extra money by serving the user individually targeted ads, but it can't do so because the user's user agent is protecting the user's privacy. However, the request to get a signed News package is actually made to adtech.example, containing the user's AdTech cookies. To achieve their goals and earn more money, AdTech's decides to create news.example packages on the fly, bake in individually targeted ads plus an AdTech user ID for profile enrichment, and sign the whole thing with News's key.

This is a case of cross-site tracking. The user is on a news.example webpage, convinced that their user agent protects them from AdTech tracking them on this site, but instead they got a signed package with tracking built in.

### How the Attack Relates To Other Means of Cross-Site Tracking

[...]

PR #424, opened 2019-04-24
Commit the threat model from #422

jyasskin:
@johnwilander, here's an attempt to make your threat model a little more formal so that we can evaluate a range of attacks and mitigations against it. Did I capture roughly what the WebKit team is thinking? Thanks for engaging!

### DIFF
--- a/anti-tracking-threat-model.md
+++ b/anti-tracking-threat-model.md
@@ -156,7 +156,18 @@ signatures to check against.
 
 When requesting a signed package:
 
-1. The request chain must be credential-less. Maybe we can leverage some Fetch policy here.
-1. It must be an HTTP GET request.
-1. The request URL must not have a query string or fragment.
-1. The path of the package request must be the same as the path on the target domain.
+1. The request to the distributor must be credential-less. i.e. the [credentials
+   mode](https://fetch.spec.whatwg.org/#concept-request-credentials-mode) must
+   be `"omit"`. This prevents AdTech from learning the user's identity from its
+   first-party cookies.
+1. It must be an HTTP GET request to prevent, for example, a POST request body
+   from including a user ID.
+1. The request URL on the distributor must not have a query string. Fragments
+   aren't sent to the server and would be blocked after the redirect, if
+   necessary, by anti-tracking measures that are independent of web packages.
+1. The path of the package request must be the same as the path on the target
+   domain to prevent the distributor from encoding a user ID in the path.
+
+Note that this still allows AdTech to encode a user ID in the *signed* path and
+inject Javascript to decode it into local variables and `pushState()` the URL
+back to what the publisher's Javascript expects.
--- a/anti-tracking-threat-model.md
+++ b/anti-tracking-threat-model.md
@@ -1,5 +1,23 @@
 # Anti-tracking Threat Model
 
+## How the Attack Relates To Other Means of Cross-Site Tracking
+
+Often when we criticize new, technically distinct tracking vectors, we are told
+that “you can track users in so many ways so why care about this one?” In the
+case of signed packages we hear about means of tracking such as doctored links
+where cross-site tracking is built into the URL, or server-side exchanges of
+personally identifiable information such as users' email addresses.
+
+First, we don't think past mistakes and flaws in web technologies is a valid argument for why new web technologies should enable cross-site tracking.
+
+Second, WebKit is working hard to prevent cross-site tracking, including new
+limits and restrictions on old technologies. Piling on more such work is not
+acceptable to us.
+
+Finally, the success of new web technologies such as signed packages relies on
+better security and privacy guarantees than what we've had in the past. We want
+progression in this space, not the status quo.
+
 ## The Actors
 
 * **The user.** This is the human who relies on the user agent to protect their
@@ -53,24 +71,6 @@ AdTech user ID for profile enrichment, and sign the whole thing with News's key.
 This is a case of cross-site tracking. The user is on a `news.example` webpage,
 convinced that their user agent protects them from AdTech tracking them on this
 site, but instead they got a signed package with tracking built in.
-
-## How the Attack Relates To Other Means of Cross-Site Tracking
-
-Often when we criticize new, technically distinct tracking vectors, we are told
-that “you can track users in so many ways so why care about this one?” In the
-case of signed packages we hear about means of tracking such as doctored links
-where cross-site tracking is built into the URL, or server-side exchanges of
-personally identifiable information such as users' email addresses.
-
-First, we don't think past mistakes and flaws in web technologies is a valid argument for why new web technologies should enable cross-site tracking.
-
-Second, WebKit is working hard to prevent cross-site tracking, including new
-limits and restrictions on old technologies. Piling on more such work is not
-acceptable to us.
-
-Finally, the success of new web technologies such as signed packages relies on
-better security and privacy guarantees than what we've had in the past. We want
-progression in this space, not the status quo.
 
 ## Potential Mitigations and Fixes
 
--- a/anti-tracking-threat-model.md
+++ b/anti-tracking-threat-model.md
@@ -6,14 +6,17 @@
   privacy.
 * **The user agent.** This is the web browser that tries to protect the user's
   privacy.
-* **AdTech or `adtech.example`.** This is a website that the user engages with as
-  first-party site and that has a financial interest in
-  1. knowing what the user does on other websites to augment its rich profile of
-     the user and
-  2. individual targeting of ads, based on its rich profile of the user.
-* **News or `news.example`.** This is a news website which wants its articles to
-  be served as signed packages with the user agent's URL bar showing
-  `news.example`.
+* **Distributor.** An entity that delivers a signed package to either the user or another distributor.
+  * **AdTech or `adtech.example`.** This is a distributor that the user also
+    engages with as a first-party site and that has a financial interest in
+    1. knowing what the user does on other websites to augment its rich profile
+       of the user and
+    2. individual targeting of ads, based on its rich profile of the user.
+* **Publisher.** An entity that owns a domain and publishes content there. The
+  publisher may not actually author the content, but they put their name on it.
+  * **News or `news.example`.** This is a publisher for a news website which
+    wants its articles to be served as signed packages with the user agent's URL
+    bar showing `news.example`.
 
 ## The Threat
 
--- a/anti-tracking-threat-model.md
+++ b/anti-tracking-threat-model.md
@@ -1,8 +1,8 @@
 # Anti-tracking Threat Model
 
-WebKit and other browsers are trying to reduce websites' ability to track users
-across the internet. While the success of their effort remains to be seen, we
-don't want Web Packaging to make the problem any more difficult.
+WebKit and other browser engines are trying to reduce websites' ability to track
+users across the web. We don't want Web Packaging to make this effort any more
+difficult.
 
 This document contains a proposed threat model for the best-case outcome of that anti-tracking effort. We do not yet claim there is consensus about any of:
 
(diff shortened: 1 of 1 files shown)


anti-tracking-threat-model.md
```diff
@@ -0,0 +1,196 @@
...
+
+1. AdTech cannot manipulate the DNS system on a per-user basis.
+1. There is a limit to the complexity AdTech can convince News to add to its
+   serving infrastructure. For example, News is willing to ignore unknown query
+   parameters and fragments, but is not willing to ignore unexpected path
+   segments. TODO: Can we describe this limit any more precisely?
+
+## Attacker goals that we want to frustrate
+
+1. The user does not want AdTech to be able to augment its profile of them while
```
yoavweiss: Nit: the points are not written as attacker goals

anti-tracking-threat-model.md
```diff
@@ -0,0 +1,196 @@
...
+   `news.example`, perhaps by offering to handle the technical complications of
+   signing packages.
+   1. Alternately, News could set up a service that signs packages AdTech sends
+      it.
+1. When a user clicks a link from `adtech.example` to
+   `https://adtech.example/news.example.sxg`, they send identifying information
+   to the `adtech.example` server. This could be cookies or a user ID encoded in
+   the query, path, or even hostname.
+1. Instead of signing `news.example`'s content directly, AdTech embeds the
+   user's identity in that content and signs the result on the fly.
```
yoavweiss: With Server Timing, we have response headers that are directly readable from JS, so embedding the UID through them might be feasible even without on-the-fly signing. So we need to make sure that the distributor cannot add arbitrary non-signed headers to the internal response, as well as that the external response Server Timing headers are not exposed to the navigated page.
I believe that's already the case, tbh, but IMO it's worthwhile to explicitly note that.

anti-tracking-threat-model.md
```diff
@@ -0,0 +1,196 @@
...
+   mode](https://fetch.spec.whatwg.org/#concept-request-credentials-mode) must
+   be `"omit"`. This prevents AdTech from learning the user's identity from its
+   first-party cookies.
+1. It must be an HTTP GET request to prevent, for example, a POST request body
+   from including a user ID.
+1. The request URL on the distributor must not have a query string. Fragments
+   aren't sent to the server and would be blocked after the redirect, if
+   necessary, by anti-tracking measures that are independent of web packages.
+1. The path of the package request must be the same as the path on the target
+   domain to prevent the distributor from encoding a user ID in the path.
```
yoavweiss: It also seems like the distributor can use the referrer in order to include a UID (e.g. redirect through a unique path, and have a permissive ReferrerPolicy). Mitigation can be to enforce stricter ReferrerPolicy on redirects or in general.

anti-tracking-threat-model.md
```diff
@@ -0,0 +1,196 @@
+# Anti-tracking Threat Model
+
+WebKit and other browsers are trying to reduce websites' ability to track users
+across the internet. While the success of their effort remains to be seen, we
```
johnwilander: I'd say web rather than Internet since we're talking about web browsers here.

anti-tracking-threat-model.md
```diff
@@ -0,0 +1,196 @@
+# Anti-tracking Threat Model
+
+WebKit and other browsers are trying to reduce websites' ability to track users
```
johnwilander: browsers –> browser engines

anti-tracking-threat-model.md
```diff
@@ -0,0 +1,196 @@
...
+   we want to frustrate](#attacker-goals-that-we-want-to-frustrate).
+1. Whether the mitigations actually mitigate the attacks.
+1. Whether the costs of the mitigations are worth the benefits.
+1. Probably other things.
+
+Nonetheless, we feel it's important to start with a concrete proposal in mind
+when discussing how to evolve both this threat model and the Web Packaging
+proposals.
+
+Often when we criticize new, technically distinct tracking vectors, we are told
```
johnwilander: "We" has become confused here. It shouldn't be the same "we" criticizing tracking vectors and "we" are told.

anti-tracking-threat-model.md
```diff
@@ -0,0 +1,196 @@
...
+that “you can track users in so many ways so why care about this one?” In the
+case of signed packages we hear about means of tracking such as doctored links
+where cross-site tracking is built into the URL, or server-side exchanges of
+personally identifiable information such as users' email addresses.
+
+Browsers are working hard to prevent cross-site tracking, including with new
+limits and restrictions on old technologies. Web Packaging must not add to that work, although it's acceptable for Web Packaging to prevent tracking only after that other work has been done.
```
johnwilander: I don't think "that other work" is defined here. The other work is potentially years into the future and I don't think we should indicate that Web Packaging can punt on tracking issues in wait for such "other work."

anti-tracking-threat-model.md
```diff
@@ -0,0 +1,196 @@
...
+those mitigations should call out the use cases they break.
+
+## Attacker Capabilities
+
+1. AdTech has significant first-party traffic which means most users have an
+   `adtech.example` cookie holding a unique ID, even in browsers with
+   multi-keyed caches.
+1. AdTech can convince News to give them a `news.example` certificate's private
+   key or to tell a CA that AdTech has permission to receive certificates valid
+   to sign exchanges for `news.example`.
```
johnwilander: We should add that AdTech can help or convince News to set up a signing *service* as an alternative to handing over keys. With such a service, AdTech can create the package, calculate its hash, and send the hash over for signing, all on the fly.

anti-tracking-threat-model.md
```diff
@@ -0,0 +1,196 @@
...
+   This seems like an implausible capability at first glance, but publishers
+   routinely give CDNs this ability to terminate TLS traffic, and a
+   generally-trusted AdTech could convince publishers that it's merely doing
+   what a good CDN would, for cheaper.
+1. `adtech.example` can serve a link to `news.example` and expect users to click
+   on it and then browse around `news.example`.
+   1. This link can point to a resource that redirects to `news.example`.
+   1. This link can point to a signed exchange or web package containing
+      content signed by `news.example`'s certificate.
+1. AdTech can make any number of signatures with private keys it controls.
```
johnwilander: See my comment above on certificates vs private keys.

anti-tracking-threat-model.md
```diff
@@ -0,0 +1,196 @@
...
+implications for the rest of the web platform.
+
+## Sign identifying information into the package
+
+### The Attack
+
+1. AdTech convinces News to give AdTech a package-signing certificate for
+   `news.example`, perhaps by offering to handle the technical complications of
+   signing packages.
+   1. Alternately, News could set up a service that signs packages AdTech sends
```
johnwilander: Here the potential for a service is mentioned. See my comment on this above.

anti-tracking-threat-model.md
```diff
@@ -0,0 +1,196 @@
...
+#### Public signature repository
+
+Another potential mitigation would be some kind of public repository of
+signatures to check against.
+
+This does not prevent any attacks, but could make them detectable.
+
+#### Make package loads stateless
+
+When requesting a signed package:
```
johnwilander: I've only given casual thought to how a webpage can instruct the browser to restrict a load because it is for a signed package. Two ideas: 1) a /.well-known/ location or 2) a special HTTP redirect similar to an upgrade.

anti-tracking-threat-model.md
sleevi: I think there’s value in continuing to separate these concepts.

It would seem that your threat considerations consider the following scenarios:
1. AdTech creates a private key and obtains a certificate on News’ behalf (e.g. as CDNs commonly do)
1. News generates the key and then shares with AdTech
1. News holds the private key, but then signs on AdTech’s behalf (as commonly practiced by CDNs, typically called Split-TLS)
1. TLS Delegated Creds have a fourth model, where News generates the key, and then authorizes a key for AdTech

These each seem to represent different combinations, with the last being particularly relevant in the discussion of certificates and keys

anti-tracking-threat-model.md
sleevi: Isn’t this correct as is? For example, Brave shares the same Browser Engine as Edge and Chromium, but represent their own approaches to reduce tracking ability.

johnwilander: WebKit is an engine used in multiple browsers and web rendering applications. If the text should indeed refer to browsers, it should be Safari. But web technologies are mostly about engines.

anti-tracking-threat-model.md
jyasskin: I've elaborated this section with @sleevi's 4 possibilities. I think they mostly differ in how visible AdTech's misbehavior would be to News, and how long that misbehavior could keep affecting users after News terminated the relationship with AdTech.

I agree that we need to be precise about which of "certificate" or "private key" is correct in each location, but because they mean different things, I don't think we can talk about just one of them. I'm using phrasing that implies that a certificate signs the package, which matches https://wicg.github.io/webpackage/draft-yasskin-http-origin-signed-responses.html#cross-origin-cert-req, but strictly a private key (or keypair?) signs the package and the certificate just vouches for the authority of that key. I'm happy to take better wording.

anti-tracking-threat-model.md
jyasskin: Indeed. https://increment.com/security/approachable-threat-modeling/ would call these "Invariants", but it's not clear that would be intuitive to random folks reading this. I'll rephrase them.

anti-tracking-threat-model.md
jyasskin: I've rephrased this as a reference to the new wording in Attacker Capabilities and removed the sub-item.

anti-tracking-threat-model.md
jyasskin: It's a security requirement that the distributor can't add response headers to the post-redirect response. https://wicg.github.io/webpackage/draft-yasskin-http-origin-signed-responses.html#signed-headers mentions that idea, although it's in a section nobody's implementing. The application/signed-exchange format just can't represent such unsigned post-redirect response headers.

I don't think Server-Timing response headers on a 303 response get exposed to the post-redirect javascript. So I'm not sure there's really anything to say here.

anti-tracking-threat-model.md
jyasskin: As I'm saying in https://github.com/WICG/webpackage/pull/424#discussion_r279031642, I think we'll need to block that communication route on all redirects, so it's not special to web packages. Do you want me to add a separate attack describing it below?

jyasskin:
How is this doing as a first draft of the anti-tracking threat model? I'm hoping to send some more PRs digging into the proposed mitigations, but I'd like to have rough agreement on the base first.

martinthomson:
Jeffrey cited this PR in an email, and just reading through the text, I'm confused.  Are you really attempting to prevent two cooperating origins from exchanging information?

jyasskin:
My understanding from #422 is that @johnwilander and Safari's ITP2 are indeed trying to prevent two cooperating origins from exchanging information. Specifically, when following a link from AdTech to News, if AdTech can send AdTech's notion of the user's ID to News in any request where News also gets its own notion of the user's ID, then News can remember that association and use it augment AdTech's profile of the user or request user-specific ads from AdTech on all subsequent pages.

There may be limitations on AdTech's ability to get News to cooperate (see "Attacker Non-Capabilities"), but I haven't been able to get a description of those that I believe in.

hayatoito:
It seems I forgot to add a comment to this PR when adding `consider-closing` label.

[PR Triage]
I'm marking a label of "consider closing" for a PR which has not been active recently.
If there is no activity in a month, we will close PRs tentatively. Thanks!

hayatoito:
Closing.

masstez (commented):
Suggestions to improve the best results

masstez:
#713

PR #425, opened 2019-04-25
Use Case: Least Authority CDN Serving

davidstrauss:
While working on the security architecture for Drupal's planned auto-update features, I realized that many of our use cases could be solvable using SXG rather than rolling our own signing mechanism (as APT and Yum have) to support offline signatures for metadata from update servers. We already intend to distribute the metadata via a CDN that has appropriate certificates for drupal.org, but we also don't want the attack surface introduced by every POP being a potential point of compromise. Given that we control the client side in these scenarios, SXG represents one brightly lit path to trusting the metadata without relying solely on TLS or using baroque integrations with GnuPG.


draft-yasskin-webpackage-use-cases.md
```diff
@@ -372,6 +372,45 @@ Associated requirements:
...
+the origin without alteration. For these deployments, it would be useful to
+remove the authority of the CDN to serve arbitrary content.
+
+SXG offers a means to reducing CDN authority via HTTP clients configured to
+expect SXG for all responses, regardless of other trust. Applicable clients
+might include package managers, native mobile applications, desktop
+applications, and server applications.
+
+SXG would also offer a foundation for dynamically imposing such client
+behavior using a mechanism similar to HSTS. This would allow supporting clients
```
twifkak: I wonder if Must-Staple is a closer analogy to what is required. An origin could not enforce that their CDN includes a certain HTTP header on all responses, but they could enforce that the only certificate they bestow upon the CDN has a given extension. (As to whether that extension could or should be marked critical, I'm not sure.) Supporting clients would, when confronted with a TLS connection backed by such a cert, only honor responses that are signed with an SXG cert.

davidstrauss: > I wonder if Must-Staple is a closer analogy to what is required.

Perhaps I should rephrase the text, as the forward-looking idea is less about the mechanism (HSTS-style, Must-Staple, or otherwise) and more about allowing a site to inform clients to expect SXG responses. That expectation should be sticky or, at least, resistant to downgrade attacks.

I agree that an extension applied to the TLS certificate would be a technologically elegant way to solve the issue. I'd argue that such an extension should be non-critical, as that would allow existing clients to interact as before, while clients aware of the extension could enforce SXG on responses. SXG enforcement is a defense-in-depth against POP/proxy compromise, which I think is another argument for a non-critical designation.

I try to push back against new certificate extensions for practical reasons, though. They create a barrier to entry for people who implement or deploy the design. That raises some reasons I looked first to HSTS: it's broadly used, doesn't require external facilitation (e.g. by a CA), and has no up-front or ongoing costs to web site administrators.

> An origin could not enforce that their CDN includes a certain HTTP header on all responses, but they could enforce that the only certificate they bestow upon the CDN has a given extension.

This is true, but what I had in mind was the broader world of HSTS rather than just the header. In practice, qualifying HSTS headers get cached in browsers and bundled into the HSTS Preload List, both of which continue enforcement even if the header goes missing. (So, while a compromised POP could stop sending an SXG enforcement header, clients would still expect SXG.) I'm unsure whether this is the most desirable approach to SXG enforcement, but it's one established pattern.

davidstrauss:
> I believe I convinced @davidstrauss a couple months ago that his underlying use case was better served by signify or minisign.

We were convinced! We have a working implementation in testing now:

* https://github.com/drupal/php-signify
* https://github.com/mbaynton/drupal-signing-oracle

However, we extended the Signify design to support expiring intermediate keypairs. We did this for two reasons, which I'll mention here because SXG does support this use case (via X.509) better than Signify's design:

First, Drupal needs to sign artifacts produced in its online build infrastructure. These artifacts include builds of community-provided code that may be malicious. Signing these builds requires the key -- or the ability to sign -- to be available to this high-risk system. However, most cloud HSMs don't support ed25519. We don't have a natural place to deploy custom hardware for the Drupal Association, so we opted for using a VM as a signing oracle (to still create some isolation in case of build server compromise). Still, we wanted to anchor trust in something other than a key deployed to a VM in the cloud.

Second, Drupal lacks a natural key rotation interval akin to OpenBSD's six-month-ish release cycle. Drupal releases get supported for years, and users expect to be able to upgrade across fairly arbitrary point releases. This creates tension between (a) rotating across point releases (perhaps not every one) and potentially breaking the ability of older point releases to validate current builds versus (b) rotating across major releases and having keys live for 5+ years. It would also be possible to have older point releases "walk" their way to trusting the current build key, but that seemed fragile. An intermediate keypair makes rotation a question of operations effort versus security -- with no end-user impacts.

So, an intermediate keypair seemed like an elegant way to keep the root offline and support arbitrary rotations of the secrets in the deployed build infrastructure. (We currently hope to rotate intermediate keys quarterly.) Otherwise, we were quite happy with what Signify provided as a design.

We plan to secure the offline root by using YubiHSM hardware.

hayatoito:
Closing.

PR #428, opened 2019-05-01
Remove credentials from the certificate fetch.

jyasskin:
This removes one avenue by which a distributor could send its notion of
a user's ID to another origin, for server-side correlation with that
origin's cookies. That correlation would violate the goals in #422 and seems easy to block.

Thanks @kristoferbaxter and @twifkak for pointing this out.

I don't *think* it's useful to customize the content of a certificate chain to a particular user, especially since the hash of the certificate itself is locked in by the SXG's signature. But if anyone sees a way this'll break things, please speak up.

***
<a href="https://pr-preview.s3.amazonaws.com/jyasskin/webpackage/pull/428.html" title="Last updated on May 1, 2019, 8:05 PM UTC (1321699)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/webpackage/428/9634466...jyasskin:1321699.html" title="Last updated on May 1, 2019, 8:05 PM UTC (1321699)">Diff</a>

### DIFF
--- a/loading.bs
+++ b/loading.bs
@@ -706,6 +706,8 @@ reporting to a [=signed exchange report=] |report|, returns a
     :: "`none`"
     : [=request/mode=]
     :: "`cors`"
+    : [=request/credentials mode=]
+    :: "`omit`"
 1. Let |certResponse| be the result of [=fetching=] |certRequest|.
 1. Append the IP address of the server from which the user agent received the
     |certResponse| to |report|'s [=signed exchange report/cert server IP list=],


loading.bs
```diff
@@ -706,6 +706,8 @@ reporting to a [=signed exchange report=] |report|, returns a
     :: "`none`"
     : [=request/mode=]
     :: "`cors`"
+    : [=request/credentials mode=]
+    :: "`omit`"
```
twifkak: Pardon my naivete wrt the fetch monkeypatch,

https://fetch.spec.whatwg.org/#concept-request-credentials-mode says:
> Unless stated otherwise, it is "`omit`".

Is it stated otherwise somewhere? Or is duplicate to avoid confusion?

jyasskin: Argh, thanks for pointing that out. Yep, this PR is a no-op. Never mind!

kristoferbaxter: Thanks for taking a look anyway. I thought this value default was changed, https://github.com/whatwg/fetch/pull/585. 

It might still be worthwhile to ensure it's set to "omit".

twifkak: It seems like "same-origin" would address the threat mentioned in the PR description ("another origin"); is there another threat you had in mind?

kristoferbaxter: Making the value `omit` would only be for completeness sake – `same-origin` gives the same level of protection but unnecessarily sends cookies for same origin cert requests.

PR #429, opened 2019-05-02
Adds support for URI input to dump-signedexchange

EverlastingBugstopper:
This PR makes some quality of life improvements for the dump-signedexchange tool

- Adds support for reading a signed exchange from an HTTPS endpoint rather than from the file system. This feels much more natural to use and allows for faster debugging.
- Allows for more granular control over the output. Generally when I dump a signed exchange, I don't want to see the entire HTML, I'm interested in the signature and the headers.

EDIT:
- The tool will also warn the user if -i or -uri is not passed instead of simply hanging indefinitely

EDIT 2:
- Allows the version to be specified with the `-version` flag which defaults to the latest version
- Allows a header such as `AMP-Cache-Transform: google;v=1` to be sent along with the `-uri` request with the `-requestHeader` flag
- Fixes a few instances of `singedexchange` and makes them `signedexchange`

EDIT 3:
- Added info about json flag to README
- Allows more than one header to be passed with -requestHeader flag
- Mime type factored to version.go
- Input flags take precedence over pipe

### DIFF
--- a/go/signedexchange/README.md
+++ b/go/signedexchange/README.md
@@ -148,11 +148,50 @@ You can dump the content of your sxg file by `dump-signedexchange`. If you want 
 dump-signedexchange -i example.org.hello.sxg
 ```
 
-If `-verify` command-line flag is specified, `dump-signedexchange` checks if the signed exchange is valid.
+If the `-json` flag is passed, the output will be in JSON.
+
+```
+dump-signedexchange -i example.org.hello.sxg -json
+```
+
+You can also dump the content of a signed exchange from a URI. If you want to see the content of the signed exchange you're hosting at https://example.org/hello.html, run this command. By default this will request the latest version.
+
+```
+dump-signedexchange -uri https://example.org/hello.html
+```
+
+If the specified URI requires a special header to serve a signed exchange, you can pass request headers via the `-requestHeader` flag. The header key and value should be separated by a `:`.
+
+```
+dump-signedexchange -uri https://example.org/hello.html -requestHeader AMP-Cache-Transform:any -requestHeader "foo:bar"
+```
+
+When the `-uri` flag is passed to `dump-signedexchange`, you can specify the sxg version to request by passing a `-version` flag. For instance, if you wanted to request a `1b2` signed exchange, you would run the following command. By default, the version is `1b3`.
+
+```
+dump-signedexchange -uri https://example.org/hello.html -version=1b2
+```
+
+`dump-signedexchange` can also operate on piped input. For instance, you could run the following command to retrieve a b3 signed exchange.
+
+```
+curl -H "AMP-Cache-Transform:any" -H "Accept:application/signed-exchange;v=b3" https://example.org/hello.html | dump-signedexchange
+```
+
+`dump-signedexchange` can print the information you want about your signed exchange. By default, both the headers and the payload are printed, but they can be suppressed by passing `-headers=false` and `-payload=false`.
+
+If you would like only the signature to be printed, pass the `-signature` flag.
+
+```
+dump-signedexchange -i example.org.hello.sxg -signature
+```
+
+If the `-verify` command-line flag is specified, `dump-signedexchange` checks if the signed exchange is valid.
 
 By default, `dump-signedexchange` fetches certificate chain from network (from the URL you specified as `-certUrl` parameter of `gen-signedexchange`). But if `-cert filename` flag is given, `dump-signedexchange` reads certificates from `filename`.
 
 For example, If you want to verify `example.org.hello.sxg` using certificates in `cert.cbor`, run this command.
+
 ```
 dump-signedexchange -i example.org.hello.sxg -verify -cert cert.cbor
 ```
--- a/go/signedexchange/cmd/dump-signedexchange/main.go
+++ b/go/signedexchange/cmd/dump-signedexchange/main.go
@@ -1,40 +1,101 @@
 package main
 
 import (
+	"bytes"
 	"encoding/json"
 	"flag"
 	"fmt"
 	"io"
 	"io/ioutil"
 	"log"
+	"net/http"
 	"os"
+	"strings"
 	"time"
 
 	"github.com/WICG/webpackage/go/signedexchange/structuredheader"
 
 	"github.com/WICG/webpackage/go/signedexchange"
+	"github.com/WICG/webpackage/go/signedexchange/version"
 )
+
+type headerArgs []string
+
+func (h *headerArgs) String() string {
+	return fmt.Sprintf("%v", *h)
+}
+
+func (h *headerArgs) Set(value string) error {
+	*h = append(*h, value)
+	return nil
+}
+
+var latestVersion = string(version.AllVersions[len(version.AllVersions)-1])
 
 var (
-	flagInput     = flag.String("i", "", "Signed-exchange input file")
-	flagSignature = flag.Bool("signature", false, "Print only signature value")
-	flagVerify    = flag.Bool("verify", false, "Perform signature verification")
 	flagCert      = flag.String("cert", "", "Certificate CBOR file. If specified, used instead of fetching from signature's cert-url")
+	flagHeaders   = flag.Bool("headers", true, "Print headers")
+	flagFilename  = flag.String("i", "", "Signed-exchange input file")
 	flagJSON      = flag.Bool("json", false, "Print output as JSON")
+	flagPayload   = flag.Bool("payload", true, "Print payload")
+	flagSignature = flag.Bool("signature", false, "Print only signature value")
+	flagURI       = flag.String("uri", "", "Signed-exchange uri")
+	flagVerify    = flag.Bool("verify", false, "Perform signature verification")
+	flagVersion   = flag.String("version", latestVersion, "Signed exchange version")
+
+	flagRequestHeader = headerArgs{}
 )
 
+func init() {
+	flag.Var(&flagRequestHeader, "requestHeader", "Request header arguments")
+}
+
 func run() error {
-	in := os.Stdin
-	if *flagInput != "" {
-		var err error
-		in, err = os.Open(*flagInput)
+	fi, err := os.Stdin.Stat()
+	if err != nil {
+		return err
+	}
+	var e *signedexchange.Exchange
+	var in io.Reader = nil
+	if *flagFilename != "" { // read sxg from filename
+		f, err := os.Open(*flagFilename)
 		if err != nil {
 			return err
 		}
-		defer in.Close()
+		defer f.Close()
+		in = f
+	} else if *flagURI != "" { // read sxg from network
+		client := http.DefaultClient
+		req, err := http.NewRequest("GET", *flagURI, nil)
+		if err != nil {
+			return err
+		}
+		ver, ok := version.Parse(*flagVersion)
+		if !ok {
+			return fmt.Errorf("failed to parse version %q", *flagVersion)
+		}
+		mimeType := ver.MimeType()
+		req.Header.Add("Accept", mimeType)
+		for _, h := range flagRequestHeader {
+			chunks := strings.SplitN(h, ":", 2)
+			req.Header.Add(strings.TrimSpace(chunks[0]), strings.TrimSpace(chunks[1]))
+		}
+		resp, err := client.Do(req)
+		if err != nil {
+			return err
+		}
+		in = resp.Body
+		defer resp.Body.Close()
+	} else if (fi.Mode() & os.ModeCharDevice) == 0 { // read sxg from pipe
+		in = os.Stdin
 	}
 
-	e, err := signedexchange.ReadExchange(in)
+	if in == nil {
+		flag.PrintDefaults()
+		return nil
+	}
+
+	e, err = signedexchange.ReadExchange(in)
 	if err != nil {
 		return err
 	}
@@ -51,18 +112,22 @@ func run() error {
 
 	if *flagSignature {
 		fmt.Println(e.SignatureHeaderValue)
-		return nil
-	}
-
-	if *flagVerify {
-		if err := verify(e, certFetcher, verificationTime); err != nil {
-			return err
+	} else {
+		if *flagHeaders {
+			e.PrettyPrintHeaders(os.Stdout)
 		}
-		fmt.Println()
-	}
 
-	e.PrettyPrintHeaders(os.Stdout)
-	e.PrettyPrintPayload(os.Stdout)
+		if *flagPayload {
+			e.PrettyPrintPayload(os.Stdout)
+		}
+
+		if *flagVerify {
+			fmt.Println()
+			if err := verify(e, certFetcher, verificationTime); err != nil {
+				return err
+			}
+		}
+	}
 
 	return nil
 }
@@ -116,8 +181,14 @@ func jsonPrintHeaders(e *signedexchange.Exchange, certFetcher signedexchange.Cer
 		sigs,
 		e,
 	}
-	s, _ := json.MarshalIndent(f, "", "  ")
-	w.Write(s)
+	buf := new(bytes.Buffer)
+	enc := json.NewEncoder(buf)
+	enc.SetEscapeHTML(false)
+	enc.SetIndent("", "   ")
+	if err := enc.Encode(&f); err != nil {
+		return err
+	}
+	w.Write(buf.Bytes())
 
 	return nil
 }
--- a/go/signedexchange/signedexchange.go
+++ b/go/signedexchange/signedexchange.go
@@ -236,7 +236,7 @@ func (e *Exchange) decodeExchangeHeaders(dec *cbor.Decoder) error {
 			return fmt.Errorf("signedexchange: failed to decode top-level array header: %v", err)
 		}
 		if n != 2 {
-			return fmt.Errorf("singedexchange: length of header array must be 2 but %d", n)
+			return fmt.Errorf("signedexchange: length of header array must be 2 but %d", n)
 		}
 		if err := e.decodeRequestMap(dec); err != nil {
 			return err
--- a/go/signedexchange/verifier.go
+++ b/go/signedexchange/verifier.go
@@ -64,7 +64,7 @@ func extractSignatureFields(pi structuredheader.ParameterisedIdentifier) (*Signa
 	return sig, nil
 }
 
-// CertFetcher takes certificat URL and returns certificate bytes in
+// CertFetcher takes certificate URL and returns certificate bytes in
 // application/cert-chain+cbor format.
 type CertFetcher = func(url string) ([]byte, error)
 
--- a/go/signedexchange/version/version.go
+++ b/go/signedexchange/version/version.go
@@ -15,7 +15,7 @@ const (
 
 const HeaderMagicBytesLen = 8
 
-var AllVersions = []Version {
+var AllVersions = []Version{
 	Version1b1,
 	Version1b2,
 	Version1b3,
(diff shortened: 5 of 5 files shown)


EverlastingBugstopper:
cc: @gabbifish @twifkak

go/signedexchange/cmd/dump-signedexchange/main.go
```diff
@@ -13,42 +14,64 @@ import (
...
-			return err
+		if *flagVerify {
+			fmt.Println()
+			if err := verify(e); err != nil {
+				return err
+			}
 		}
-		fmt.Println()
+	} else {
+		fmt.Println("Need to pass -uri with URI of sxg or -i with file location of sxg")
```
twifkak: I use `dump-signedexchange` at the end of a pipe sometimes. I'd prefer if you updated this to allow the case where stdin != /dev/pty or whatever. Alternatively, maybe we could make `-i -` a valid syntax if it isn't already.

EverlastingBugstopper: I didn't know you could use `dump-signedexchange` like that! Makes a lot of sense though. I've added that functionality back and documented a use case in the README.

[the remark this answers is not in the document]
EverlastingBugstopper: Turns out the `flag` library can print the usage. I've changed this message to simply print the usage, though it doesn't return anything. Do you think that pattern is OK even though the function doesn't return an error?

go/signedexchange/cmd/dump-signedexchange/main.go
```diff
@@ -5,50 +5,81 @@ import (
...
-		var err error
-		in, err = os.Open(*flagInput)
+	fi, err := os.Stdin.Stat()
+	var e *signedexchange.Exchange
+	if *flagFilename != "" {
+		in, err := os.Open(*flagFilename)
 		if err != nil {
 			return err
 		}
 		defer in.Close()
```
twifkak: These four lines are the same in each branch of the if. Maybe just `var in` outside the `if` and set it in each branch.

go/signedexchange/cmd/dump-signedexchange/main.go
twifkak: Nice! Thanks.

go/signedexchange/README.md
```diff
@@ -148,11 +148,32 @@ You can dump the content of your sxg file by `dump-signedexchange`. If you want
 dump-signedexchange -i example.org.hello.sxg
 ```
 
+You can also dump the content of a signed exchange from a URI. If you want to see the content of the signed exchange you're hosting at https://example.org/hello.html, run this command. By default this will request a b3 signed exchange.
+
+```
+dump-signedexchange -uri https://example.org/hello.html
+```
+
+`dump-signedexchange` can also operate on piped input. For instance, you could run the following command to retrieve a b2 package.
```
nyaxt: Should we use b3 since it is now the default version?

go/signedexchange/README.md
```diff
@@ -148,11 +148,32 @@ You can dump the content of your sxg file by `dump-signedexchange`. If you want
...
+
+```
+curl -H "AMP-Cache-Transform:any" -H "Accept:application/signed-exchange;v=b2" https://example.org/hello.html | dump-signedexchange
```
nyaxt: ditto

go/signedexchange/cmd/dump-signedexchange/main.go
```diff
@@ -5,50 +5,81 @@ import (
...
+		e, err = signedexchange.ReadExchange(in)
+		if err != nil {
+			return err
+		}
+	} else if *flagURI != "" {
+		client := http.DefaultClient
+		req, err := http.NewRequest("GET", *flagURI, nil)
+		req.Header.Add("Accept", "application/signed-exchange;v=b3")
+		req.Header.Add("AMP-Cache-Transform", "google;v=1")
```
nyaxt: Would you specify `any` here or add a commandline flag to modify this? (I'd like to avoid having `google` as default)

go/signedexchange/cmd/dump-signedexchange/main.go
```diff
@@ -5,50 +5,81 @@ import (
...
 		}
 		defer in.Close()
+		e, err = signedexchange.ReadExchange(in)
+		if err != nil {
+			return err
+		}
+	} else if *flagURI != "" {
+		client := http.DefaultClient
+		req, err := http.NewRequest("GET", *flagURI, nil)
+		req.Header.Add("Accept", "application/signed-exchange;v=b3")
```
nyaxt: Would you make `b3` part configurable via command line flag?

go/signedexchange/cmd/dump-signedexchange/main.go
```diff
@@ -5,50 +5,81 @@ import (
...
+		resp, err := client.Do(req)
+		if err != nil {
+			return err
+		}
+		defer resp.Body.Close()
+		e, err = signedexchange.ReadExchange(resp.Body)
+		if err != nil {
+			return err
+		}
+	} else if (fi.Mode() & os.ModeCharDevice) == 0 {
```
nyaxt: I think we can rely on `strings.TrimSpace(*flagFilename) == "-"` instead

go/signedexchange/cmd/dump-signedexchange/main.go
twifkak: Maybe even best to leave AMP-Cache-Transform out of this binary entirely and have a generic -requestHeader flag.

go/signedexchange/cmd/dump-signedexchange/main.go
twifkak: Either works for me, but I like the current approach the best. Allows pipe usage to work without specifying `-i` at all, while still presenting the usage info when not in a pipe.

go/signedexchange/cmd/dump-signedexchange/main.go
```diff
@@ -5,50 +5,81 @@ import (
...
+		in := os.Stdin
+		defer in.Close()
+		e, err = signedexchange.ReadExchange(in)
+		if err != nil {
+			return err
+		}
 	}
+	if e != nil {
```
hajimehoshi: I prefer early returns:

```
if e == nil {
    flag.PrintDefaults()
    return nil
}
```

go/signedexchange/cmd/dump-signedexchange/main.go
EverlastingBugstopper: This should be done, if there is an obvious better way to do this, please let me know.

go/signedexchange/cmd/dump-signedexchange/main.go
EverlastingBugstopper: I've kept this as is, input can come from the pipe, the -i flag or the -uri flag

EverlastingBugstopper:
Comments should be addressed at this point, let me know if there are further improvements that should be made.

go/signedexchange/cmd/dump-signedexchange/main.go
```diff
@@ -1,53 +1,122 @@
...
+func getAcceptHeader(value string) (string, error) {
+	v, parsed := version.Parse(value)
+	if parsed {
+		return fmt.Sprintf("application/signed-exchange;v=%s", v[1:]), nil
+	}
+	return "", fmt.Errorf("Could not parse version string \"%s\"", value)
+}
+
+func getRequestHeader(value string) ([]string, error) {
+	h := strings.Split(value, ":")
```
twifkak: Use [`strings.SplitN`](https://golang.org/pkg/strings/#SplitN), as it's possible for values to contain colons (e.g. `Location` or `Link`).

go/signedexchange/cmd/dump-signedexchange/main.go
```diff
@@ -1,53 +1,122 @@
...
+		if err != nil {
+			return err
+		}
+		in = resp.Body
+		defer resp.Body.Close()
+		receivedInput = true
 	}
 
-	e, err := signedexchange.ReadExchange(in)
+	if !receivedInput {
```
twifkak: (opt, nit) Set `in` to `nil` at declaration time and compare to `nil` here.

EverlastingBugstopper:
What do people here think about defaulting `-payload` to false and `-verify` to true? When I use this tool I only really care about the headers/verification and don't want the payload printed. I feel that's probably typical and those would make sensible defaults. That said, I don't want to change the default behavior of the tool w/o explicit buy-in.

ithinkihaveacat:
Quick comment, haven’t looked at this PR closely: with these changes, is it possible to cleanly output just the payload? Am wondering because this is something I’m interested in extracting (to pass to other tooling).

Having convenient access to the payload was part of the original purpose of my PR #428, but I think after the comments are addressed it won’t output the payload at all, so it would be nice to be able to get it via this PR ;-). (Don’t care whether it’s the default or not.)

EverlastingBugstopper:
> Quick comment, haven’t looked at this PR closely: with these changes, is it possible to cleanly output just the payload? 

Yes! You can simply pass `-headers=false` and only the payload will be printed :)

go/signedexchange/cmd/dump-signedexchange/main.go
```diff
@@ -1,53 +1,118 @@
...
+	flagURI           = flag.String("uri", "", "Signed-exchange uri")
+	flagVerify        = flag.Bool("verify", false, "Perform signature verification")
+	flagVersion       = flag.String("version", latestVersion, "Signed exchange version")
+	flagRequestHeader = flag.String("requestHeader", "", "Request header for signed exchange")
 )
 
+func getAcceptHeader(value string) (string, error) {
+	v, parsed := version.Parse(value)
+	if parsed {
+		return fmt.Sprintf("application/signed-exchange;v=%s", v[1:]), nil
```
nyaxt: Would you factor this out as `func (v Version) MimeType() string` in `version.go`?

go/signedexchange/cmd/dump-signedexchange/main.go
```diff
@@ -1,53 +1,118 @@
...
-	flagCert      = flag.String("cert", "", "Certificate CBOR file. If specified, used instead of fetching from signature's cert-url")
+	flagCert          = flag.String("cert", "", "Certificate CBOR file. If specified, used instead of fetching from signature's cert-url")
+	flagHeaders       = flag.Bool("headers", true, "Print headers")
+	flagFilename      = flag.String("i", "", "Signed-exchange input file")
+	flagPayload       = flag.Bool("payload", true, "Print payload")
+	flagSignature     = flag.Bool("signature", false, "Prints only signature value")
+	flagURI           = flag.String("uri", "", "Signed-exchange uri")
+	flagVerify        = flag.Bool("verify", false, "Perform signature verification")
+	flagVersion       = flag.String("version", latestVersion, "Signed exchange version")
+	flagRequestHeader = flag.String("requestHeader", "", "Request header for signed exchange")
```
nyaxt: Would you do this in a way similar to https://github.com/WICG/webpackage/blob/master/go/signedexchange/cmd/gen-signedexchange/main.go#L52 ?

go/signedexchange/cmd/dump-signedexchange/main.go
```diff
@@ -1,53 +1,118 @@
...
+	} else {
+		if *flagHeaders {
+			e.PrettyPrintHeaders(os.Stdout)
+		}
 
-	if *flagVerify {
-		if err := verify(e); err != nil {
-			return err
+		if *flagPayload {
+			e.PrettyPrintPayload(os.Stdout)
```
irori: Currently this tool prints mice-decoded payload if `-verify` is given (because `verify()` decodes the payload), but after this patch payload is not decoded at this point.

I know this is confusing and we should fix it (#431), but can we keep the current behavior for now?

go/signedexchange/cmd/dump-signedexchange/main.go
```diff
@@ -1,53 +1,118 @@
...
-	if *flagInput != "" {
-		var err error
-		in, err = os.Open(*flagInput)
+	fi, err := os.Stdin.Stat()
+	if err != nil {
+		return err
+	}
+	var e *signedexchange.Exchange
+	var in io.Reader = nil
+	if (fi.Mode() & os.ModeCharDevice) == 0 { // read sxg from pipe
```
irori: Would you move this branch to the last? I think the flags should take precedence.

go/signedexchange/cmd/dump-signedexchange/main.go
EverlastingBugstopper: I believe this behavior is unchanged

EverlastingBugstopper:
@irori @nyaxt comments should be addressed

EverlastingBugstopper:
The newly added `-json` flag coerces the strings `<` `>` and `&` to UTF-8. For instance, when outputting the `Link` header it will look like this:

```    "Link": [
      "\u003chttps://cdn.ampproject.org/v0.js\u003e;rel=preload;as=script"
    ],
```

I have a fix that will print the JSON w/proper encoding that I can include in this PR if y'all think it would be beneficial.

ithinkihaveacat:
@EverlastingBugstopper Ah, I didn't notice that! Yes I think that would be good to fix here (or in a different PR). It seems like the reason to encode those characters is to try to protect against problems that result from embedding JSON within HTML but that seems a bit unlikely in this case.

go/signedexchange/README.md
```diff
@@ -148,11 +148,50 @@ You can dump the content of your sxg file by `dump-signedexchange`. If you want
...
+dump-signedexchange -i example.org.hello.sxg -json
+```
+
+You can also dump the content of a signed exchange from a URI. If you want to see the content of the signed exchange you're hosting at https://example.org/hello.html, run this command. By default this will request the latest version.
+
+```
+dump-signedexchange -uri https://example.org/hello.html
+```
+
+If the specified URI requires a special header to serve a signed exchange, you can pass request headers via the -requestHeader flag. The header key and value should be separated by a `:` and the value should be enclosed with double quotations `""`
```
irori: > the value should be enclosed with double quotations `""`

This might be confusing, as `dump-signedexchange` itself doesn't interpret quotations. Quotations may be needed by command-line shells, but exact rules for quotation / escaping could be different by shells (so it's probably out of scope of this document).

go/signedexchange/cmd/dump-signedexchange/main.go
```diff
@@ -7,34 +7,92 @@ import (
...
+	if err != nil {
+		return err
+	}
+	var e *signedexchange.Exchange
+	var in io.Reader = nil
+	if *flagFilename != "" { // read sxg from filename
+		in, err = os.Open(*flagFilename)
 		if err != nil {
 			return err
 		}
```
irori: `defer in.Close()` here?

go/signedexchange/README.md
ithinkihaveacat: I'm not sure if I'm reading the docs right, but the examples don't seem to match the docs—?

The examples seem to match curl's syntax (aside from the actual switch name), which seems like a good approach--well-understood by developers, and various tools seem to now output curl syntax for debugging (e.g. Chrome and Safari developer tools).

go/signedexchange/cmd/dump-signedexchange/main.go
EverlastingBugstopper: Not sure exactly what needs to happen here as `io.Reader has no field or method Close`

go/signedexchange/README.md
EverlastingBugstopper: @ithinkihaveacat The examples should match? Could you point out what specifically is confusing? After this PR the tool will be able to take sxg input from a curl command and also by passing the URI via flag

EverlastingBugstopper:
I've pushed fix for JSON encoding (don't feel a separate PR is necessary, the change is small) and removed the unnecessary `""` pointers from the README

go/signedexchange/cmd/dump-signedexchange/main.go
irori: Ah, then how about this:
```
f, err := os.Open(*flagFilename)
if err != nil {
  return err
}
defer f.Close()
in = f
```

go/signedexchange/README.md
```diff
@@ -148,11 +148,50 @@ You can dump the content of your sxg file by `dump-signedexchange`. If you want
...
+dump-signedexchange -uri https://example.org/hello.html -requestHeader AMP-Cache-Transform:any -requestHeader "foo:bar"
+```
+
+When the `uri` flag is passed to `dump-signedexchange`, you can specify the sxg version to request by passing a `version` flag. For instance, if you wanted to request a `1b2` package, you would run the following command. By default, the version is `1b3`.
+
+```
+dump-signedexchange -uri https://example.org/hello.html -version=1b2
+```
+
+`dump-signedexchange` can also operate on piped input. For instance, you could run the following command to retrieve a b3 package.
```
irori:
```suggestion
`dump-signedexchange` can also operate on piped input. For instance, you could run the following command to retrieve a b3 signed exchange.
```

go/signedexchange/README.md
```diff
@@ -148,11 +148,50 @@ You can dump the content of your sxg file by `dump-signedexchange`. If you want
...
+dump-signedexchange -i example.org.hello.sxg -json
+```
+
+You can also dump the content of a signed exchange from a URI. If you want to see the content of the signed exchange you're hosting at https://example.org/hello.html, run this command. By default this will request the latest version.
+
+```
+dump-signedexchange -uri https://example.org/hello.html
+```
+
+If the specified URI requires a special header to serve a signed exchange, you can pass request headers via the -requestHeader flag. The header key and value should be separated by a `:`
```
irori:
```suggestion
If the specified URI requires a special header to serve a signed exchange, you can pass request headers via the `-requestHeader` flag. The header key and value should be separated by a `:`.
```

go/signedexchange/README.md
```diff
@@ -148,11 +148,50 @@ You can dump the content of your sxg file by `dump-signedexchange`. If you want
...
+
+```
+dump-signedexchange -uri https://example.org/hello.html -requestHeader AMP-Cache-Transform:any -requestHeader "foo:bar"
+```
+
+When the `uri` flag is passed to `dump-signedexchange`, you can specify the sxg version to request by passing a `version` flag. For instance, if you wanted to request a `1b2` package, you would run the following command. By default, the version is `1b3`.
```
irori:
```suggestion
When the `-uri` flag is passed to `dump-signedexchange`, you can specify the sxg version to request by passing a `-version` flag. For instance, if you wanted to request a `1b2` signed exchange, you would run the following command. By default, the version is `1b3`.
```

go/signedexchange/README.md
```diff
@@ -148,11 +148,50 @@ You can dump the content of your sxg file by `dump-signedexchange`. If you want
...
+
+`dump-signedexchange` can also operate on piped input. For instance, you could run the following command to retrieve a b3 package.
+
+```
+curl -H "AMP-Cache-Transform:any" -H "Accept:application/signed-exchange;v=b3" https://example.org/hello.html | dump-signedexchange
+```
+
+`dump-signedexchange` can print the information you want about your signed exchange. By default, both the headers and the payload are printed, but they can be suppressed by passing `-headers=false` and `-payload=false`.
+
+If you would like only the signature to be printed, pass the `signature` flag.
```
irori:
```suggestion
If you would like only the signature to be printed, pass the `-signature` flag.
```

EverlastingBugstopper:
@irori Comments addressed

PR #444, opened 2019-06-14
dump-certurl: Add check for certificate's validity period

irori:
Print warning or error message if main certificate has a validity period
longer than 90 days.


irori:
cc: @yutakahirano

PR #448, opened 2019-06-28
CBOR header key must not contain UPPERCASE alphabet

kumagi:
https://tools.ietf.org/html/draft-yasskin-http-origin-signed-responses-05#section-3.2

```
For each response header field, the header field's lowercase name
      as a byte string to the header field's value as a byte string.
```

We should check response header's key does not contains uppercase name along with this spec.

### DIFF
--- a/go/signedexchange/signedexchange.go
+++ b/go/signedexchange/signedexchange.go
@@ -137,6 +137,10 @@ func (e *Exchange) decodeRequestMap(dec *cbor.Decoder) error {
 		if err != nil {
 			return err
 		}
+		key_str := string(key)
+		if key_str != strings.ToLower(key_str) {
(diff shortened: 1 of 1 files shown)


go/signedexchange/signedexchange.go
```diff
@@ -192,6 +192,9 @@ func (e *Exchange) decodeResponseMap(dec *cbor.Decoder) error {
 		if err != nil {
 			return err
 		}
+		if string(key) != strings.ToLower(string(key)) {
```
irori: Please assign `string(key)` to a temporary variable and use it here and line 209 ([]byte -> string conversion makes a copy).

kumagi: OK, I'll assign it into `key_str`.

kumagi:
I appended the same logic into `decodeRequestMap` too.

PR #452, opened 2019-07-08
Improve impl 03

jyasskin:
Preview at https://jyasskin.github.io/webpackage/improve-impl-03/draft-yasskin-httpbis-origin-signed-exchanges-impl.html. Have I missed anything in describing how Chrome uses the b3 format? I'll publish -impl-03 on July 21.


jyasskin:
Thanks @irori. That's fixed now.

PR #460, opened 2019-07-18
go/bundle: Move decoder / encoder code to separate files

irori:
This patch splits `bundle.go` into three files:

- `bundle.go` - definition of common data types and some utility functions
- `encoder.go` - bundle generation code
- `decoder.go` - bundle parsing code

Code moves only, no functional changes.


irori:
cc: @toyoshim @yutakahirano

irori:
Thanks for the quick review!

PR #450, opened 2019-07-05
Update the bundle format

jyasskin:
Preview at https://jyasskin.github.io/webpackage/update-bundle-format/draft-yasskin-wpack-bundled-exchanges.html.

Sorry for the late mail: I belatedly realized the [IETF deadline](https://datatracker.ietf.org/meeting/105/important-dates/) is Monday, and it’d be nice to have an update to the bundle format published then. We’ll be able to publish another revision at the beginning of the IETF week.

This update includes:

* An invariant fallback URL, like signed exchanges have.
* A version number, so we can easily know to fall back to a redirect.
* Some infrastructure to identify what kind of error broke the parse, which can feed into both Network Error Logging and #397’s discussion of when to fall back.
* The [index](https://jyasskin.github.io/webpackage/update-bundle-format/draft-yasskin-wpack-bundled-exchanges.html#index-section) maps URLs to a Variants value + a list of the responses for each possible Variant-Key.
* A new [signatures](https://jyasskin.github.io/webpackage/update-bundle-format/draft-yasskin-wpack-bundled-exchanges.html#signatures-section) section allows authorities to vouch for particular subsets of the bundle. For cross-origin trust, which I think will get defined in the loading spec, I think the requirement that the validity-url is same-origin is going to force us to make those single-origin subsets, even when there might be a certificate that is trusted to sign multiple origins. I’ve run the basic structure by @sleevi who liked that the signed fields are all together in a byte string.

PR #461, opened 2019-07-18
go/bundle: Add support for PrimaryURL (aka fallbackURL)

irori:
- Adds `PrimaryURL` field to `Bundle` struct, and adds parsing / serializing code for that
- Updates loadMetadata() to reflect the spec changes of #450
  - Now it returns a fallback URL if available
  - Updated spec ref comments

### DIFF
--- a/go/bundle/decoder.go
+++ b/go/bundle/decoder.go
@@ -39,6 +39,7 @@ func FindSection(sos []sectionOffset, name string) (sectionOffset, uint64, bool)
 
 type meta struct {
 	version        version.Version
+	primaryURL     *url.URL
 	sectionOffsets []sectionOffset
 	sectionsStart  uint64
 	manifestURL    *url.URL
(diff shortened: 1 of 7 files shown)


irori:
cc: @toyoshim @yutakahirano

go/bundle/decoder.go
```diff
@@ -303,134 +304,158 @@ var knownSections = map[string]struct{}{
 }
 
 // https://wicg.github.io/webpackage/draft-yasskin-dispatch-bundled-exchanges.html#load-metadata
-func loadMetadata(bs []byte) (*meta, error) {
+func loadMetadata(bs []byte) (*meta, error, *url.URL) {
```
nyaxt: I'd like to keep `error` as the last return arg list.

two ideas:
1. reorder the return arg list, or
2. create a struct which implements error interface, which allows fallbackUrl extraction

irori: Created LoadMetadataError struct which embeds an error interface.

irori:
Thank you for the review!

PR #486, opened 2019-08-16
go: Refactor certchain.go

irori:
- `CertChainItem` struct is renamed to `AugmentedCertificate` (to match the spec).
- Added methods that {encode,decode} an `AugmentedCertificate` {to,from}
  `cbor.{Encoder,Decoder}`. These will be used from go/bundle to process the
  signatures section.
- Utility methods `Validate()` and `CertSha256()` are added.

### DIFF
--- a/go/signedexchange/certurl/certchain.go
+++ b/go/signedexchange/certurl/certchain.go
@@ -1,7 +1,10 @@
+// Package certurl implements a parser and a serializer for application/cert-chain+cbor format.
+// See https://wicg.github.io/webpackage/draft-yasskin-http-origin-signed-responses.html#cert-chain-format for the spec.
 package certurl
 
 import (
 	"bytes"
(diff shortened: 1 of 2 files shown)


go/signedexchange/certurl/certchain.go
```diff
@@ -12,13 +15,13 @@ import (
 	"github.com/WICG/webpackage/go/signedexchange/cbor"
 )
 
-type CertChainItem struct {
+type AugmentedCertificate struct {
```
nyaxt: Would you describe what `AugmentedCertificate` is as a comment?

nyaxt: (Would you add a specref?)

irori: Spec ref added.

PR #487, opened 2019-08-19
go: Refactor mice versioning code

irori:
Pure refactoring, no behavioral changes.

### DIFF
--- a/go/signedexchange/verifier.go
+++ b/go/signedexchange/verifier.go
@@ -14,7 +14,6 @@ import (
 
 	"github.com/WICG/webpackage/go/internal/signingalgorithm"
 	"github.com/WICG/webpackage/go/signedexchange/certurl"
-	"github.com/WICG/webpackage/go/signedexchange/mice"
 	"github.com/WICG/webpackage/go/signedexchange/structuredheader"
 	"github.com/WICG/webpackage/go/signedexchange/version"
 )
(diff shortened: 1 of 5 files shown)


go/signedexchange/verifier.go
```diff
@@ -353,18 +352,8 @@ func verifyTimestamps(sig *Signature, verificationTime time.Time) error {
...
 func verifyPayload(e *Exchange, signature *Signature) ([]byte, error) {
-	var integrityStr string
-	var enc mice.Encoding
-	switch e.Version {
-	case version.Version1b1:
-		enc = mice.Draft02Encoding
-		integrityStr = "mi-draft2"
-	case version.Version1b2, version.Version1b3:
-		enc = mice.Draft03Encoding
-		integrityStr = "digest/" + enc.ContentEncoding()
```
hajimehoshi: I was wondering if this change keeps this behavior.

irori: Any counter examples? `IntegrityIdentifier()` returns `"digest/mi-sha256-03"` for `Draft03Encoding`, which matches `"digest/" + enc.ContentEncoding()`.

hajimehoshi: Hm, I could not see the assumption from this PR.

irori: Well, the original code assumed that the digest algorithm and the HTTP content encoding share the same name. This holds for [http-mice-03](https://tools.ietf.org/html/draft-thomson-http-mice-03#section-6), but it's not necessarily true.

PR #490, opened 2019-08-20
go/bundle: Add sign-bundle tool

irori:
This adds `sign-bundle` command line tool. It takes an existing bundle, a certificate chain (generated by `gen-certurl`) and a private key, and creates a new bundle with a "signatures" section added.

The generated signatures section includes a signature that covers all exchanges whose hostname matches the certificate. Also, this tool encodes those exchanges with the mi-sha256-03 content encoding.

### DIFF
--- a/go/bundle/README.md
+++ b/go/bundle/README.md
@@ -2,9 +2,11 @@
 This directory contains a reference implementation of [Bundled HTTP Exchanges](https://wicg.github.io/webpackage/draft-yasskin-wpack-bundled-exchanges.html) spec.
 
 ## Overview
-We currently provide two command-line tools: `gen-bundle` and `dump-bundle`.
+We currently provide three command-line tools: `gen-bundle`, `sign-bundle` and `dump-bundle`.
 
 `gen-bundle` command is a bundle generator tool. `gen-bundle` consumes a set of http exchanges (currently in the form of [HAR format](https://w3c.github.io/web-performance/specs/HAR/Overview.html) or static files in a local directory), and emits a bundled exchange file.
+
+`sign-bundle` command attaches a signature to a bundle. `sign-bundle` takes an existing bundle file, a certificate and a private key, and emits a new bundle file with cryptographic signature for the bundled exchanges added.
 
 `dump-bundle` command is a bundle inspector tool. `dump-bundle` dumps the enclosed http exchanges of a given bundled exchange file in a human readable form.
 
@@ -22,7 +24,9 @@ We recommend using `go get` to install the command-line tool.
 go get -u github.com/WICG/webpackage/go/bundle/cmd/...
 ```
 
-### Usage
+## Usage
+
+### gen-bundle
 `gen-bundle` generates a bundled exchange file from a HAR file.
 
 One convenient way to generate HAR file is via Chrome Devtools. Navigate to "Network" panel, and right-click on any resource and select "Save as HAR with content".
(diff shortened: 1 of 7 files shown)


irori:
@jyasskin Could you review `signer.go` to see if the signing message is correct?

go/bundle/README.md
```diff
@@ -39,6 +43,21 @@ gen-bundle -dir static -baseURL https://www.example.com/ -o foo.wbn
 ```
 You can use `-startURL` command-line flag to specify the entry point of the bundle, as a relative URL from `-baseURL`. Currently, this just makes the exchange for `-startURL` the first entry in the bundled exchange file.
 
+### sign-bundle
+`sign-bundle` updates a bundle attaching a cryptographic signature of its exchanges. To use this tool, you need a pair of a private key and a certificate in the `application/cert-chain+cbor` format. See [go/signedexchange](../signedexchange/README.md) for more information on how to create a key and certificate pair.
+
+Assuming you have a key and certificate pair for `example.org`, thils command will sign all exchanges in `unsigned.wbn` whose URL's hostname is `example.org`, and writes a new bundle to `signed.wbn`.
```
kinu: nit: thils -> this

go/bundle/signature/signer.go
```diff
@@ -0,0 +1,193 @@
...
+
+type Signer struct {
+	Version version.Version
+	Certs   certurl.CertChain
+	PrivKey crypto.PrivateKey
+	Rand    io.Reader
+	SignedSubset
+}
+
+// SignedSubset represents the "sisnged-subset" structure.
```
kinu: a "signed-subset" structure

go/bundle/README.md
irori: fixed

go/bundle/signature/signer.go
irori: fixed

go/bundle/cmd/sign-bundle/main.go
```diff
@@ -0,0 +1,133 @@
...
+
+	newSignatures, err := signer.UpdateSignatures(b.Signatures)
+	if err != nil {
+		return err
+	}
+	b.Signatures = newSignatures
+	return nil
+}
+
+func main() {
```
nyaxt: Optional nit: we may want to copy main()/run() idiom: https://github.com/WICG/webpackage/blob/master/go/bundle/cmd/dump-bundle/main.go#L62

go/bundle/signature/signer.go
```diff
@@ -0,0 +1,193 @@
...
+
+// UpdateSignatures updates bundle.Signatures by adding the cert chain and
+// the signature of exchanges added with AddExchange.
+func (s *Signer) UpdateSignatures(signatures *bundle.Signatures) (*bundle.Signatures, error) {
+	if signatures == nil {
+		signatures = &bundle.Signatures{}
+	}
+
+	authorityIndex := len(signatures.Authorities)
+	signatures.Authorities = append(signatures.Authorities, s.Certs...)
```
jyasskin: I think this is ok, but it may lead to duplicate intermediate certificates in the authorities list.

go/bundle/cmd/sign-bundle/main.go
irori: Done.

go/bundle/signature/signer.go
irori: Added a TODO comment to deduplicate intermediate certificates.

irori:
Thanks for the reviews!

PR #493, opened 2019-08-26
go/bundle: Add signature verifier

irori:
After this patch, `dump-bundle` tool will verify exchanges based on the
content of the `signatures` section.

The `Dump` method in `bundle.go` is moved to `dump-bundle/main.go` to avoid
circular package dependency between `go/bundle` and `go/bundle/signature`.


go/bundle/cmd/dump-bundle/main.go
```diff
@@ -23,6 +26,61 @@ func ReadBundleFromFile(path string) (*bundle.Bundle, error) {
 	return bundle.Read(fi)
 }
 
+func DumpExchange(e *bundle.Exchange, b *bundle.Bundle, verifier *signature.Verifier) error {
```
hajimehoshi: Is it OK not to take io.Writer?

irori: I think stdout-only is fine for now, as user can redirect the output.

PR #495, opened 2019-09-03
gen-bundle: Add option to create a bundle from URL list

irori:
After this patch, `gen-bundle -URLList urls.txt` will fetch URLs listed
in `urls.txt` and create a bundle from their responses.

### DIFF
--- a/go/bundle/README.md
+++ b/go/bundle/README.md
@@ -4,7 +4,7 @@ This directory contains a reference implementation of [Bundled HTTP Exchanges](h
 ## Overview
 We currently provide three command-line tools: `gen-bundle`, `sign-bundle` and `dump-bundle`.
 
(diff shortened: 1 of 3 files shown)


go/bundle/README.md
```diff
@@ -37,11 +45,33 @@ Once you have the har file, generate the bundled exchange file via:
 gen-bundle -har foo.har -o foo.wbn
 ```
 
+#### From a URL list
+
+`gen-bundle` also accepts `-URLList FILE` flag. `FILE` is a plain text file with one URL on each line. `gen-bundle` fetches these URLs and put the responses into the bundle. For example, you could create `urls.txt` with:
+
+```
+# This is a comment.
```
kinu: nit:

# A line starting with '#' is a comment.

irori: done

PR #503, opened 2019-10-10
Add a pre-installation use case.

jyasskin:
I don't fully understand this use case, and why other solutions don't work well, so please suggest improvements.

@bkardell @riju

### DIFF
--- a/draft-yasskin-webpackage-use-cases.md
+++ b/draft-yasskin-webpackage-use-cases.md
@@ -384,16 +384,19 @@ internet connection. Thereafter, the application should use the normal Service
 Worker update mechanism to stay up to date.
 
 One way to accomplish this would be to pre-create a browser profile in the
-device's default browser, and navigate it to each of the pre-installed apps
-before recording the device image. However, this doesn't help users who change
-their default browser.
+device's default browser and navigate it to each of the pre-installed apps
+before recording the device image. However, this means end-users miss the
+browser's initial setup flow and possibly that any "unique" cookies the sites
+set are now shared across everyone who bought the device. It also doesn't help
+users who change their default browser.
 
 If multiple browsers supported an unsigned web package format, with an option to
-trust it as if it were signed if it's in a particular read-only section of the
-filesystem, and if registering a Service Worker from a page inside a package
-passed the full package contents to the Service Worker's `install` event, the
-device manufacturer could provide web packages for each pre-installed
-application that would work in the user's chosen browser.
+trust it as if it were signed if it's in a particular section of the filesystem
+that's as protected as the browser's executable, and if registering a Service
+Worker from a page inside a package passed the full package contents to the
+Service Worker's `install` event, the device manufacturer could provide web
+packages for each pre-installed application that would work in the user's chosen
+browser.
 
 Associated requirements:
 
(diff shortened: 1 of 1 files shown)


draft-yasskin-webpackage-use-cases.md
```diff
@@ -372,6 +376,30 @@ Associated requirements:
...
+
+Device manufacturers would like to ship their devices with some web applications
+pre-installed and usable even if the application is first used without an
+internet connection. Thereafter, the application should use the normal Service
+Worker update mechanism to stay up to date.
+
+One way to accomplish this would be to pre-create a browser profile in the
+device's default browser, and navigate it to each of the pre-installed apps
+before recording the device image. However, this doesn't help users who change
+their default browser.
```
bkardell: it's also a pretty weird build step, isn't it?   If you built a million devices each of them would do some browsing before shipping?  and if you have actual user profiles that will be established somehow along the way, I'm not sure how that would work either.

jyasskin: Yeah, maybe you can build images from the post-browsing filesystem, but now all your cookies are shared. And the browser's out-of-box experience is gone.

riju:
I basically meant use cases like a Digital Signage, similar to [Chrome's kiosk mode](https://developer.chrome.com/apps/manifest/kiosk_enabled), without PlayStore/WebStore and user's ability to install apps. The device can be a Point of Sale device or a Shopping mall / Library Catalog system, usually single-purpose, where the app is managed by some admin. 

Chrome Apps on CrOS in kiosk mode fills this gap, but if someone wants to run say, their downstream chromium on Linux and not use Chrome extensions but Web Standards, WebPackaging might help to pre-install their app in such a scenario.

kinu:
> I basically meant use cases like a Digital Signage, similar to [Chrome's kiosk mode](https://developer.chrome.com/apps/manifest/kiosk_enabled), without PlayStore/WebStore and user's ability to install apps. The device can be a Point of Sale device or a Shopping mall / Library Catalog system, usually single-purpose, where the app is managed by some admin.

If the device comes with web packaging in the way @jyasskin described in the change I think it'd work like that but are there something specifically more you wanted to see / evolve?

bkardell:
I'd like to + in @zdobersek here..

jyasskin:
@riju In the Digital Signage/Point of Sale/Shopping mall/Library Catalog case, would the device likely be completely disconnected from the internet? Or is it that it's easier to manage a single-app device by pushing the app to it, rather than hosting the app on a web server?

If these kiosk apps don't use the Service Worker mechanism to stay up to date, this may actually be a different use case than pre-installed apps.

I'm also happy if you want to just send me the change to the text that you think would improve the description.

bkardell:
@jyasskin I'm not entirely sure - It seems like connecting to the internet to stay up to date is optional in any case once you have the pwa? It's quite possible that you never update because the device works just fine and they have no interwebs... Is there a reason it requires an entirely different architecture or solution really?

jyasskin:
@bkardell I think the no-internet case is enabled by the yes-internet design, but because the package in that case is not really "on the web", I've had more trouble justifying why it should be an IETF goal. But maybe folks will be easier to convince than my mental model of them. :)

jyasskin:
I'm going to merge this, but @riju, feel free to suggest changes at any point.

PR #504, opened 2019-10-14
Frontend compromise use case

jyasskin:
I think this describes the easier half of #425, with the second use case mentioned in https://github.com/WICG/webpackage/pull/425#pullrequestreview-300951626 left for another PR.

@davidstrauss, does it make sense to use this instead of your PR?

Rendered version at https://jyasskin.github.io/webpackage/frontend-compromise-use-case/draft-yasskin-webpackage-use-cases.html#compromised-frontend.


davidstrauss:
> @davidstrauss, does it make sense to use this instead of your PR?

I believe it does, yes. Thank you for the write-up!

PR #506, opened 2019-10-21
Add JavaScript library for generating bundles

irori:
The `gen-bundle` Go tool is convenient for simple use cases, but developers who want to support WebBundle generation for their site may need more fine-grained control.

This adds a simple JavaScript library (npm package) for creating application/webbundle resources, which could be used as a building block of popular frontend development toolchain (such as plugins for gulp, webpack, etc.).

### DIFF
--- a/js/bundle/README.md
+++ b/js/bundle/README.md
@@ -1,0 +1,29 @@
+# js/bundle
+This directory contains a JavaScript library for serializing (and parsing, in the future) the `application/webbundle` format defined in the [Bundled HTTP Exchanges](https://wicg.github.io/webpackage/draft-yasskin-wpack-bundled-exchanges.html) draft spec.
+
+## Installation
+Using npm:
+```bash
+npm install webbundle
+```
+
+## Usage
+Please be aware that the API is not yet stable and is subject to change any time.
+
+Example:
+```javascript
+const webbundle = require('webbundle');
+const fs = require("fs");
+
+const primaryURL = 'https://example.com/';
+let builder = new webbundle.BundleBuilder(primaryURL);
+builder.setManifestURL('https://example.com/manifest.json');
+builder.addExchange(
+    primaryURL,                          // URL
+    200,                                 // response code
+    {'Content-Type': 'text/html'},       // response headers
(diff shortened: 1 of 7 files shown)


irori:

irori:
If you want to try this, download [webbundle-0.0.1.tgz](https://drive.google.com/file/d/1kWZQDei8paRk5fyiBRHszYlmP6IHExME/view?usp=sharing) and run `npm install webbundle-0.0.1.tgz`.

irori:
@jyasskin @kinu WDYT?

I think having JS-based tool would lower the hurdle for developers who want to integrate bundle generation step to their workflow.

js/bundle/src/encoder.ts
```diff
@@ -0,0 +1,141 @@
...
+        for (let [key, value] of this.entries()) {
+            m.set(byteString(key), byteString(value));
+        }
+        return encoder.pushAny(m);
+    }
+}
+
+// Throws an error if `urlString` is not a valid exchange URL.
+function validateExchangeURL(urlString: string): void {
+    let url = new URL(urlString);
```
kinu: This could throw if urlString is a relative URL, right?  Maybe write about that somewhere?

Should we also error out if protocol is not http/https?

js/bundle/README.md
```diff
@@ -0,0 +1,29 @@
+# js/bundle
+This directory contains a JavaScript library for serializing (and parsing, in the future) the `application/webbundle` format defined in the [Bundled HTTP Exchanges](https://wicg.github.io/webpackage/draft-yasskin-wpack-bundled-exchanges.html) draft spec.
```
kinu: Maybe also write out that currently it doesn't support origin-signed exchanges.

irori: Done.

js/bundle/src/encoder.ts
irori: Done.

irori:
It seems we can't publish `webbundle` package because the name is too similar to existing package `web-bundle` (which is not related to web packaing).

So I've renamed this package to `wicg-webbundle`.

PR #508, opened 2019-10-24
gen-bundle: Add -headerOverride flag

irori:
This can be used to inject arbitrary headers to bundled responses. For example, `Access-Control-Allow-Origin: *` may be added to workaround CORS errors in untrusted bundles.

### DIFF
--- a/go/bundle/cmd/gen-bundle/main.go
+++ b/go/bundle/cmd/gen-bundle/main.go
@@ -6,10 +6,22 @@ import (
 	"log"
 	"net/url"
 	"os"
+	"strings"
 
 	"github.com/WICG/webpackage/go/bundle"
 	"github.com/WICG/webpackage/go/bundle/version"
 )
+
+type headerArgs []string
+
+func (h *headerArgs) String() string {
+	return fmt.Sprintf("%v", *h)
+}
+
+func (h *headerArgs) Set(value string) error {
+	*h = append(*h, value)
+	return nil
+}
 
 var (
 	flagVersion     = flag.String("version", string(version.VersionB1), "The webbundle format version")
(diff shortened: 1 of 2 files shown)


irori:

marcoscaceres:
Marked as  substantive for IPR from ash-nazg.

marcoscaceres:

go/bundle/cmd/gen-bundle/main.go
```diff
@@ -6,11 +6,23 @@ import (
...
 
 	"github.com/WICG/webpackage/go/bundle"
 	"github.com/WICG/webpackage/go/bundle/version"
 )
 
+type headerArgs []string
+
+func (h *headerArgs) String() string {
+	return fmt.Sprintf("%v", *h)
+}
```
hajimehoshi: I'd write

```go
func (h headerArgs) String() string {
	return fmt.Sprintf("%v", h)
}
```

go/bundle/cmd/gen-bundle/main.go
```diff
@@ -6,11 +6,23 @@ import (
...
+
+func (h *headerArgs) Set(value string) error {
+	*h = append(*h, value)
+	return nil
+}
```
hajimehoshi: ditto

irori: We cannot use non-pointer `headerArgs` here because this function overwrites the pointed value.

go/bundle/cmd/gen-bundle/main.go
irori: Let me keep using `*headerArgs` here for consistency with `Set()`.

Also, `fmt.Sprintf("%v", h)` would recursively call `h.String()` so we'll need a type conversion to prevent that, which is a bit unfortunate.

go/bundle/cmd/gen-bundle/main.go
hajimehoshi: Oh you're right.

PR #515, opened 2019-11-01
js/bundle: Add tests

irori:
Also bumps package version to 0.0.2.


js/bundle/tests/decoder_test.js
```diff
@@ -0,0 +1,52 @@
+const wbn = require('../lib/wbn');
+
+describe('Bundle', () => {
+  const bundleBuffer = (() => {
```
horo-t: I think we should have a pre-generated test bundle file to detect unexpected regressions (or format change).

js/bundle/tests/decoder_test.js
```diff
@@ -0,0 +1,52 @@
...
+    const builder = new wbn.BundleBuilder('https://example.com/');
+    builder.setManifestURL('https://example.com/manifest.json');
```
horo-t: We should have no-manifest wbn test case.

js/bundle/tests/decoder_test.js
irori: Done.

js/bundle/tests/decoder_test.js
irori: The pre-generated wbn doesn't have a manifest.

PR #516, opened 2019-11-04
Replace bundle demo link to a newer one

kinu:
https://youtu.be/xAujz66la3Y this one uses the term "Web Bundles"
(Same scenario & materials with the previous one, with a little more editing)


kinu:
I probably should have merged this into the previous commit... just replacing a link.  (The video can be replaced/re-edited further though)

PR #519, opened 2019-11-06
TravisCI: Test nodejs module

irori:


irori:
@jyasskin Would you take a look?

.travis.yml
```diff
@@ -6,6 +6,16 @@ jobs:
         - GO111MODULE=on
       script: go test -v ./go/...
 
+    - name: "JavaScript"
+      language: node_js
+      node_js:
+       - "node"
+      install:
+       - npm --prefix ./js/bundle install
```
jyasskin, line 14: There is an [`npm ci`](https://docs.npmjs.com/cli/ci.html) command that would probably be better here.

PR #524, opened 2019-11-08
js/bundle: Support node 8.x

sisidovski:
Just a tiny fix.

I'm not sure this is intended or not, but global `URL` class was added from Node 10.x.

This change will support 8.x which is still at the maintenance LTS stage, actually it will be eol [soon](https://github.com/nodejs/Release) though. Even if doesn't support 8.x, at least supported node version should be specified via "engines" field in package.json.

@irori Could you have a look? Thanks!


irori:

PR #529, opened 2019-11-11
Explainer on the anti-tracking problem for signed packages

jyasskin:
@johnwilander @davidben @bslassey Here's a description of how we could prevent distributors from using packages to transfer user IDs between sites. What holes do you still see in it? Do you have any favorite options for the places I've listed several possible choices?

Thanks!

[Preview](https://github.com/WICG/webpackage/blob/anti-tracking/explainers/anti-tracking.md)

### DIFF
--- a/explainers/anti-tracking.md
+++ b/explainers/anti-tracking.md
@@ -1,0 +1,247 @@
+# Anti-tracking in signed packages
+
+<!-- TOC depthTo:3 -->
+
+- [Blocking credentials](#blocking-credentials)
+  - [Possible spellings](#possible-spellings)
+  - [Tentatively-rejected spellings](#tentatively-rejected-spellings)
+- [Blocking normal link decoration](#blocking-normal-link-decoration)
+- [Blocking privileged link decoration](#blocking-privileged-link-decoration)
+- [Privileged server-side collaboration](#privileged-server-side-collaboration)
+- [Appendices](#appendices)
+  - [Security concerns](#security-concerns)
+  - [Behavior on an unexpected response](#behavior-on-an-unexpected-response)
+
+<!-- /TOC -->
+
+This explainer attempts to solve the problems identified in issues
+[#422](https://github.com/WICG/webpackage/issues/422) and
+[#423](https://github.com/WICG/webpackage/issues/423). Specifically, we're
+worried that a distributor might be able to use the fact that it can choose
+which exchange or package, signed by a publisher, to serve in response to a
+client's request, to transfer its notion of the user's identity to the
+publisher.
+
+The approach this document takes is to prevent the distributor from accessing
+its notion of the user's identity. There are several routes by which a
+distributor could learn the user's identity:
+
+1. The browser could send the distributor its top-level credentials. [Countermeasures](#blocking-credentials)
+1. The source of a link could use the same link-decoration tools as it might use
+   in a cross-origin link, which other efforts are blocking. Note that the
+   source of a link and the distributor of a package could be the same origin.
+   [Countermeasures](#blocking-normal-link-decoration)
+1. The source of a link could decorate links to its distributor to send a user
+   ID in ways an unrelated cross-origin target couldn't be expected to receive.
+   [Countermeasures](#blocking-privileged-link-decoration)
+1. The source of a link could collaborate with its distributor on the server
+   side to send a user ID in ways an unrelated cross-origin target couldn't be
+   induced to collaborate.
+   [Discussion](#privileged-server-side-collaboration)
+
+These are tackled individually in the following sections.
+
+## Blocking credentials
+
+The navigation to a signed package (or exchange) must be done without
+credentials. This is roughly [credentials
+mode](https://fetch.spec.whatwg.org/#concept-request-credentials-mode) ==
+`"omit"`, with changes to Fetch to make it work for navigations.
+
+There are [security concerns](#security-concerns) for allowing an attacker to
+load an arbitrary site without credentials, so if the target of an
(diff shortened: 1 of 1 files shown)


explainers/anti-tracking.md
```diff
@@ -0,0 +1,247 @@
...
+   [Discussion](#privileged-server-side-collaboration)
+
+These are tackled individually in the following sections.
+
+## Blocking credentials
+
+The navigation to a signed package (or exchange) must be done without
+credentials. This is roughly [credentials
+mode](https://fetch.spec.whatwg.org/#concept-request-credentials-mode) ==
+`"omit"`, with changes to Fetch to make it work for navigations.
```
davidben: Ideally we'd be able to talk about which network partition this uses too, but I'm not sure if the spec framework for that exists yet. :-/

explainers/anti-tracking.md
```diff
@@ -0,0 +1,247 @@
...
+logs instead.
+
+## Appendices
+
+### Security concerns
+
+There are worries that uncredentialed navigation could be used in an attack on
+the target site, since the main request won't send or save cookies, but
+subrequests will. We don't actually know a concrete attack that's enabled by
+uncredentialed navigations, but we'd like to prevent surprises anyway.
```
davidben: This is a bit contrived, but here's an odd scenario.

publisher.example makes some content only available to paying subscribers. This website is written to the web of today and doesn't need fancy client-side machinery, so it uses server-side logic on the top-level HTML fetch by checking some first-party cookie.

 Some user subscribes and pays this publisher. The attacker now directs the user to some article but it's uncredentialed. Now even though the top-level URL says publisher.example and the user is a paying customer, the user is greeted with a request for money. The user now gets upset at the publisher for asking for even more money.

(Realistically the user will probably just thing something weird happened, click the "login" button, and things will probably settle quickly enough. But it is rather weird.)

explainers/anti-tracking.md
```diff
@@ -0,0 +1,247 @@
...
+
+Similar worries came up in the discussion of cross-origin prefetch, which led to
+the new `"prenavigate"` operation, for which the target site opts in using an
+[`Allow-Uncredentialed-Navigation`
+header](https://github.com/w3c/resource-hints/issues/82#issuecomment-529951528).
```
davidben: Something that might be worth mentioning: SXGs ensures the credentialed and uncredentialed versions of the top-level resources are the same, which motivates for the package-only options.

explainers/anti-tracking.md
```diff
@@ -0,0 +1,247 @@
...
+uncredentialed navigation isn't a signed package, the navigation should
+[fail](#behavior-on-an-unexpected-response).
+
+There are lots of possible spellings for a navigation that omits credentials:
+
+### Possible spellings
+
+<a id="credentials-omit"></a>
+
+#### An attribute to set [credentials-mode](https://fetch.spec.whatwg.org/#concept-request-credentials-mode) to "omit"
```
davidben: Something that might be worth writing out explicitly:

If I'm understanding things right, this option and the second option are qualitatively different from the remaining ones and prenavigate. These say that there are two ways a referrer may choose to enter a first-party site, credentialed or uncredentialed. The other options don't change the semantics of top-level navigations (aside from the SXG origin substitution capability itself), but are trying to slot an uncredentialed package fetch somehow.

explainers/anti-tracking.md
jyasskin: Which do we want it to use? I think it's the uncredentialed partition from the origin of the SXG, and the contents get to cross into the credentialed partition for the origin that signed them?

explainers/anti-tracking.md
jyasskin: I've written this.

explainers/anti-tracking.md
jyasskin: I've now described this as "surprising users". I agree that the user will probably just click "login" or "reload" in that case, although we'd need to run a user study to be confident of that, if folks think it's important.

explainers/anti-tracking.md
jyasskin: They don't guarantee they're the same. They basically allow the site to assert that the uncredentialed version is good enough, which is similar to the assertion with the `Allow-Uncredentialed-Navigation` header.

davidben: Ah yeah, that's a better phrasing.

explainers/anti-tracking.md
davidben: There isn't really a singular uncredentialed partition right now. We do partition credentialed and uncredentialed connections, but it was more a consequence of not confusing connection-based auth (client certs, NTLM, Channel ID at one point, etc) than an actual partition. In theory a smarter connection pool could pool across those without messing that up.

I don't think anyone is partitioning the HTTP cache by credentialed vs uncredentialed. (I'm *sure* sites always correctly labels their `Vary: Cookie` resources. :-) ) And pages can freely make uncredentialed subresource requests and either write data via URLs or read data via CORS-readable responses. I believe that means credentialed vs uncredentialed is useful for threat models that consider intentional state mechanisms (so CSRF and such), but that it's not useful for threat models which consider random caches being misused as state (which we seem to be converging on for anti-tracking).

https://github.com/kinu/speculative-loading talks about using a temporary partition, which would probably require a bit of spec machinery if that's what we settle on.

jyasskin: I've added some handwaving about figuring this out. I'll merge this now, but we can update it as we get the spec text worked out.

twifkak:
Addresses #422.

PR #538, opened 2019-11-23
Add "X-Content-Type-Options: nosniff" to example of firebase hosting config

sykmhmh:
fixed for
```
Signed exchange response without "X-Content-Type-Options: nosniff" header is not supported.
```


marcoscaceres:

PR #540, opened 2019-11-28
go/bundle: Support non-empty variants in decoder

irori:
This enables dump-bundle to read bundles that have multiple variants
for single URL.

Also this adds a roundtrip test (serialize and then deserialize) for a
bundle with variants.

### DIFF
--- a/go/bundle/bundle_test.go
+++ b/go/bundle/bundle_test.go
@@ -66,6 +66,40 @@ func createTestBundle(t *testing.T, ver version.Version) *Bundle {
 	return bundle
 }
 
+func createTestBundleWithVariants(ver version.Version) *Bundle {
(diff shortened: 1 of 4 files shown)


go/bundle/bundle_test.go
```diff
@@ -100,3 +134,24 @@ func TestWriteAndRead(t *testing.T) {
 		}
 	}
 }
+
+func TestWriteAndReadWithVariants(t *testing.T) {
+	for _, ver := range version.AllVersions {
+		if !ver.HasVariantsSupport() {
```
hajimehoshi: optional: `SupportsVariants`?

irori: Done.

Issue #347, opened 2018-12-07
Two new fields in SXG format to support subresource loading


horo-t:
I want to introduce two new fields in application/signed-exchange format.
 - Alternative-Signed-Exchange-Subresources map in unsigned filed.
 - Allowed-Alternative-Signed-Exchange-Subresources list in signed filed.
 
## Problem
Currently content publishers can sign their HTML contents using their own private keys. User Agents (UAs) can trust the signed contents as if the contents are served from the publisher’s origins even if they are served from other distributors’ origins. The signed contents can be served from any distributors’ origins. But if the publisher wants to serve subresources such as scripts and images from the distributors’ origin, the publisher needs to change the subresource URLs in the HTML to point to each distributors’ URL and need to sign for each distributor. The proposed two fields can solve this problem.

**Alternative-Signed-Exchange-Subresources map:**
A map from the original subresource requests to the SXG URLs. This field is not signed. So the distributor can change this field to point to their URLs.

**Allowed-Alternative-Signed-Exchange-Subresources list:**
The subresource URL list which can be served using SXG instead of fetching the original URL. This field is signed by the publisher. So the distributor can’t change this field.

## Example
**Publisher:** https://publisher.example/article_1.html
```
  <script src="framework.js"></script>
  <img src="article_1.jpg>
```

**SXG in Publisher:** https://publisher.example/article_1.html.sxg
```
[
  // URL
  'https://publisher.example/article_1.html',
  // Signature
  'sig1: sig=*...; integrity="digest/mi-sha256";cert-url="https://publisher.example/cert"',
  // [New field] Alternative-Signed-Exchange-Subresources
  // The key of the mapping may need Accept headers info in order to enable content
  // negotiation (e.g. for WebP).
  [
    [{':url': 'https://publisher.example/framework.js', 'accept': '*/*'},
     'https://publisher.example/framework.js.sxg'],
    [{':url': 'https://publisher.example/article_1.jpg', 'accept': '*/*'},
     'https://publisher.example/article_1.jpg.sxg']
  ],
  // Signed headers
  [
    { ':method': 'GET', 'accept': '*/*' },
    {
      ':status': '200',
      // [New field]
      ':allowed-alternative-signed-exchange-subresources':
          '"https://publisher.example/framework.js",'
          '"https://publisher.example/article_1.jpg"',
      'content-encoding': 'mi-sha256-03',
      'content-type': 'text/html; charset=utf-8',
      'digest': 'mi-sha256-03=....'
    },
[...]

PR #542, opened 2019-12-04
Add an explainer for subresource signed exchanges

horo-t:
I uploaded explainer documents of subresource signed exchanges to my
repository (https://github.com/horo-t/subresource-signed-exchange).
But they should be in this webpackage repository.
So this patch copies them from "horo-t/subresource-signed-exchange"
repository.

Spec issue: https://github.com/WICG/webpackage/issues/347
TAG review: https://github.com/w3ctag/design-reviews/issues/352


explainers/signed-exchange-alternate-link-explainer.md
```diff
@@ -0,0 +1,164 @@
+# Signed Exchange alternate link
+## Introduction
+We want to extend the usage of the existing [rel=alternate](https://html.spec.whatwg.org/multipage/links.html#rel-alternate) link header for signed exchange. Using this link header, the content publishers can declare that the resource is available in signed exchange format. This can be used both by the crawlers of aggregator sites  (SNS, News site, search engine..) and by the UAs. The crawlers can cache and serve the signed exchange of the content in their own server. The UAs can provide the users with a way to save the page in signe exchange format. And also signed exchange alternate links can be used to recursively prefetch appropriate subresource signed exchanges while prefetching the main resource signed exchange.
+
+## Use Cases
+### Signed Exchange discovery of main resource
```
jyasskin:
```suggestion
### Discovery of a signed exchange for the main resource
```

However, I'm not sure this use needs to be in an explainer at all. The "Otherwise" branch in https://html.spec.whatwg.org/multipage/links.html#rel-alternate already says that "The keyword creates a hyperlink referencing an alternate representation of the current document. The nature of the referenced document is given by the hreflang, and type attributes." There's nothing new in saying that a signed exchange can be an alternate representation of a document. So maybe delete this whole sub-section.

explainers/signed-exchange-alternate-link-explainer.md
```diff
@@ -0,0 +1,164 @@
...
+1. Content publishers can declare the URL of signed exchange format of the content using an alternate link HTTP header or using an alternate link HTML element.
+   - Example of signed exchange alternate link HTTP header in the HTTP response headers of the main resource of the content (https://publisher.example/article.html):
+      ```
+      Link: <https://sxg.publisher.example/article.html.sxg>;
+              rel="alternate";
+              type="application/signed-exchange;v=b3";
+              anchor="https://publisher.example/article.html"
```
jyasskin: Omit an `anchor` parameter if it's the same URL as the resource it appears in.
```suggestion
```

explainers/signed-exchange-alternate-link-explainer.md
```diff
@@ -0,0 +1,164 @@
+# Signed Exchange alternate link
+## Introduction
+We want to extend the usage of the existing [rel=alternate](https://html.spec.whatwg.org/multipage/links.html#rel-alternate) link header for signed exchange. Using this link header, the content publishers can declare that the resource is available in signed exchange format. This can be used both by the crawlers of aggregator sites  (SNS, News site, search engine..) and by the UAs. The crawlers can cache and serve the signed exchange of the content in their own server. The UAs can provide the users with a way to save the page in signe exchange format. And also signed exchange alternate links can be used to recursively prefetch appropriate subresource signed exchanges while prefetching the main resource signed exchange.
+
+## Use Cases
+### Signed Exchange discovery of main resource
+1. Content publishers can declare the URL of signed exchange format of the content using an alternate link HTTP header or using an alternate link HTML element.
```
jyasskin:
```suggestion
1. A publisher can offer a signed exchange version of a resource using either a `Link` header on that resource of a `<link rel="alternate">` element within the resource.
```

jyasskin: I don't think the paragraphs in this section are really an ordered list.

```suggestion
Content publishers can declare the URL of signed exchange format of the content using an alternate link HTTP header or using an alternate link HTML element.
```

explainers/signed-exchange-alternate-link-explainer.md
```diff
@@ -0,0 +1,164 @@
...
+              rel="alternate";
+              type="application/signed-exchange;v=b3";
+              anchor="https://publisher.example/article.html"
+      ```
+   - Example of signed exchange alternate link HTML element in the main resource of the content (https://publisher.example/article.html):
+      ```
+      <link href="https://sxg.publisher.example/article.html.sxg"
+            rel="alternate"
+            type="application/signed-exchange;v=b3"
+            anchor="https://publisher.example/article.html">
```
jyasskin: Also omit this one.
```suggestion
            >
```

explainers/signed-exchange-alternate-link-explainer.md
```diff
@@ -0,0 +1,164 @@
...
+   - When the crawlers detects the signed exchange alternate link, the crawlers can fetch the signed exchange.  And when a user is browsing the aggregator site (https://feed.example) and it has a link to the article, the signed exchange can be used for [privacy-preserving prefetching](https://wicg.github.io/webpackage/draft-yasskin-webpackage-use-cases.html#private-prefetch) using a prefetch `<link>` element and `<a>` link to the signed exchange .
+      ```
+      <link href="https://feed.example/article.html.sxg"
+            rel="prefetch">
+      <a href="https://feed.example/article.html.sxg">feed.example</a>
+      ```
+   - While the user of the UA is browsing an article (article.html), and if there is a signed exchange alternate link, the UA can provide the user with a way to save the page in signe exchange format. The saved file can be used to share with other users.
+
+### Recursive subresource signed exchange prefetch
+1. A crawler of an aggregator site fetches an article (https://publisher.example/article.html).
```
jyasskin: I don't think crawlers are likely to respect subresource signed exchanges. They can just go to the original host, which can offer the signed exchange as a same-document alternative. So remove from here down to "The aggregator site will serve"?

explainers/signed-exchange-alternate-link-explainer.md
```diff
@@ -0,0 +1,164 @@
...
+              rel="preload";
+              as="script"
+      Link: <https://cdn.publisher.example/lib.js>;
+              rel="allowed-alt-sxg";
+              header-integrity="sha256-XXXXXX"
+      ```
+1. The crawler fetches and verifies the subresource signed exchange (lib.js.sxg).
+1. The aggregator site will serve the signed exchanges (article.html.sxg and lib.js.sxg) from their own server.
+
+1. While a user is browsing the aggregator site (https://feed.example), a prefetch link element is inserted because the use is likely to want to read the article.
```
jyasskin:
```suggestion
While a user is browsing an aggregator site (https://feed.example), the aggregator guesses that the user is likely to want to read a particular article (https://publisher.example/article.html) and so inserts a prefetch link pointing to a signed exchange version of that article.
```

explainers/signed-exchange-alternate-link-explainer.md
```diff
@@ -0,0 +1,164 @@
...
+   ```
+   <link rel="prefetch" href="https://feed.example/sxg.publisher.example/article.html.sxg">
+   ```
+1. The UA prefetches the signed exchange (article.html.sxg).
```
jyasskin:
```suggestion
When the UA prefetches the signed exchange (article.html.sxg), the aggregator server includes a declaration that one of `article.html`'s subresources (https://cdn.publisher.example/lib.js) is also available from the same aggregator.
```

explainers/signed-exchange-alternate-link-explainer.md
```diff
@@ -0,0 +1,164 @@
...
+   1. The response from the server has an alternate link of subresource signed exchange:
+      ```
+      Link: <https://feed.example/sxg.publisher.example/lib.js.sxg>;
+              rel="alternate";
+              type="application/signed-exchange;v=b3";
+              anchor="https://cdn.publisher.example/lib.js"
+      ```
+   1. And the inner response of the main resource signed exchange (article.html.sxg) has a preload header and an allowed-alt-sxg header:
```
jyasskin:
```suggestion
To prevent an attacker from loading an incompatible version of the subresource, the resource _inside_ the signed exchange has to identify the exact version of the replacement signed exchange using a `Link: ... rel="allowed-alt-sxg"` with the hash of the signed headers (which themselves include a hash of the content).

To prevent a tracker from conveying a user ID in their choice of which subresources to prefetch, the inner resource has to preload the same subresources that the aggregator prefetches.
```

explainers/signed-exchange-alternate-link-explainer.md
```diff
@@ -0,0 +1,164 @@
...
+      ```
+      Link: <https://cdn.publisher.example/lib.js>;
+              rel="preload";
+              as="script"
+      Link: <https://cdn.publisher.example/lib.js>;
+              rel="allowed-alt-sxg";
+              header-integrity="sha256-XXXXXX"
+      ```
+1. The UA recursively prefetches lib.js.sxg.
+1. If the user clicks a link the article’s signed exchange, both the main resource of the article and the script are loaded from the prefetched signed exchanges.
```
jyasskin:
```suggestion
If the user navigates to the expected article, both the main resource of the article and the script subresource are loaded from the prefetched signed exchanges.
```

explainers/signed-exchange-alternate-link-explainer.md
```diff
@@ -0,0 +1,164 @@
...
+          rel="allowed-alt-sxg";
+          variants-05="accept;image/jpeg;image/webp";
+          variant-key-05="image/webp";
+          header-integrity="sha256-BBB"
+    Link: <https://publisher.example/image>; rel=preload; as=image;
+    ```
+If a UA supports WebP, the UA should prefetch **image_webp.sxg** which content is WebP format. Otherwise the UA should prefetch **image_jpeg.sxg** which content is JPEG format.
+
+## Security and Privacy Considerations
+The UA must fetch the alternate signed exchange subresource (lib.js.sxg) even if there is the original subresource (lib.js) in the HTTPCache. Otherwise it leaks the state of publisher’s site to the distributor of the signed exchange.
```
jyasskin: Also, to prevent tracking (user ID transfer), if the aggregator failed to prefetch a subresource that the main resource preloads, the UA must drop all of the subresource prefetches. If the aggregator prefetches a superset of the preloaded subresources, the UA must drop the ones that weren't preloaded.

explainers/signed-exchange-alternate-link-explainer.md
```diff
@@ -0,0 +1,164 @@
...
+
+## Proposal
+While processing preload link HTTP headers in prefetched main resource signed exchange’s inner response:
+1. Check whether matching “allowed-alt-sxg” link HTTP header in the inner response exists or not. (Note that if the allowed-alt-sxg link HTTP header has variants and variant-key attributes, the UA must execute the algorithm written in [HTTP Representation Variants](https://httpwg.org/http-extensions/draft-ietf-httpbis-variants.html) spec to find the matching header.)
+1. If exists, check whether matching “alternate” link HTTP header in the outer response exists or not.
+1. If exists, prefetches the matching signed exchange instead of prefetching the original resource URL.
+1. The prefetched signed exchange will be stored to the SignedExchangeCache of the Document. And it will be passed to the next Document and used while processing the preload link header. This behavior is written in [Signed Exchange subresource substitution explainer](./signed-exchange-subresource-subtitution-explainer.md).
+
+## Detailed design discussion
```
jyasskin: Probably just remove this sub-header.

explainers/signed-exchange-alternate-link-explainer.md
```diff
@@ -0,0 +1,164 @@
...
+              as="script"
+      Link: <https://cdn.publisher.example/lib.js>;
+              rel="allowed-alt-sxg";
+              header-integrity="sha256-XXXXXX"
+      ```
+1. The UA recursively prefetches lib.js.sxg.
+1. If the user clicks a link the article’s signed exchange, both the main resource of the article and the script are loaded from the prefetched signed exchanges.
+
+## Proposal
+While processing preload link HTTP headers in prefetched main resource signed exchange’s inner response:
```
jyasskin:
```suggestion
While prefetching an HTML resource:
```

Then write an explicit loop that checks if every preload has a matching and allowed prefetch before deciding to use all of the prefetches.

explainers/signed-exchange-alternate-link-explainer.md
horo-t: Done

explainers/signed-exchange-alternate-link-explainer.md
horo-t: Done

explainers/signed-exchange-alternate-link-explainer.md
```diff
@@ -0,0 +1,164 @@
+# Signed Exchange alternate link
+## Introduction
+We want to extend the usage of the existing [rel=alternate](https://html.spec.whatwg.org/multipage/links.html#rel-alternate) link header for signed exchange. Using this link header, the content publishers can declare that the resource is available in signed exchange format. This can be used both by the crawlers of aggregator sites  (SNS, News site, search engine..) and by the UAs. The crawlers can cache and serve the signed exchange of the content in their own server. The UAs can provide the users with a way to save the page in signe exchange format. And also signed exchange alternate links can be used to recursively prefetch appropriate subresource signed exchanges while prefetching the main resource signed exchange.
```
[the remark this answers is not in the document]
horo-t: done

explainers/signed-exchange-subresource-subtitution-explainer.md
```diff
@@ -0,0 +1,315 @@
...
+Signed Exchange has solved this
+[privacy-preserving prefetching](https://wicg.github.io/webpackage/draft-yasskin-webpackage-use-cases.html#private-prefetch)
+problem for main resources. If the publisher is providing the article in signed
+exchange format (article.html.sxg), the UA can prefetch the signed exchange from
+the feed site’s own server while the user is browsing the feed site. But there
+is no way to prefetch subresources in a privacy-preserving manner yet.
+
+Our proposal can solve this problem:
+1. The publisher provides the script file in signed exchange format (lib.js.sxg)
+along with its [header integrity value](#header-integrity-of-signed-exchange).
```
jyasskin: The publisher doesn't have to independently provide the header integrity value, right? That's computable from the SXG itself?

explainers/signed-exchange-subresource-subtitution-explainer.md
```diff
@@ -0,0 +1,315 @@
...
+1. The UA prefetches the signed exchange of the script from the feed site’s own
+server while the user is browsing the feed site.
+1. The HTTP response of the article from the publisher's server (https://publisher.example/article.html) has an allowed-alt-sxg link header to declare
```
jyasskin:
```suggestion
1. The signed HTTP response inside the signed exchange (https://publisher.example/article.html) has an `allowed-alt-sxg` link header to declare
```

explainers/signed-exchange-subresource-subtitution-explainer.md
```diff
@@ -0,0 +1,315 @@
...
+that the UA can load the script from the prefetched signed exchange.
```
jyasskin:
```suggestion
exactly what version of the subresource is compatible with this main resource.
```

explainers/signed-exchange-subresource-subtitution-explainer.md
```diff
@@ -0,0 +1,315 @@
...
+   ```
+   Link: <https://cdn.publisher.example/lib.js>;
+         rel="allowed-alt-sxg";
+         header-integrity="sha256-..."
+   ```
+   So, the UA can load the JS file from the prefetched signed exchange. If the
```
jyasskin: I think this paragraph can go away?

explainers/signed-exchange-subresource-subtitution-explainer.md
```diff
@@ -0,0 +1,315 @@
...
+   publisher is providing the article in signed exchange format
+   (article.html.sxg), the allowed-alt-sxg link header is in the inner response
+   headers of the signed exchange.
+
+### Reading article while offline
```
jyasskin: This is probably better solved by bundles.

explainers/signed-exchange-subresource-subtitution-explainer.md
```diff
@@ -0,0 +1,315 @@
...
+elements to prefetch the signed exchanges. The service worker responds to the
+fetch events of the prefetch with the signed exchanges in the CacheStorage.
+1. When the user clicks a link to the article’s signed exchange, the content and
+all necessary subresources are loaded from the cached signed exchange. So the
+user can read the article while offline.
+
+
+# Proposal
+
+1. Introduce SignedExchangeCache and PrefetchedSignedExchangeCache which are
```
jyasskin: These seem like Chromium-internal concepts rather than things the TAG needs to think about.

explainers/signed-exchange-subresource-subtitution-explainer.md
```diff
@@ -0,0 +1,315 @@
...
+   check of the body), stores the parsedExchange which is the result of
+   [parsing a signed exchange](https://wicg.github.io/webpackage/loading.html#ref-for-parsing-a-signed-exchange)
+   (= inner request URL and inner response) to the SignedExchangeCache.
+1. While
+[navigating across documents](https://html.spec.whatwg.org/multipage/browsing-the-web.html#navigating-across-documents),
+copy the parsedExchanges in the SignedExchangeCache of the source document to
+the PrefetchedSignedExchangeCache of the target document. This is intended to
+provide a way to pass the cached signed exchange across origins even if the UA
+is using origin isolated HTTPCache mechanism.
+1. While processing
```
jyasskin: This seems very similar to other explainer's algorithm. Maybe add anything that's missing there and remove this one?

explainers/signed-exchange-subresource-subtitution-explainer.md
```diff
@@ -0,0 +1,315 @@
...
+exchange’s response header must have digest header.
+
+The UA needs to check that this value of the prefetched subresource signed
+exchange is same as the header-integrity attribute of allowed-alt-sxg link
+header in the response from the publisher. This is intended to prevent the
+distributor from encoding a tracking ID into the subresource signed exchange.
+
+We can’t use the
+[SRI’s integrity](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity)
+for this purpose, because SRI’s integrity can be used only for verifying the
```
jyasskin:
```suggestion
for this purpose, because SRI’s integrity only covers the content body and not any of the headers.
```

explainers/signed-exchange-subresource-subtitution-explainer.md
```diff
@@ -0,0 +1,315 @@
...
+integrity of the content body. So if the UA use the SRI’s integrity value in
+allowed-alt-sxg link header, we can use the subresource signed exchanges to
+track the users by changing content-type and detecting the image loading
+failure.
+
+## Multiple subresource signed exchanges
```
jyasskin: This section is covered in the other explainer.

explainers/signed-exchange-subresource-subtitution-explainer.md
```diff
@@ -0,0 +1,315 @@
...
+   header have matching (url and header-integrity) parsedExchange in
+   PrefetchedSignedExchangeCache. If the check passes, set request's stashed
+   exchange to the parsedExchange so the resource will be loaded from the cached
+   signed exchange.
+   - After processing the link HTTP headers, clears the
+   PrefetchedSignedExchangeCache.
+
+# Detailed design discussion
+## Header integrity of signed exchange
+We use the SHA256 hash value of *signedHeaders* in
```
jyasskin:
```suggestion
The `header-integrity` parameter of the `allowed-alt-sxg` link is the SHA256 hash value of the *signedHeaders* value from the
```

explainers/signed-exchange-subresource-subtitution-explainer.md
```diff
@@ -0,0 +1,315 @@
...
+[application/signed-exchange format](https://wicg.github.io/webpackage/draft-yasskin-http-origin-signed-responses.html#application-signed-exchange)
+for integrity checking. This signedHeaders is *"the canonical serialization of
+the CBOR representation of the response headers of the exchange represented by
+the application/signed-exchange resource, excluding the Signature header
+field"*. So this value doesn’t change even if the publisher signs the content
+again or changes the signing key. This header-integrity value can verify the
```
jyasskin:
```suggestion
again or changes the signing key. This header-integrity value also guarantees the content body hasn't changed, because the signed headers are required to include a `Digest` header with a hash of the content body.
```

explainers/signed-exchange-subresource-subtitution-explainer.md
```diff
@@ -0,0 +1,315 @@
...
+      header-integrity="sha256-YYY"
+```
+
+## Lifetime of the entry in SignedExchangeCache
+The UA must check both the
+[signature expire time](https://wicg.github.io/webpackage/loading.html#exchange-signature-expiration-time)
+of the signed exchange and Cache-Control header of the outer response. The UA
+may discard the entry if it is expired.
+
+## Content negotiation using Variants and Variant-Key
```
jyasskin: Only include this section in one of the explainers.

explainers/signed-exchange-subresource-subtitution-explainer.md
```diff
@@ -0,0 +1,315 @@
...
+allowed-alt-sxg link header, we can use the subresource signed exchanges to
+track the users by changing content-type and detecting the image loading
+failure.
+
+## Multiple subresource signed exchanges
+If there are multiple matching subresource signed exchanges (example:
+script.js.sxg and image.jpg.sxg), the UA must check that there is no error in
+the all signed exchanges (eg: sig matching, URL matching, Merkle Integrity
+error) before processing the content of the signed exchanges. This is intended
+to prevent the distributor from encoding a tracking ID into the set of
```
jyasskin: And to prevent the distributor from selecting a version of the subresource that isn't compatible with the selected version of the main resource, which might introduce a security hole.

explainers/signed-exchange-subresource-subtitution-explainer.md
```diff
@@ -0,0 +1,315 @@
+# Signed Exchange subresource substitution
+
+## Introduction
+
+We want to introduce a new **rel=allowed-alt-sxg** link HTTP header which will
```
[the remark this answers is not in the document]
horo-t: Done

explainers/signed-exchange-subresource-subtitution-explainer.md
```diff
@@ -0,0 +1,315 @@
+# Signed Exchange subresource substitution
```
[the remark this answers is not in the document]
horo-t: Merged.

explainers/signed-exchange-subresource-subtitution-explainer.md
horo-t: Merged

explainers/signed-exchange-subresource-subtitution-explainer.md
horo-t: Merged

explainers/signed-exchange-subresource-subtitution-explainer.md
horo-t: Merged.

explainers/signed-exchange-subresource-subtitution-explainer.md
```diff
@@ -0,0 +1,455 @@
...
+"[prefetch](https://html.spec.whatwg.org/multipage/links.html#link-type-prefetch)"
+link (eg: `<link rel="prefetch"
+href="https://cdn.feed.example/cdn.publisher.example/lib.js.sxg" as="script">`):
+   - If succeeded prefetching the signed exchange (including merkle integrity
+   check of the body), stores the parsedExchange which is the result of
+   [parsing a signed exchange](https://wicg.github.io/webpackage/loading.html#ref-for-parsing-a-signed-exchange)
+   (= inner request URL and inner response) to the SignedExchangeCache.
+1. While
+[navigating across documents](https://html.spec.whatwg.org/multipage/browsing-the-web.html#navigating-across-documents),
+copy the parsedExchanges in the SignedExchangeCache of the source document to
```
jyasskin: Presumably except for the one that serves the navigation itself?

explainers/signed-exchange-subresource-subtitution-explainer.md
```diff
@@ -0,0 +1,455 @@
...
+the PrefetchedSignedExchangeCache of the target document. This is intended to
+provide a way to pass the cached signed exchange across origins even if the UA
+is using origin isolated HTTPCache mechanism.
+1. While processing
```
jyasskin: This step should set up what it's trying to do. I think that's something like

> The navigated-to document has a set of preloads for which it uses the `allowed-alt-sxg` link relation to declare that they can be served by signed exchanges. The UA either serves all of them from SXGs prefetched by the previous page, or none of them.

explainers/signed-exchange-subresource-subtitution-explainer.md
```diff
@@ -0,0 +1,455 @@
...
+   - Check whether matching "allowed-alt-sxg" link HTTP header
+   (`Link: <https://cdn.publisher.example/lib.js>;rel="allowed-alt-sxg";header-integrity="sha256-.."`)
+   exists or not. (Note that if the allowed-alt-sxg link HTTP header has
+   variants and variant-key attributes, the UA must execute the algorithm
+   written in
+   [HTTP Representation Variants spec](https://httpwg.org/http-extensions/draft-ietf-httpbis-variants.html)
+   to find the matching header.)
+   - Check whether all preload links which have matching allowed-alt-sxg link
+   header have matching (url and header-integrity) parsedExchange in
+   PrefetchedSignedExchangeCache. If the check passes, set request's stashed
```
jyasskin: "request" hasn't been introduced yet.

explainers/signed-exchange-subresource-subtitution-explainer.md
```diff
@@ -0,0 +1,455 @@
...
+   exchange to the parsedExchange so the resource will be loaded from the cached
+   signed exchange.
+   - After processing the link HTTP headers, clears the
+   PrefetchedSignedExchangeCache.
+
+## Recursive subresource signed exchange prefetch
```
jyasskin: This also looks like the same scenario as the previous section.

explainers/signed-exchange-subresource-subtitution-explainer.md
```diff
@@ -0,0 +1,455 @@
...
+subresource, prefetch the subresource signed exchange instead of prefetching the
+original resource URL.
+1. The prefetched signed exchange will be stored to the SignedExchangeCache of
+the Document. And it will be passed to the next Document and used while
+processing the preload link header. This behavior is written in above.
+
+
+# Detailed design discussion
+## Header integrity of signed exchange
+The `header-integrity` parameter of the `allowed-alt-sxg` link is the SHA256
```
jyasskin: There should probably be a section that starts with a definition of the `allowed-alt-sxg` link relation, including an example header expressing it.

explainers/signed-exchange-subresource-subtitution-explainer.md
```diff
@@ -0,0 +1,455 @@
...
+allowed-alt-sxg link header, we can use the subresource signed exchanges to
+track the users by changing content-type and detecting the image loading
+failure.
+
+## Multiple subresource signed exchanges
+If there are multiple matching subresource signed exchanges (example:
+script.js.sxg and image.jpg.sxg), the UA must check that there is no error in
+the all signed exchanges (eg: sig matching, URL matching, Merkle Integrity
+error) before processing the content of the signed exchanges. This means that
+the UA can use the subresource signed exchanges only when they are defined in
```
jyasskin: I don't follow "defined in the header".

explainers/signed-exchange-subresource-subtitution-explainer.md
```diff
@@ -0,0 +1,455 @@
...
+for this purpose, because SRI’s integrity only covers the content body and not
+any of the headers. So if the UA use the SRI’s integrity value in
+allowed-alt-sxg link header, we can use the subresource signed exchanges to
+track the users by changing content-type and detecting the image loading
+failure.
+
+## Multiple subresource signed exchanges
+If there are multiple matching subresource signed exchanges (example:
+script.js.sxg and image.jpg.sxg), the UA must check that there is no error in
+the all signed exchanges (eg: sig matching, URL matching, Merkle Integrity
```
jyasskin: Maybe

```suggestion
any of the signed exchanges (eg: sig matching, URL matching, Merkle Integrity
```

?

explainers/signed-exchange-subresource-subtitution-explainer.md
```diff
@@ -0,0 +1,455 @@
...
+We can’t use the
+[SRI’s integrity](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity)
+for this purpose, because SRI’s integrity only covers the content body and not
+any of the headers. So if the UA use the SRI’s integrity value in
+allowed-alt-sxg link header, we can use the subresource signed exchanges to
+track the users by changing content-type and detecting the image loading
+failure.
+
+## Multiple subresource signed exchanges
+If there are multiple matching subresource signed exchanges (example:
```
jyasskin: It's not clear what "matching" means here. I think you're referring to subresource preloads that are provided by prefetches on the previous page?

explainers/signed-exchange-subresource-subtitution-explainer.md
```diff
@@ -0,0 +1,455 @@
...
+and [Deterministic History Leaks](https://hillbrad.github.io/sri-addressable-caching/sri-addressable-caching.html#history_leaks)
+described in [the document about Subresource Integrity Addressable Caching](https://hillbrad.github.io/sri-addressable-caching/sri-addressable-caching.html).
+
+To avoid these privacy issues, we introduced the limitation that we can only use
+the signed exchanges which were prefetched by the referrer page.
+
+## Can’t we merge allowed-alt-sxg to preload header?
+If we can declare the header-integrity value in the existing preload link HTTP
+header, we don’t need to introduce a new "allowed-alt-sxg" link HTTP header. But
+it becomes complicated when supporting the
```
jyasskin: This might be easier to read as something like, "However, the imagesrcset attribute allows a single preload link to declare multiple different target URLs, and it's difficult to embed a `header-integrity` value for each of those URLs into the existing syntax. Instead, we use a separate link that gives a hash of the expected content for each of the possible URLs, while the preload tag continues to select which of the URLs is actually used."

(I do want to double-check that you can defend that claim that it's difficult to embed. Why couldn't it just come after the XXXw field?)

explainers/signed-exchange-subresource-subtitution-explainer.md
```diff
@@ -0,0 +1,455 @@
...
+          variant-key-05="image/jpeg";
+          header-integrity="sha256-AAA"
+    Link: <https://publisher.example/image>;
+          rel="allowed-alt-sxg";
+          variants-05="accept;image/jpeg;image/webp";
+          variant-key-05="image/webp";
+          header-integrity="sha256-BBB"
+    Link: <https://publisher.example/image>; rel=preload; as=image;
+    ```
+If a UA supports WebP, the UA should prefetch **image_webp.sxg** which content
```
jyasskin:
```suggestion
If a UA supports WebP, the UA must prefetch **image_webp.sxg** holding a WebP image.
```

This isn't "should", since it's defined by the variants algorithm.

explainers/signed-exchange-subresource-subtitution-explainer.md
```diff
@@ -0,0 +1,455 @@
...
+is WebP format. Otherwise the UA should prefetch **image_jpeg.sxg** which
+content is JPEG format.
+
+## Security and Privacy Considerations
+- The publishers can know whether the referrer page has prefetched the signed
+exchange subresources or not by checking the resource timing information. But
+this only exposes 1 bit information (= yes or no) because UAs can use the cached
+signed exchange only if the required signed exchanges are all available.
+- The UA can use the prefetched signed exchange subresources, only when they
+were prefetched in the referrer page. This is intended to avoid leaking the
```
jyasskin:
```suggestion
were prefetched in the immediate referrer page. This is intended to avoid leaking the
```

It's not clear to me exactly what information gets leaked if we do allow a series of navigations, especially same-origin navigations, to continue using the same set of prefetches. Maye the benefit isn't there, since the first page can just re-prefetch the same-origin content, but then we should say that instead of being vague about the leak.

explainers/signed-exchange-subresource-subtitution-explainer.md
```diff
@@ -0,0 +1,455 @@
...
+prefetching state to succeeding pages.
+- The UA check the header integrity value, so the distributor of the subresource
+signed exchange can’t inject arbitrary resources to the publisher’s page. This
+prevents distributors from sending tracking IDs to the publisher’s page.
+- The UA need to check both
+[the request URL](https://wicg.github.io/webpackage/loading.html#exchange-request-url)
+and the header integrity value of the signed exchange to avoid
```
jyasskin: "The UA needs to check that the signed request URL matches the preload link and not just that the header-integrity value matches, since the header-integrity hash doesn't cover the request URL."

explainers/signed-exchange-subresource-subtitution-explainer.md
```diff
@@ -0,0 +1,455 @@
...
+[Origin Laundering](https://hillbrad.github.io/sri-addressable-caching/sri-addressable-caching.html#origin_laundering)
+attack. The URL is not in signedHeaders, so the header integrity value can’t
+check the URL is correct.
+- The UA must fetch the signed exchange subresource
+(https://cdn.feed.example/cdn.publisher.example/lib.js.sxg) while processing the
+prefetch link element  (`<link rel="prefetch"
+href="https://cdn.feed.example/cdn.publisher.example/lib.js.sxg" as="script">`)
+even if there is the original subresource (http://cdn.publisher.example/lib.js)
```
jyasskin:
```suggestion
even if the HTTP cache already contains the original subresource (http://cdn.publisher.example/lib.js)
```

Although it's not clear to me in this system how the UA would even know what the original subresource URL is before fetching the SXG.

explainers/signed-exchange-subresource-subtitution-explainer.md
```diff
@@ -0,0 +1,455 @@
...
+in the HTTPCache. Otherwise it leaks the state of  publisher’s site to the
+distributor of the signed exchange.
+- The UA must fetch the alternate signed exchange subresource (lib.js.sxg) even
```
jyasskin: This is a duplicate of the previous bullet point.

explainers/signed-exchange-subresource-subtitution-explainer.md
```diff
@@ -0,0 +1,455 @@
...
+if there is the original subresource (lib.js) in the HTTPCache. Otherwise it
+leaks the state of publisher’s site to the distributor of the signed exchange.
+- To prevent tracking (user ID transfer), if the aggregator failed to prefetch a
```
jyasskin: This is very similar to the first bullet in this list. Put them next to each other.

explainers/signed-exchange-subresource-subtitution-explainer.md
```diff
@@ -0,0 +1,455 @@
...
+subresource that the main resource preloads, the UA must drop all of the
+subresource prefetches. If the aggregator prefetches a superset of the preloaded
+subresources, the UA must drop the ones that weren't preloaded.
+- The UA must ignore the not-yet-completed subresource signed exchange prefetch
```
jyasskin: This one should probably start, "If a replaced subresource prefetch hasn't completed by the time the UA would start fetching it in the course of loading the next page, the UA must cancel that prefetch and fetch the resource from its original URL. This prevents the distributor from ..."

explainers/signed-exchange-subresource-subtitution-explainer.md
```diff
@@ -0,0 +1,455 @@
...
+to prevent distributors from interfering the publisher’s page. (Eg.
+Intentionally block or delay the subresource loading.) So if the user clicks the
+link to the article while the UA is still prefetching the subresource signed
+exchange, the UA must fetch the subresource from the original URL after
+navigation.
+- The UA can use the prefetched signed exchange subresources only if the preload
+HTTP link headers for the subresources exist in the response headers. It is
+because the UA need to check the availability of all required signed exchanges
```
jyasskin: Isn't this the same as the "To prevent tracking (user ID transfer)" item?

explainers/signed-exchange-subresource-subtitution-explainer.md
```diff
@@ -0,0 +1,455 @@
...
+before start loading subresources.
+
+## [Self-Review Questionnaire: Security and Privacy](https://www.w3.org/TR/security-privacy-questionnaire/)
+1. What information might this feature expose to Web sites or other parties, and
+for what purposes is that exposure necessary?
+   - The existence of the alternative signed exchange in HTTP Cache is exposed.
```
jyasskin: Is this true? Who is it exposed to and under what circumstances? When is the alternative SXG even in the HTTP cache? Which partition?

explainers/signed-exchange-subresource-subtitution-explainer.md
```diff
@@ -0,0 +1,455 @@
...
+1. Is this specification exposing the minimum amount of information necessary to
+power the feature?
+   - Yes. This proposal has limitations such as "all subresource SXG must be
```
jyasskin: Say what the limitations accomplish here, not just that some phrases exist in other parts of this document.

explainers/signed-exchange-subresource-subtitution-explainer.md
```diff
@@ -0,0 +1,455 @@
...
+   finished prefetching, otherwise ignored", "subresource SXG must be prefetched
+   even if the original subresource exists in HTTPCache".
+1. How does this specification deal with personal information or
+personally-identifiable information or information derived thereof?
+   - Signed Exchange should not include personal information.
```
jyasskin: Also, any personal information that was incorrectly included in a signed exchange would be the information of the aggregator that fetched the SXG, and not the end user.

The use of `<link rel=alternate>` to identify the SXG for the current page could inform the UA to omit credentials in fetching that SXG, which would prevent any personal information from being accidentally included.

explainers/signed-exchange-subresource-subtitution-explainer.md
```diff
@@ -0,0 +1,455 @@
...
+declare exactly what version of the subresource is compatible with this main
+resource.
+   ```
+   Link: <https://cdn.publisher.example/lib.js>;
+         rel="allowed-alt-sxg";
+         header-integrity="sha256-..."
+   ```
+
+
+### Recursive subresource signed exchange prefetch
```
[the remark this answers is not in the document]
horo-t: I have one question.

I think "allowed-alt-sxg" is useful even when the HTML is not served via signed exchange.

1. An aggregator site prefetches "https://aggregator.example/publisher.example/lib.js.sxg".
2. The user clicks a link to "https://publisher.example/article.html".
3. The HTTP response header has `Link: <https://publisher.example/lib.js>; rel="allowed-alt-sxg"; header-integrity="sha256-..."` header.
4. When the UA detects <script src="./lib.js">, the UA checks the header-integrity attribute value of the link rel="allowed-alt-sxg" header and the header-integrity value of "lib.js.sxg" in cache:
   - If they match, the UA will use "lib.js.sxg" in the cache.
   - Otherwise, the UA will fetch "https://publisher.example/lib.js".

Do you think this use case is not useful?
If so, I will remove this use case and make "allowed-alt-sxg" work only when the HTML is served via signed exchange.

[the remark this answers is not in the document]
jyasskin: I feel like that's unlikely to be helpful: the referrer has to guess that a page with a particular set of preloaded and replaceable subresources will be the next link clicked, without being able to predict exactly which link it is? I do agree that it's technically possible.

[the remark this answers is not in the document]
horo-t: OK. I agree with you.
I will remove this use case, and merge "Privacy-preserving prefetching subresources" and "Recursive subresource signed exchange prefetch".

explainers/signed-exchange-subresource-subtitution-explainer.md
```diff
@@ -0,0 +1,455 @@
...
+
+1. Introduce two new cache mechanism SignedExchangeCache and
+PrefetchedSignedExchangeCache in a Document.
+   - SignedExchangeCache keeps the signed exchanges which are
+prefetched from the Document.
+   - PrefetchedSignedExchangeCache keeps the signed
+exchanges which were passed from the referrer Document which triggered the
+navigation to the current Document.
+1. While processing
+"[prefetch](https://html.spec.whatwg.org/multipage/links.html#link-type-prefetch)"
```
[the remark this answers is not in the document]
horo-t: Done.

explainers/signed-exchange-subresource-subtitution-explainer.md
```diff
@@ -0,0 +1,455 @@
...
+link (eg: `<link rel="prefetch"
+href="https://cdn.feed.example/cdn.publisher.example/lib.js.sxg" as="script">`):
+   - If succeeded prefetching the signed exchange (including merkle integrity
```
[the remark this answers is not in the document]
horo-t: Done.

explainers/signed-exchange-subresource-subtitution-explainer.md
```diff
@@ -0,0 +1,455 @@
...
+   check of the body), stores the parsedExchange which is the result of
```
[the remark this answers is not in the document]
horo-t: Done.

explainers/signed-exchange-subresource-subtitution-explainer.md
```diff
@@ -0,0 +1,455 @@
...
+   [parsing a signed exchange](https://wicg.github.io/webpackage/loading.html#ref-for-parsing-a-signed-exchange)
```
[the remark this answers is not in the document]
horo-t: Done.

explainers/signed-exchange-subresource-subtitution-explainer.md
horo-t: Yes.
Added.

explainers/signed-exchange-subresource-subtitution-explainer.md
```diff
@@ -0,0 +1,455 @@
...
+   (= inner request URL and inner response) to the SignedExchangeCache.
+1. While
+[navigating across documents](https://html.spec.whatwg.org/multipage/browsing-the-web.html#navigating-across-documents),
+copy the parsedExchanges in the SignedExchangeCache of the source document to
+the PrefetchedSignedExchangeCache of the target document. This is intended to
+provide a way to pass the cached signed exchange across origins even if the UA
+is using origin isolated HTTPCache mechanism.
```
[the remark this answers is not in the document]
horo-t: Added
> (Note that [header integrity check](#header-integrity-of-signed-exchange) prohibits the distributor from passing a tracking ID to the publisher.)

explainers/signed-exchange-subresource-subtitution-explainer.md
horo-t: Changed to refer "linked resource fetch setup steps".

explainers/signed-exchange-subresource-subtitution-explainer.md
horo-t: Removed "Privacy-preserving prefetching subresources".

explainers/signed-exchange-subresource-subtitution-explainer.md
horo-t: Added.

explainers/signed-exchange-subresource-subtitution-explainer.md
horo-t: Done.
How about this?

explainers/signed-exchange-subresource-subtitution-explainer.md
horo-t: Changed to
> If there are multiple subresource preloads that are provided by prefetches on
the previous page in signed exchange format

explainers/signed-exchange-subresource-subtitution-explainer.md
horo-t: Done

explainers/signed-exchange-subresource-subtitution-explainer.md
horo-t: Done.

If we don't need to think about privacy, when the HTTP cache has "lib.js", UAs don't need to fetch "lib.js.sxg" when the UA see rel=preload and rel=allowed-alt-sxg and rel=alternate for better performance. But we can't do that for the privacy reason.

explainers/signed-exchange-subresource-subtitution-explainer.md
```diff
@@ -0,0 +1,405 @@
...
+The UA needs to check that this value of the prefetched subresource signed
+exchange is same as the header-integrity attribute of allowed-alt-sxg link
+header in the response from the publisher. This is intended to prevent the
+distributor from encoding a tracking ID into the subresource signed exchange.
+
+We can’t use the
+[SRI’s integrity](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity)
+for this purpose, because SRI’s integrity only covers the content body and not
+any of the headers. So if the UA use the SRI’s integrity value in
+allowed-alt-sxg link header, we can use the subresource signed exchanges to
```
jyasskin:
```suggestion
`allowed-alt-sxg` link header, a tracker can use signed exchanges that differ in headers like their `Content-Type` to transfer a user ID via a pattern of parsing failures.
```

explainers/signed-exchange-subresource-subtitution-explainer.md
```diff
@@ -0,0 +1,405 @@
...
+[application/signed-exchange format](https://wicg.github.io/webpackage/draft-yasskin-http-origin-signed-responses.html#application-signed-exchange)
+for integrity checking. This signedHeaders is *"the canonical serialization of
+the CBOR representation of the response headers of the exchange represented by
+the application/signed-exchange resource, excluding the Signature header
+field"*. So this value doesn’t change even if the publisher signs the content
+again or changes the signing key. This header-integrity value also guarantees
+the content body hasn't changed, because the signed headers are required to
+include a `Digest` header with a hash of the content body.
+
+The UA needs to check that this value of the prefetched subresource signed
```
jyasskin: I *think* this paragraph is already explained elsewhere.

explainers/signed-exchange-subresource-subtitution-explainer.md
```diff
@@ -0,0 +1,405 @@
...
+```
+
+This `header-integrity` parameter is the SHA256 hash value of the
+*signedHeaders* value from the
+[application/signed-exchange format](https://wicg.github.io/webpackage/draft-yasskin-http-origin-signed-responses.html#application-signed-exchange)
+for integrity checking. This signedHeaders is *"the canonical serialization of
+the CBOR representation of the response headers of the exchange represented by
+the application/signed-exchange resource, excluding the Signature header
+field"*. So this value doesn’t change even if the publisher signs the content
+again or changes the signing key. This header-integrity value also guarantees
```
jyasskin:
```suggestion
again or changes the signing key, but it does change if any of the headers or body change. (It catches changes to the body because a valid signed exchange's headers have to include a `Digest` value that covers the body.)
```

explainers/signed-exchange-subresource-subtitution-explainer.md
```diff
@@ -0,0 +1,405 @@
...
+        header-integrity="sha256-XXXXXX"
+```
+
+The UA recursively prefetches lib.js.sxg.
+
+If the user navigates to the expected article, both the main resource of the
+article and the script subresource are loaded from the prefetched signed
+exchanges.
+
+# Proposal
```
jyasskin:
```suggestion
# Algorithm sketch
```

explainers/signed-exchange-subresource-subtitution-explainer.md
```diff
@@ -0,0 +1,405 @@
...
+for this purpose, because SRI’s integrity only covers the content body and not
+any of the headers. So if the UA use the SRI’s integrity value in
+allowed-alt-sxg link header, we can use the subresource signed exchanges to
+track the users by changing content-type and detecting the image loading
+failure.
+
+## Multiple subresource signed exchanges
+If there are multiple subresource preloads that are provided by prefetches on
+the previous page in signed exchange format (example: script.js.sxg and
+image.jpg.sxg), the UA must check that there is no error in the any of the
```
jyasskin: This "must" is covered above. Is there anything in this section that's not already covered? If there is, try to prune out the repetitive bits by referring to the algorithm. If not, just remove this section.

explainers/signed-exchange-subresource-subtitution-explainer.md
```diff
@@ -0,0 +1,405 @@
...
+      rel="allowed-alt-sxg";
+      header-integrity="sha256-XXX"
+Link: <https://publisher.example/narrow.jpg>;
+      rel="allowed-alt-sxg";
+      header-integrity="sha256-YYY"
+```
+
+## Lifetime of the entry in SignedExchangeCache
+The UA must check both the
+[signature expire time](https://wicg.github.io/webpackage/loading.html#exchange-signature-expiration-time)
```
jyasskin:
```suggestion
[signature expiration time](https://wicg.github.io/webpackage/loading.html#exchange-signature-expiration-time)
```

explainers/signed-exchange-subresource-subtitution-explainer.md
```diff
@@ -0,0 +1,405 @@
...
+          rel="allowed-alt-sxg";
+          variants-05="accept;image/jpeg;image/webp";
+          variant-key-05="image/webp";
+          header-integrity="sha256-BBB"
+    Link: <https://publisher.example/image>; rel=preload; as=image;
+    ```
+If a UA supports WebP, the UA must prefetch **image_webp.sxg** holding a WebP
+image. Otherwise the UA must prefetch **image_jpeg.sxg** holding a JPEG image.
+
+## Security and Privacy Considerations
```
jyasskin:
```suggestion
# Security and Privacy Considerations
```

explainers/signed-exchange-subresource-subtitution-explainer.md
```diff
@@ -0,0 +1,405 @@
...
+- The publishers can know whether the referrer page has prefetched the signed
+exchange subresources or not by checking the resource timing information. But
+this only exposes 1 bit information (= yes or no) because UAs can use the cached
+signed exchange only if the required signed exchanges are all available.
+- To prevent tracking (user ID transfer), if the aggregator failed to prefetch a
+subresource that the main resource preloads, the UA must drop all of the
+subresource prefetches. If the aggregator prefetches a superset of the preloaded
+subresources, the UA must drop the ones that weren't preloaded.
+- The UA can use the prefetched signed exchange subresources, only when they
+were prefetched in the rimmediate eferrer page. This is intended to avoid
```
jyasskin:
```suggestion
were prefetched in the immediate referrer page. This is intended to avoid
```

explainers/signed-exchange-subresource-subtitution-explainer.md
```diff
@@ -0,0 +1,405 @@
...
+leaking the prefetching state to succeeding pages.
+- The UA check the header integrity value, so the distributor of the subresource
+signed exchange can’t inject arbitrary resources to the publisher’s page. This
+prevents distributors from sending tracking IDs to the publisher’s page.
+- The UA needs to check that the signed
+[request URL](https://wicg.github.io/webpackage/loading.html#exchange-request-url)
+matches the preload link and not just that the header-integrity value matches,
+since the header-integrity hash doesn't cover the request URL.
+- The UA must fetch the signed exchange subresource
```
jyasskin: It's more that the UA's decision of whether to fetch signed exchange subresources MUST NOT depend on whether the HTTP cache already contains the original subresource. The UA is free to avoid fetching them for other reasons.

explainers/signed-exchange-subresource-subtitution-explainer.md
```diff
@@ -0,0 +1,405 @@
...
+(https://cdn.feed.example/cdn.publisher.example/lib.js.sxg) while processing the
+prefetch link element  (`<link rel="prefetch"
+href="https://cdn.feed.example/cdn.publisher.example/lib.js.sxg" as="script">`)
+even if the HTTP cache already contains the original subresource
+(http://cdn.publisher.example/lib.js). Otherwise it leaks the state of
+publisher’s site to the distributor of the signed exchange.
+- If a replaced subresource prefetch hasn't completed by the time the UA would
+start fetching it in the course of loading the next page, the UA must cancel
+that prefetch and fetch the resource from its original URL. This prevents the
+distributor from interfering the publisher’s page. (Eg. intentionally block or
```
jyasskin:
```suggestion
distributor from interfering the publisher’s page. (Eg. intentionally blocking or delaying...
```

explainers/signed-exchange-subresource-subtitution-explainer.md
```diff
@@ -0,0 +1,405 @@
...
+this only exposes 1 bit information (= yes or no) because UAs can use the cached
+signed exchange only if the required signed exchanges are all available.
+- To prevent tracking (user ID transfer), if the aggregator failed to prefetch a
+subresource that the main resource preloads, the UA must drop all of the
+subresource prefetches. If the aggregator prefetches a superset of the preloaded
+subresources, the UA must drop the ones that weren't preloaded.
+- The UA can use the prefetched signed exchange subresources, only when they
+were prefetched in the rimmediate eferrer page. This is intended to avoid
+leaking the prefetching state to succeeding pages.
+- The UA check the header integrity value, so the distributor of the subresource
```
jyasskin:
```suggestion
- The UA checks the header integrity value, so the distributor of the subresource
```

explainers/signed-exchange-subresource-subtitution-explainer.md
```diff
@@ -0,0 +1,405 @@
...
+signed exchange can’t inject arbitrary resources to the publisher’s page. This
+prevents distributors from sending tracking IDs to the publisher’s page.
```
jyasskin:
```suggestion
prevents distributors from sending mismatched versions or tracking IDs to the publisher’s page.
```

explainers/signed-exchange-subresource-subtitution-explainer.md
```diff
@@ -0,0 +1,405 @@
...
+1. Does this specification allow an origin some measure of control over a user
+agent’s native UI?
+    - No
+1. What temporary identifiers might this this specification create or expose to
+the web?
+    - None
+1. How does this specification distinguish between behavior in first-party and
+third-party contexts?
+    - This feature treats all entities (the distributor of signed exchange, the
+    publisher site, the origin of subresource URL) as third-party origins.
```
jyasskin: I think the publisher's site winds up treated as a first-party origin, since it gets to be top-level after the navigation.

The behavior doesn't line neatly up under the heading of first- vs third-party, but I think we're trying to treat the distributor as a third-party, and of course cross-origin subresources are third-parties.

horo-t: Done.

explainers/signed-exchange-subresource-subtitution-explainer.md
```diff
@@ -0,0 +1,405 @@
...
+
+The UA recursively prefetches lib.js.sxg.
+
+If the user navigates to the expected article, both the main resource of the
+article and the script subresource are loaded from the prefetched signed
+exchanges.
+
+# Proposal
+
+- While prenavigating an HTML resource in signed exchange format:
```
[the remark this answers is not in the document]
horo-t: Added a link to the github issue.

explainers/signed-exchange-subresource-subtitution-explainer.md
```diff
@@ -0,0 +1,405 @@
...
+    the original resource URL.
+    1. If the UA successfully prefetches the signed exchange (including the
+    merkle integrity check of the body), it stores the parsedExchange which is
+    the result of
+    [parsing a signed exchange](https://wicg.github.io/webpackage/loading.html#parsing-a-signed-exchange)
+    (= inner request URL and inner response) to a new cache mechanism which is
+    attached to the Document.
+- While
+[navigating across documents](https://html.spec.whatwg.org/multipage/browsing-the-web.html#navigating-across-documents),
+the UA copies the parsedExchanges in the cache of the source document to the
```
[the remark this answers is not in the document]
horo-t: Done.

PR #543, opened 2019-12-12
make certurl.CreateOCSPRequest compliant with rfc5019

tomokinat:
https://tools.ietf.org/html/rfc5019#section-5 requires a OCSP client to use GET for small enough OCSP requests. Although [this is mainly for enabling HTTP caching](https://tools.ietf.org/html/rfc2560#appendix-A.1.1) and `certurl.FetchOCSPResponse` doesn't cache any response, it's worth supporting as `certurl.CreateOCSPRequest` is also exported.

In order to add a test which checks the "POST" side of the function, we need data of `len("http://<responder>/<encoded OCSP req>") > 255`. I though like creating a dummy cert whose responder url is super long was the easiest (as I wrote in the test tentatively), I don't know where the existing `test-cert.pem` came from, which has a responder url "http://ocsp.digicert.com" and doesn't look like a "dummy". Would appreciate some guidance.


irori:

go/signedexchange/certurl/ocsp.go
```diff
@@ -3,39 +3,61 @@ package certurl
...
-	request, err := http.NewRequest("POST", ocspUrl, bytes.NewReader(ocspRequest))
-	if err != nil {
-		return nil, err
+
+	// Use GET for small requests per https://tools.ietf.org/html/rfc5019#section-5.
+
+	// Use QueryEscape here instead of PathEscape to escape not only '/' but also '+' and '='
+	// to align with the exmaple in https://tools.ietf.org/html/rfc5019#section-5.
+	// Note that the string to be escaped doesn't contain other symbols as it is base64 encoded.
+	getUrl, err := ocspUrl.Parse(url.QueryEscape(base64.StdEncoding.EncodeToString(ocspRequest)))
```
irori: Per https://tools.ietf.org/html/rfc2560#appendix-A.1.1, `getUrl` should be constructed as follows:
```
{url}/{url-encoding of base-64 encoding of the DER encoding of the OCSPRequest}
```
`ocspUrl.Parse(...)` may produce different result, for example if `ocspUrl` is `http://example.com/ocsp`.

go/signedexchange/certurl/ocsp_test.go
```diff
@@ -1,33 +1,61 @@
...
 	"io/ioutil"
 	"testing"
-
-	. "github.com/WICG/webpackage/go/signedexchange/certurl"
-	"github.com/WICG/webpackage/go/signedexchange"
 )
 
-func TestOCSP(t *testing.T) {
-	expectedResponderURL := "http://ocsp.digicert.com"
+func TestCreateOCSPRequestSamll(t *testing.T) {
```
irori: Samll -> Small

go/signedexchange/certurl/ocsp.go
tomokinat: Thanks, fixed it.

go/signedexchange/certurl/ocsp_test.go
tomokinat: Done.

irori:
Thanks. Would you add `test-cert-long.pem` and unskip the new test?

go/signedexchange/certurl/ocsp.go
```diff
@@ -10,8 +10,25 @@ import (
...
+func createGetURL(responderURL, escapedData string) string {
+	// GET URL is constructed as follows:
+	//   {url}/{url-encoding of base-64 encoding of the DER encoding of the OCSPRequest}
+	// Code below is for avoiding duplicated slashes, except for existing ones in {url}.
+	// Common ways to do this with "path" or "net/url" don't work as
+	// - base64 encoded data may include slashes literally and
+	// - even if url-escaped beforehand it may result in double-escaping by url.URL methods
+	// unless we take extra care with url.Path and url.RawPath.
+	// We want to keep {url} as it is rather than path.Clean it, so we stick to a basic string manipulation.
+	if strings.HasSuffix(responderURL, "/") {
```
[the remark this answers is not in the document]
tomokinat: Then I'd make it align with Chromium, thanks!

tomokinat:
I was going to wait for response from @hajimehoshi , but anyway I added one and started to count that test. I believe this PR is now ready to merge, except for few points I forgot to mention:

- I only added test-cert-long.pem to align with existing test-cert.pem although there were several intermediate products.
- I made command-line preferGET flag default to false for backwards compatibility and as there's no strong reason to use GET if we don't cache the response.

Do they look good to you?

PR #546, opened 2019-12-16
go: Fix test names in sct_test.go

irori:
Thanks @kumagi for noticing this.


tomokinat:
nit: I'd suggest `t.Fatalf` rather than `t.Errorf` and then `return`

irori:
> nit: I'd suggest `t.Fatalf` rather than `t.Errorf` and then `return`

Done.

PR #547, opened 2019-12-18
js/bundle: Add command line tool

irori:
This adds `gen-webbundle` CLI tool to the `wbn` package, which enables JS developers to build a web bundle from a local directory, without having to install the Go tools.

Internally it uses `addFilesRecursively`, a new high-level method of `BundleBuilder`.


js/bundle/src/encoder.ts
```diff
@@ -66,6 +69,35 @@ export class BundleBuilder {
...
+    for (const file of files) {
+      const filePath = path.join(dir, file);
+      if (fs.statSync(filePath).isDirectory()) {
+        this.addFilesRecursively(filePath, baseURL + file + '/');
+      } else if (file === 'index.html') {
+        // If the file name is 'index.html', create an entry for baseURL itself
+        // and another entry for baseURL/index.html which redirects to baseURL.
+        // This matches the behavior of gen-bundle.
+        this.addFile(filePath, baseURL);
+        this.addExchange(baseURL + file, 301, { Location: './' }, '');
```
sisidovski: My understanding might be wrong though, if there is a file located at `/foo/bar/baz/index.html`, this code looks redirecting to `baseURL/index.html`. Is that a correct behavior?

js/bundle/package.json
```diff
@@ -9,6 +9,9 @@
     "build": "tsc",
     "test": "jasmine tests/*.js"
   },
+  "bin": {
+    "gen-webbundle": "./bin/gen-webbundle"
```
sisidovski: IMO: Just using package name (wbn) as a command name does more make sense for me. Why do you make this name?

js/bundle/src/encoder.ts
irori: Note that the recursive call at line 88 updates `baseURL` adding the directory name.

Here, `baseURL`'s value should be `<user-specified-baseURL>/foo/bar/baz/`, so this exchange would redirect `<user-specified-baseURL>/foo/bar/baz/index.html` to `<user-specified-baseURL>/foo/bar/baz/`.

js/bundle/package.json
irori: Yeah, that makes sense to me. Renamed the command.

We may add other CLI features in the future (such as `dump-bundle` equivalent), but then we could switch to the subcommand style (e.g. `wbn gen ...` / `wbn dump ...`).

irori:
@sisidovski Thank you for the review!
@kinu Would you approve this PR?

js/bundle/src/encoder.ts
sisidovski: Ah! I was missing it! Thank you for the comment!

irori:
Tweaked a bit: changed the argument order of `addFiles(Recursively)` to take URL as first argument, for consistency with `addExchange()`.

PR #548, opened 2019-12-19
js/bundle: Allow method chaining in bundle builder

irori:
This makes BundleBuilder methods return the builder itself, so that callers can chain method calls, like this:
```javascript
  const bundle = (new BundleBuilder(primaryURL))
    .setManifestURL(...)
    .addExchange(...)
    .addExchange(...)
    .createBundle();
```
This is a common idiom of Builder classes.


irori:
@kinu ptl

PR #552, opened 2020-02-04
Update the unsigned-bundle explainer

jyasskin:
To prepare for TAG and Mozilla review.


explainers/navigation-to-unsigned-bundles.md
```diff
@@ -0,0 +1,363 @@
...
+- [Stakeholder feedback](#stakeholder-feedback)
+- [Acknowledgements](#acknowledgements)
+
+<!-- /TOC -->
+
+## Proposal
+
+We propose to use the [Web Bundle
+format](https://wicg.github.io/webpackage/draft-yasskin-wpack-bundled-exchanges.html),
+without signatures, to give users a format they can create from any website
```
kinu: I'm slightly worried if this may sound like it excludes any types of signatures, e.g. raw-key signed bundles or app-level signatures... while some use cases would need that.  (While, it may be good to start with this limited focus)

explainers/navigation-to-unsigned-bundles.md
```diff
@@ -0,0 +1,363 @@
...
+that claims it.
+
+### Loading a trusted unsigned bundle.
+
+The browser selects an **unsigned bundle scope** that's based on the bundle's
+URL in the same way the [service worker's scope
+restriction](https://w3c.github.io/ServiceWorker/#path-restriction) is based on
+the URL of the service worker script. If the bundle is served with the
+[`Service-Worker-Allowed`
+header](https://w3c.github.io/ServiceWorker/#service-worker-allowed), that sets
```
kinu: Oh so are we considering literally reusing Service-Worker-Allowed header here?  It feels weird to me while I agreed that it could work in the similar way to how SW scope works.

explainers/navigation-to-unsigned-bundles.md
```diff
@@ -0,0 +1,363 @@
...
+### Process of loading a bundle
+
+When the browser navigates to an `application/webbundle;v=___` resource, or
+opens a file that the filesystem identifies as that MIME type, it first parses
+the resource enough to pull out the bundle's [primary
+URL](https://wicg.github.io/webpackage/draft-yasskin-wpack-bundled-exchanges.html#name-load-a-bundles-metadata)
+and a set of flags to control loading (not yet defined in the format). One of
+these flags defines whether the claimed URLs in the bundle are expected to be
+trusted, which for unsigned bundles means the bundle is served from the origin
+that claims it.
```
kinu: Sorry, just to clarify... this claimed URL is valid only if the bundle itself is served from the origin, is that what this also means?

explainers/navigation-to-unsigned-bundles.md
```diff
@@ -0,0 +1,363 @@
...
+
+Because the user might save a bundle to a file whose path includes private
+information like their username, APIs like
+[`Window.location`](https://developer.mozilla.org/en-US/docs/Web/API/Window/location)
+and
+[`Document.URL`](https://developer.mozilla.org/en-US/docs/Web/API/Document/URL)
+have to return the claimed URL and not the full encoded URL. Avoiding the new
+scheme in these APIs will also probably improve compatibility.
+
+### Subresource loads with an attached bundle
```
kinu: Here this sounds like it's excluding navigations while the text below mentions navigations too

explainers/navigation-to-unsigned-bundles.md
```diff
@@ -0,0 +1,363 @@
...
+The bundle that's attached to the request above will eventually get attached to
+the environment settings object created for the response, from where subsequent
+fetches can use it.
+
+For trusted bundles, any fetch, both navigations and subresources, within the
+bundle's scope checks the bundle for a response before going to the network. If a navigation fetch finds a response inside the bundle,
+
+For untrusted bundles, all fetches should check inside the bundle before going
+to the network. [It's not clear whether subresource requests that aren't inside
+the bundle should be able to touch the network at all.](#network-access)
```
kinu: Maybe mention some options, e.g. some feature policy can be probably used together to restrict that?

explainers/navigation-to-unsigned-bundles.md
```diff
@@ -0,0 +1,363 @@
...
+user's name, which they might not want exposed to a website they saved for
+offline use.
+
+### Process of loading a bundle
+
+When the browser navigates to an `application/webbundle;v=___` resource, or
+opens a file that the filesystem identifies as that MIME type, it first parses
+the resource enough to pull out the bundle's [primary
+URL](https://wicg.github.io/webpackage/draft-yasskin-wpack-bundled-exchanges.html#name-load-a-bundles-metadata)
+and a set of flags to control loading (not yet defined in the format). One of
```
kinu: So here do we suppose we'll have some flag inside the format to say trusted vs untrusted?  Should we explicitly mention that the resulting 'trusted' flag set inside the browser may be different from the claimed one, e.g. if unsigned bundle says it wants to be trusted but it's served from different origins?  Or, should the flag only be able to say that it should be treated as 'untrusted' (i.e. downgrading only) even if an unsigned bundle is served from the same origin?

explainers/navigation-to-unsigned-bundles.md
```diff
@@ -0,0 +1,363 @@
...
+
+#### Save as MHTML
+
+This is currently supported by Chromium using the `Save As...` | `Webpage,
+Single File` dialog.
+
+MHTML isn't a random-access format, which means that the entire prefix of the
+file needs to be parsed to get to any particular resource. This probably
+wouldn't be an issue for single web pages, but for full websites or bundles of
+multiple websites, it can be slow.
```
kinu: Should we also mention content negotiation, or ability to have multiple resources for single location?
https://wicg.github.io/webpackage/draft-yasskin-wpack-use-cases.html#name-essential-2

explainers/navigation-to-unsigned-bundles.md
```diff
@@ -0,0 +1,363 @@
...
+
+#### Save as Web Archive
+
+This is currently supported by Safari using the `Save As...` | `Web Archive` dialog.
+
+A Web Archive is in the semi-documented binary plist format and isn't random
+access either. Web Archives also currently allow
```
kinu: Could we start with the basic shortcomings stem from the format (e.g. content negotiation etc) before talking about a specific wontfix'ed issue?

explainers/navigation-to-unsigned-bundles.md
```diff
@@ -0,0 +1,363 @@
...
+  their URLs with unique or semi-unique IDs. These can usually be removed by
+  removing query parameters, but if a significant number of users cleaned their
+  URLs, the tracking could move to path segments.
+
+## Considered alternatives
+
+### Alternate formats considered
+
+There are several existing ways to serialize a web page or website, but all of
+them have problems:
```
kinu: nit: problems -> shortcomings for the usage described in this explainer ...or something?  (For some particular use cases there can be reasons to prefer them to WBN, I suppose)

explainers/navigation-to-unsigned-bundles.md
jyasskin: I've added content negotiation to the list, but the bigger problem is the lack of documentation. It's hard to tell even what capabilities the format has.

explainers/navigation-to-unsigned-bundles.md
kinu: Treating it like a suborigin for same-origin untrusted bundle makes sense to me. I'm not so sure about the redirect for 'trusted' but cross-origin bundle, but no strong opinion on that one for now. (Quick cross-checking appreciated for potential subresource loading cases @horo-t @irori @hayatoito)

explainers/navigation-to-unsigned-bundles.md
kinu: I'm agreeing that we need something. Okay, maybe we can start with this and see what people say...

explainers/navigation-to-unsigned-bundles.md
```diff
@@ -0,0 +1,363 @@
...
+the URL of the service worker script. If the bundle is served with the
+[`Service-Worker-Allowed`
+header](https://w3c.github.io/ServiceWorker/#service-worker-allowed), that sets
+the unsigned bundle scope to the value of that header. (We are also considering
+using a differently-named header that applies only to bundles.)
+
+Bundles can also contain a
+[manifest](https://wicg.github.io/webpackage/draft-yasskin-wpack-bundled-exchanges.html#name-parsing-the-manifest-sectio)
+whose [`scope` key](https://www.w3.org/TR/appmanifest/#scope-member) can narrow
+the scope but not broaden it.
```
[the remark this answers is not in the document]
irori: I'm feeling inclined to drop this. PWAs having subresources out of its manifest's navigation scope is legitimate, but cannot be bundled as-is under this rule.

Probably the overhead of (waiting for the bytes + parsing) manifest is small, but since the internal redirect to the primary URL is on the critical path of bundle navigation I would like to keep it as fast as possible.

explainers/navigation-to-unsigned-bundles.md
jyasskin: My thinking about 'trusted'-but-cross-origin comes from using the same flag for signed-but-expired packages. We've talked about a need for non-publishers to be able to pick between having a signed-but-expired package redirect vs act-as-unsigned, and the 'trusted' flag seems appropriate for that, and cross-origin is a similar "can't verify the trust" situation.

horo-t: Re: potential subresource loading cases

For subresource loading cases like [this demo](https://docs.google.com/document/d/18fFrURT6xD1GnqJwCx2K3Z81Te9Iv-r2ZsCkOvDH_04/edit#), I don't think we need this 'trusted' flag in the bundle. And also we don’t need primaryUrl/fallbackUrl for subresource web bundle. So we may need to make primaryUrl/fallbackUrl optional.

kinu: Thanks. It'd be nice if the format can just accommodate both while loading spec part can differ on how to interpret these flags.  While... for navigation case we're probably good to start with this text.

PR #554, opened 2020-02-06
dump-bundle: Treat JavaScript resources as text

irori:
So that dump-bundle will print response body by default.

### DIFF
--- a/go/bundle/cmd/dump-bundle/main.go
+++ b/go/bundle/cmd/dump-bundle/main.go
@@ -4,6 +4,7 @@ import (
 	"flag"
 	"fmt"
 	"log"
+	"mime"
 	"os"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


go/bundle/cmd/dump-bundle/main.go
```diff
@@ -81,6 +81,10 @@ func DumpExchange(e *bundle.Exchange, b *bundle.Bundle, verifier *signature.Veri
 	return nil
 }
 
+func isTextType(mimeType string) bool {
+	return strings.HasPrefix(mimeType, "text/") || strings.HasPrefix(mimeType, "application/javascript")
```
hajimehoshi: `HasPrefix` for `application/javascript` seems a little odd. I think this can be `application/foobar+javascript`.

irori: The intention here is that this function should return true for all of the followings:

- `text/*`
- `application/javascript`
- `application/javascript; charset=utf-8` (this seems not very rare today)

It seems there's no registered media type of the form `application/foobar+javascript`.
https://www.iana.org/assignments/media-types/media-types.xhtml

(There're many `*/*+xml` and `*/*+json` types, but this patch is not trying to interpret those.)

hajimehoshi: Oh I see. How about using `mime` package?

https://play.golang.org/p/gLepcpuj_4a

irori: Done.

PR #559, opened 2020-02-21
Fix #556: describe the origin-computation algorithm better.

### LINKED ISSUE #556: "These are encoded so that the normal algorithm for computing an origin from a URL works" is false
https://github.com/WICG/webpackage/blob/master/explainers/navigation-to-unsigned-bundles.md#urls-for-bundle-components is incorrect as due to the scheme you'll get an opaque origin.

jyasskin:
I believe that with the encoding suggested here, we can just add "package" to the list at https://url.spec.whatwg.org/#concept-url-origin that includes "https", and that'll make the origin algorithm work for the new scheme. @annevk pointed out that my explanation was too brief; does this do a better job, or is there some other wording that would express it better?


annevk:
Seems fine. Is there a normative description available too somewhere?

jyasskin:
There's no specification yet.

PR #560, opened 2020-02-27
Sketch how package: URLs should be rendered.

jyasskin:
@annevk requested more information on how `package:` URLs should be rendered, so here's a sketch.

I haven't yet run this by @estark37 and her team. Emily, does this look like a reasonable start? Are you running away in horror? Do y'all have any better ideas?

### DIFF
--- a/explainers/navigation-to-unsigned-bundles.md
+++ b/explainers/navigation-to-unsigned-bundles.md
@@ -346,16 +346,15 @@ Since it's a new scheme, we have to think about how browsers should render
 spec](https://url.spec.whatwg.org/#url-rendering) is probably not appropriate
 for the default display of `package:` URLs, as users won't understand the
 significance of its "host" part,
-`https!,,distributor.example,otherpackage.wbn;q=query$https!,,publisher.example`.
+`https:,,distributor.example,otherpackage.wbn;q=query$https:,,publisher.example`.
 
 We suggest that, in places the browser would render just a URL's host, it render
 the host of the *bundle's* URL, so just `distributor.example` in the above
-example. When the user opens the full URL for editing, it should expand to the full `package:` URL with encoding.
-
-Browsers can investigate showing a more detailed view, with the two components
-displayed separately, or even a full browser of the possible resource URLs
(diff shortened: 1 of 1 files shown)


annevk:
Doesn't that mean that as far as the user is concerned the authority is with the publisher? (I.e., same as with HTTP URLs today.) And if that's the case, what warrants all the complexity again?

estark37:
I'd like to see if we can avoid exposing `package:` URLs in browser UI all together. They look terrifying :) Are there strong use cases for editing them directly? Could we say that either the MVP does not involve editing `package:` URLs _or_ the MVP must involve some sort of custom UI for selecting resources within the package?

Showing the publisher hostname seems technically correct, but also could potentially lead to a proliferation of "double address bars" like we saw with AMP documents; the publisher and/or distributor thinks that it will be confusing to have the distributor hostname showing in the omnibox, so they add a second address bar in the web content showing the publisher hostname -- leading to confusion and sadness. I'm still getting my head around the use cases for unsigned/untrusted bundles so I'm not sure if this seems likely to be a big problem or if I can come up with a better suggestion. Is it likely that the distributor hostname will be meaningless to humans, like a generic CDN domain? Or do you expect distributor hostnames to usually be pretty meaningful, like `offline-ebooks.com` or something?

A minor note is that the currently proposed v1 solution (show the publisher hostname, reveal the full URL on click) might be a significant implementation investment. Even just eliding `www` subdomain labels was a large implementation effort (as I understand it) because of edge cases with the cursor jumping around when you select, dealing with different ways to focus the omnibox, etc. So it could turn out that implementing custom UI for `package:` URLs is a simpler v1 approach anyway.

kinu:
Okay, maybe the modified proposal could be: show the distributor's URL in omnibox (ideally  making it non-editable), but internally use package: URL.  @estark37 let us know if you have further comments on this idea. Thanks!

jyasskin:
This is superseded by the [URL rendering bits](https://github.com/jyasskin/webpackage/blob/package-scheme-explainer/explainers/bundle-urls-and-origins.md#rendering-the-url-bar) in #587.

Issue #526, opened 2019-11-11
WebBundle for sub-resources


Jxck:
Current Spec seems based on top level url(html) + subresouces.
but I think it can use for bundling subresource like JS modules.

in current real world, JS modules are depends on tiny modules in npm.
if you write correct import maps and serve node_modules directory on server will solve dependencies
on the browser.

but npm modules are so tiny (substack way), so there cause tons of fetch likely over 100.
and Chrome Team reports this will cause IPC bottle neck.

https://docs.google.com/document/d/1ds9EEkcDGnt-iR8SAN-_7nsOfw7gsMfhZjzZ_QAIyjM/edit

finally they recommend to use webpack and make them in single bundle.js. and developers doing that.

this means we don't use ES modules directory, even if spec has done, and impl has almost done in browser. `import` notation is only a keyword for webpack, and `<script type=text/javascript src=bundle.js>`. too sad to see them.

if we can use WebBundles for bundling npm modules and fetch them in one time.
and also browser extract and handle them as if they are fetched each-by-each.
browser can cache as separated req/res pare, we can use them via `import`.

webbundle for subresources are also fine for WebFont(multi weight), CSS Sprite(bundle images), Favicon(apple-touch-icon, favicon.ico etc).

so I believe webbundle will change our web content deployment.

jyasskin:
I think this issue is fixed by #565, which documents that we think it's an important use case to solve.

PR #565, opened 2020-03-23
Add an explainer for subresource loading (#526)

hayatoito:
Add an explainer for subresource loading (#526).

cc: @jyasskin, @kinu, @irori, @horo-t

### DIFF
--- a/explainers/subresource-loading.md
+++ b/explainers/subresource-loading.md
@@ -1,0 +1,112 @@
+# Explainer: Subresource loading with Web Bundles
+
+Last updated: April 2020
+
+We propose a new approach to load a large number of resources efficiently using
+a format that allows multiple resources to be bundled, e.g.
+[Web Bundles](https://web.dev/web-bundles/).
+
+## Backgrounds
+
+- Loading many unbundled resources is still slower in 2020. We concluded that
+  [bundling was necessary in 2018](https://v8.dev/features/modules#bundle), and
+  our latest local measurement still suggests that.
+
+- The output of JS bundlers (e.g. webpack) doesn't interact well with the HTTP
+  cache. They are pretty good tools but configuring them to work in an optimal
+  way is tough, and sometimes they'are also incompatible with new requirements
+  like
+  [dynamic bundling](https://github.com/azukaru/progressive-fetching/blob/master/docs/dynamic-bundling/index.md)
+  (e.g. small edit with tree shaking could invalidate everything).
+
+- With JS bundlers, execution needs to wait for the full bytes to come. Ideally
+  loading multiple subresources should be able to utilize full streaming and
+  parallelization, but that's not possible if all resources are bundled as one
+  javascript. (For JS modules execution still needs to be waited for the entire
+  tree due to the current
+  [deterministic execution model](https://docs.google.com/document/d/1MJK0zigKbH4WFCKcHsWwFAzpU_DZppEAOpYJlIW7M7E/edit#heading=h.5652gd5ks5id))
+
+- Related issues: [#411](https://github.com/WICG/webpackage/issues/411),
+  [#526](https://github.com/WICG/webpackage/issues/526)
+
+## Proposal
+
+We propose a new `link rel=”webbundle”` notion [**1**] that lets the browser
+fetch bundled resources and attach them to the page. Such as:
+
+```html
+<link
+  rel="webbundle"
+  href="https://example.com/dir/subresources.wbn"
+  scope="https://example.com/scope/"
+/>
+```
+
+If the main document has such a `link` element, a browser must do:
+
+- A browser must fetch the specified Web Bundle, `subresources.wbn`.
+
+- All the subresources whose URLs are under the scope
+  `https://example.com/scope/` must [**2**] be loaded from the specified Web
+  Bundle.
+
+[**1**] The syntax is tentative. The API surface is not so important for us yet.
+Any alternative ideas are welcome. This is basically about mapping a set of
+resources to a particular bundled resource. Something like
+[Fetch maps proposal](https://discourse.wicg.io/t/proposal-fetch-maps/4259)
+might be able to extend to this case.
+
+[**2**] The exact behavior is TBD. There are several possible behaviors when the
+bundle does not contain the subresource; e.g. just fail, fallback to the
+network, or allow users to specify its behavior with a `link` element's
+attribute.
+
+The primary requirement to avoid fetching the same bytes twice is that "If a
+subresource under that scope is needed later in the document, that later fetch
+should block until at least the index of the bundle has downloaded to see if
+it's there."
+
+It seems secondary to then say that if the subresource within that scope isn't
+in the bundle, its fetch should fail or otherwise notify the developer: that
+just prevents delays in starting the subresource fetch.
+
+## Example
+
+### The bundle
+
+Suppose that the bundle, `subresources.wbn`, includes the following resources:
+
+```
+- https://example.com/scope/a.js (which depends on ./b.js)
+- https://example.com/scope/b.js
+- https://example.com/scope/c.png
+- … (omitted)
+```
+
+### The main document
+
+```html
+<link rel="webbundle"
+    href="https://example.com/dir/subresources.wbn"
+    scope="https://example.com/scope/"
+/>
+
+<script type=”module” src=”https://example.com/scope/a.js”></script>
+<img src=https://example.com/scope/c.png>
+```
+
+Then, a browser must fetch the bundle, `subresources.wbn`, and load
+subresources, `a.js`, `b.js`, and `c.png`, from the bundle.
+
+## Subsequent loading and Caching
+
+On the subsequent navigations for the same page/origin, it’s desirable that the
+resources that are already fetched in the bundle shouldn't be refetched. We're
+thinking about two possibilities here:
+
+1. Embed the details of version / cached resources in the URL
+2. Use a cache digest or something (and also embed it in the request)
+
+Then the server should be able to dynamically subset the Bundle and serve it.
+Note that here we assume this cache should be same-origin only therefore should
+not leak the cross-origin information.


hayatoito:
It looks ipr check said:

> hayatoito is unknown, they need to link their Github account with a W3C account.

Okay, I've just linked my GitHub acccount, @hayatoito, with W3C account (hito).  I guess this is sufficient to pass checks, hopefully.

hayatoito:
Regarding https://labs.w3.org/repo-manager/pr/id/WICG/webpackage/565,

> hayatoito should join the Web Platform Incubator Community Group .

I've also sent a request to join WICG.

hayatoito:
+cc: @yoavweiss

hayatoito:
+cc: @hiroshige-g

explainers/subresource-loading.md
```diff
@@ -0,0 +1,102 @@
...
+
+```html
+<link
+  rel="webbundle"
+  href="https://example.com/dir/subresources.wbn"
+  scope="https://example.org/scope/"
+/>
+```
+
+Both the `href` and the `scope` need to be the same-origin.
```
horo-t: I don't think they need to be the same-origin, and actually the example URLs are not the same-origin.
When I wrote [this example](https://docs.google.com/document/d/18fFrURT6xD1GnqJwCx2K3Z81Te9Iv-r2ZsCkOvDH_04/edit#heading=h.irqypigubog7), my idea was:
  - If the page is controlled by a service worker, the service worker can intercept any requests from the page and can respond with arbitrary resources.
  - So, it is safe (=no difference from service worker from the security point of view) even if they are not the same origin, when the page can access the web bundle in the CORS manner.

explainers/subresource-loading.md
```diff
@@ -0,0 +1,102 @@
...
+- Loading many unbundled resources is still slower in 2020. We concluded that
+  [bundling was necessary in 2018](https://v8.dev/features/modules#bundle), and
+  our latest local measurement still suggests that.
+
+- Caching stories with JS bundlers (e.g. webpack) is tough. They are pretty good
+  tools but configuring them to work in an optimal way is tough, and sometimes
+  also conflicts with new requirements like dynamic bundling (e.g. small edit
+  with tree shaking could invalidate everything).
+
+- With JS bundlers, parsing and execution needs to wait for the full bytes to
```
irori: I think this is not true for parsing.

Chrome is already doing streaming parsing / compiling for classic scripts (probably scaling it to more than one processor is hard, though), and for module scripts each module is parsed / compiled as soon as it is downloaded, without waiting for the entire module tree.

explainers/subresource-loading.md
```diff
@@ -0,0 +1,102 @@
...
+
+- A browser must fetch the specified Web Bundle, `subresources.wbn`.
+
+- All the subresources whose URLs are under the scope
+  `https://example.org/scope/` must [**2**] be loaded from the specified Web
+  Bundle.
+
+[**1**] The syntax is tentative. The API surface is not so important for us yet.
+Any alternative ideas are welcome. This is basically about mapping a set of
+resources to a particular bundled resource. Something like Fetch maps proposal
```
irori:
```suggestion
resources to a particular bundled resource. Something like [Fetch maps proposal](https://discourse.wicg.io/t/proposal-fetch-maps/4259)
```

explainers/subresource-loading.md
```diff
@@ -0,0 +1,102 @@
...
+Then, a browser must fetch the bundle, `subresources.wbn`, and load
+subresources, `a.js`, `b.js`, and `c.png`, from the bundle.
+
+## Subsequent loading and Caching
+
+On the subsequent navigations for the same page/origin, it’s desirable that the
+resources that are already fetched in the bundle shouldn't be refetched. We're
+thinking about two possibilities here:
+
+1. Embed the details of version / cached resources in the URL
```
irori: Could you expand this a bit? I'm not sure how this enables resource subsetting.

explainers/subresource-loading.md
hayatoito: Thanks.
@kinu, what do you think? I'm afraid that I misunderstood your suggestions.

I'm wondering what the explainer should mention here. We have 3 URLs:

1. The main document's URL
2. href
3. scope

I thought 2 and 3 need to be the same origin, however, it is okay that 1 is in a different origin.
It's a simple mistake I forgot to update example URLs.

Any suggestions? If we don't have a clear answer, I think it is okay to mention it as TBD.

explainers/subresource-loading.md
```diff
@@ -0,0 +1,102 @@
+# Explainer: Subresource loading with Web Bundles
+
+Last updated: March 2020
+
+We propose a new approach to load a large number of resources efficiently using
+a format that allows multiple resources be be bundled, e.g.
```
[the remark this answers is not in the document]
hayatoito: Ah, thanks! This is typo. "allows multiple resources *to* be bundled" might be better, I think.
I'll fix that.

explainers/subresource-loading.md
hayatoito: Thanks! Let me add a link.

explainers/subresource-loading.md
kinu: Yeah I suggested this but I remember I discussed that cross-origin seems okay too with you. I want to think through all the implications... for example it apparently complicates CSP story. But I agree that maybe we can just drop this for now.

explainers/subresource-loading.md
hayatoito: @kinu 
Do you have any opinions? I guess we don't have a concrete idea yet.

explainers/subresource-loading.md
kinu: Would just dropping 'parsing and' make sense?

explainers/subresource-loading.md
kinu: (^^I commented before I see Hayato's response) @hayatoito I was indeed thinking about 2 and 3 indeed, but thinking about this more it's a little unclear too. (As @horo-t mentioned SW could already override it)  I think I'm fine with just dropping this sentence.

hayatoito: Thanks! Okay, let's just drop this for now.
Once we have a more clear idea, let's add something back.

explainers/subresource-loading.md
hayatoito: Thanks. Let's drop the parsing.

hayatoito:
I guess we prefer squash merge in this repository. So I will squash the commits before merging, anyway.

explainers/subresource-loading.md
hayatoito: Can we address this in a follow-up commit? We can adress that later, I think.

irori: sgtm.

hayatoito:
I've squashed the commits.

explainers/subresource-loading.md
kinu: Yeah... I think I was thinking *something* like: Server can maintain versioned bundles, and client can send a request with some URL parameters that indicates the version of the bundle they already have in cache (so that the server can identify the updated resources and send them back).  This is mostly about updating-fetch for same bundle cases though, and haven't really thought about how to do for similarly looking bundle on different pages cases.

hayatoito: @kinu Thanks! I appreciate that! That's helpful. I think we can update the explainer with the more details once we can explore this area. Let me close this as resolved for now.

hayatoito:
@jyasskin Could you review this PR? It would be great if you could take a look.

For reviewers, please feel free to merge this. I think we can update the explainer with more details anytime later.

explainers/subresource-loading.md
```diff
@@ -0,0 +1,101 @@
...
+```
+
+Then, a browser must fetch the bundle, `subresources.wbn`, and load
+subresources, `a.js`, `b.js`, and `c.png`, from the bundle.
+
+## Subsequent loading and Caching
+
+On the subsequent navigations for the same page/origin, it’s desirable that the
+resources that are already fetched in the bundle shouldn't be refetched. We're
+thinking about two possibilities here:
```
jkrems, line 94: Both of those option sound like exclusion-based fetching. What we've observed while working on a similar system is that there are scale boundaries where it stops working. When the number of individual items is high, it breaks the length limits for URLs (1) and a digest becomes increasingly unreliable (2).

The third solution is to have an official way to batch resources by inclusion instead. It requires an "inclusion digest" or "inclusion list" instead of an "cached/exclusion digest" though.

kinu: @jkrems
Thanks! I agree that especially for dynamic bundling it seems to need some "inclusion signal" in requesting phase. I suspect it still could be benefited if it's combined with cache/exclusion approach for better cache efficiency?  A straw-idea could be:

1. The bundle request should express some hash or list that indicates the set of resources it wants.  Possibly the server and client can share a logic to compute the hash / variations for the session to determine what resources are needed.
2. Platform also attaches some digest that indicates exclusive-fetching information for caching.

This explainer didn't tell anything about 1. but probably it'd be better to mention it too. Does it make better sense, or I might be still missing something?

jkrems: Makes sense, I think you covered what I had in mind! And this definitely doesn't have to block this initial PR, more a comment towards future iteration.

explainers/subresource-loading.md
```diff
@@ -0,0 +1,101 @@
...
+
+## Backgrounds
+
+- Loading many unbundled resources is still slower in 2020. We concluded that
+  [bundling was necessary in 2018](https://v8.dev/features/modules#bundle), and
+  our latest local measurement still suggests that.
+
+- Caching stories with JS bundlers (e.g. webpack) is tough. They are pretty good
+  tools but configuring them to work in an optimal way is tough, and sometimes
+  also conflicts with new requirements like dynamic bundling (e.g. small edit
```
jyasskin, line 17:
```suggestion
  they're also incompatible with new requirements like dynamic bundling (e.g. small edit
```

And please link "dynamic bundling" to a description.

explainers/subresource-loading.md
```diff
@@ -0,0 +1,101 @@
...
+a format that allows multiple resources to be bundled, e.g.
+[Web Bundles](https://web.dev/web-bundles/).
+
+## Backgrounds
+
+- Loading many unbundled resources is still slower in 2020. We concluded that
+  [bundling was necessary in 2018](https://v8.dev/features/modules#bundle), and
+  our latest local measurement still suggests that.
+
+- Caching stories with JS bundlers (e.g. webpack) is tough. They are pretty good
```
jyasskin, line 15:
```suggestion
- The caching story with JS bundlers (e.g. webpack) is tough. They are pretty good
```

Or

```suggestion
- The output of JS bundlers (e.g. webpack) doesn't interact well with the HTTP cache. They are pretty good
```

explainers/subresource-loading.md
```diff
@@ -0,0 +1,101 @@
...
+  tools but configuring them to work in an optimal way is tough, and sometimes
+  also conflicts with new requirements like dynamic bundling (e.g. small edit
+  with tree shaking could invalidate everything).
+
+- With JS bundlers, execution needs to wait for the full bytes to come. Ideally
+  loading multiple subresources should be able to utilize full streaming and
+  parallelization, while it's not possible if all resources are bundled as one
```
jyasskin, line 22:
```suggestion
  parallelization, but that's not possible if all resources are bundled as one
```

explainers/subresource-loading.md
```diff
@@ -0,0 +1,101 @@
...
+  scope="https://example.com/scope/"
+/>
+```
+
+If the main document has such a `link` element, a browser must do:
+
+- A browser must fetch the specified Web Bundle, `subresources.wbn`.
+
+- All the subresources whose URLs are under the scope
+  `https://example.com/scope/` must [**2**] be loaded from the specified Web
```
jyasskin, line 48: The primary requirement to avoid fetching the same bytes twice is that "If a subresource under that scope is needed later in the document, that later fetch should block until at least the index of the bundle has downloaded to see if it's there."

It seems secondary to then say that if the subresource within that scope isn't in the bundle, its fetch should fail or otherwise notify the developer: that just prevents delays in starting the subresource fetch.

explainers/subresource-loading.md
```diff
@@ -0,0 +1,101 @@
...
+## Subsequent loading and Caching
+
+On the subsequent navigations for the same page/origin, it’s desirable that the
+resources that are already fetched in the bundle shouldn't be refetched. We're
+thinking about two possibilities here:
+
+1. Embed the details of version / cached resources in the URL
+2. Use a cache digest or something (and also embed it in the request)
+
+Then the server should be able to dynamically subset the Bundle and serve it.
```
jyasskin, line 99: A patch format also seems plausible. If the server looks at the If-None-Match request header to learn what version the client already has cached, it can send down a different response type to patch the bundle instead of sending down the whole bundle. I don't know of any existing systems that do that, so we should talk to more people if that direction sounds good. Roberto Peon has argued for patches.

explainers/subresource-loading.md
hayatoito: Thanks. Done.

hayatoito: Ah, I didn't add link "dynamic bundling". I think it's still an early idea, and we don't have a good description of that in public,

explainers/subresource-loading.md
hayatoito: Thanks! The latter sounds good to me.

explainers/subresource-loading.md
hayatoito: Thanks! Let me add these almost "as-is".  I think that would help the reader of this explainer to understand the possible behaviors we are thinking in more details.

explainers/subresource-loading.md
hayatoito: Thanks! Yes, we are currently discussing this topic actively in other places. I think we'll cover this topic in another explainer!

explainers/subresource-loading.md
hayatoito: Done. Thanks!

explainers/subresource-loading.md
kinu: This one, maybe?

https://github.com/azukaru/progressive-fetching/blob/master/docs/dynamic-bundling/index.md

hayatoito: Thanks! Let me add a link to that. I think that is okay for us.

hayatoito:
I've squashed the commits.
Reviewers, please feel free to merge the PR if you think it's ready.

kinu:
There are other ongoing discussions but I'm going to merge this (so that more discussion can be built on top of it)

PR #570, opened 2020-04-07
Add explicit use cases to the navigation-to-bundles explainer.

jyasskin:
And an initial diagram.

This could probably use diagrams for the other two use cases, but let me know what you think of the use case descriptions and the first diagram.

### DIFF
--- a/explainers/navigation-to-unsigned-bundles.md
+++ b/explainers/navigation-to-unsigned-bundles.md
@@ -41,7 +41,7 @@ https://datatracker.ietf.org/wg/wpack/about/.
    <details>
    <summary>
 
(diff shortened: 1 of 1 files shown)


explainers/navigation-to-unsigned-bundles.md
```diff
@@ -35,6 +29,47 @@ images or videos, users have native apps like
...
+
+   </details>
+
+1. Provide a way for sites like https://www.isocfoundation.org/ to package their
+   content so communities can fetch it once over a perhaps-expensive link and
+   then share it internally without each recipient needing to be online when
+   they get it.
+1. Provide a way for sites like https://archive.org/ to distribute archived
+   content without having to rewrite all its internal links.
+
```
kinu, line 71: Maybe also say "see also the alternate formats section (link) to see why we are proposing this with a new format" or something (but up to you)

jyasskin: Thanks, done.

Issue #468, opened 2019-07-26
Clarify what URL schemes are allowed in Bundled Exchanges


irori:
Current bundle spec says that [Bundled Exchanges parser must fail if URL have a fragment or credentials](https://wicg.github.io/webpackage/draft-yasskin-wpack-bundled-exchanges.html#index-section), but says nothing about the URL scheme.

In Signed Exchange, [exchange's URL must be `https:`](https://wicg.github.io/webpackage/draft-yasskin-httpbis-origin-signed-exchanges-impl.html#application-signed-exchange). Sould Bundled Exchanges follow this, or should also allow `http:`? Since exchange is defined as HTTP request+response pair, I think non-http(s) schemes don't make sense.

jyasskin:
Especially for books, there's some use for the [`cid:`](https://tools.ietf.org/html/rfc2392) scheme for things that exist within the package but aren't reachable with an HTTP lookup. I haven't thought through all the details of which schemes actually work.

We could restrict bundle resource URLs to http(s) anyway and loosen that when we get a more concrete use case, or we could restrict them just in the browser-side loading spec for now. I don't have a strong preference here.

irori:
Got it. For now, I'm going to restrict them to http(s) in Chromium implementation.

> We could restrict bundle resource URLs to http(s) anyway and loosen that when we get a more concrete use case, or we could restrict them just in the browser-side loading spec for now. I don't have a strong preference here.

I don't have a preference either.

horo-t:
I'm trying to implement "Save page as Web Bundle" feature in Chromium (https://crbug.com/1040752). This feature will provide users with a way to save the selialized current page as a Web Bundle file. This will behave almost same as the existing "Save page as MHTML" feature.

For serializing a page, we want to use the `cid:` scheme for two purposes.

1. URL for subframes:
   - We can’t use the original URL of subframe, because there may be the same URL subframes which contents are different when they are serialized. And also inline iframe `<iframe srcdoc=”...”>` doesn’t have a URL. So we need `cid:` scheme for it.
   - See this doc https://goo.gl/GYT7Br#heading=h.jb2ehfs7vtcw about how Chromium handles iframes while generating MHTML. 

1. Style elements
   - Chrome also uses `cid:` scheme for stlye element while generating MHTML.
   - See this doc https://goo.gl/GYT7Br#heading=h.1mjt2ly8g7u about how Chromium handles style elements while generating MHTML. 
[...]

PR #578, opened 2020-05-08
js/bundle: Allow non-http(s) URLs

irori:
This makes the JS WebBundle encoder be liberal about inner URL schemes.

The format spec currently does not say which URL scheme should be
allowed, which is discussed in #468.


irori:
Thanks!

PR #587, opened 2020-07-06
Add an explainer for the package: URL scheme

jyasskin:
This tries to identify and describe all the places that the URL gets exposed in interesting ways to the platform, and define how they'll work. It probably misses several.

You can see this rendered at https://github.com/jyasskin/webpackage/blob/package-scheme-explainer/explainers/bundle-urls-and-origins.md.

### DIFF
--- a/explainers/bundle-urls-and-origins.md
+++ b/explainers/bundle-urls-and-origins.md
@@ -40,6 +40,7 @@ Participate:
 - [Detailed design discussion](#detailed-design-discussion)
   - [Exactly how do we compose the package: URL?](#exactly-how-do-we-compose-the-package-url)
 - [Considered alternatives](#considered-alternatives)
+  - [Rely on storage partitioning](#rely-on-storage-partitioning)
   - [URL encoding variants](#url-encoding-variants)
     - [Fragment-based URL scheme](#fragment-based-url-scheme)
       - [Fragments and MIME types](#fragments-and-mime-types)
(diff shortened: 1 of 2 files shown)


explainers/bundle-urls-and-origins.md
```diff
@@ -0,0 +1,621 @@
...
+field to determine which recipients can read it.
+
+* When sending from outside the bundle to inside, or from inside one bundle to
+  inside a different bundle, the sender uses the full
+  `package:bundle-url$claimed-origin`.
+* When sending from inside the bundle to inside, the sender uses just the
+  claimed origin.
+* When sending from inside the bundle to outside, the sender uses the same
+  target origin that a sender outside the bundle would use. This means the
+  message could be intercepted by something inside the same bundle, which should
```
horo-t, line 361: Sorry, I don't understand correctly. Why the message could be intercepted by something inside the same bundle?

When "package:https:,,a.example.com$https:,,b.example/page.html" sends a message to "https://c.example/target.html", postMessage() should be called with "https://c.example" `targetOrigin`.
In this case, the message can't be intercepted anything inside the same bundle.
Is my understanding incorrect?

jyasskin: My understanding writing this was that if `package:https:,,a.example.com$https:,,b.example/page.html` (a.k.a. "Packaged-B") has a `WindowProxy` whose name it sees as `https://c.example/target.html`, that `WindowProxy` could be referring to either the "real" `https://c.example/target.html` (Internet-C) fetched over the internet or `package:https:,,a.example.com$https:,,c.example/target.html` (Packaged-C) retrieved from inside the same bundle. And even if Packaged-B has some reason to prefer its message be received by Internet-C, it can't use `targetOrigin` to ensure that, because the two are expressed the same inside the `targetOrigin` field.

In the design as it's written here, Packaged-B has absolutely no way to tell those two situations apart (other than knowledge baked in when the package was created). I don't see any harm from giving it a way in the platform to tell the situations apart, but it adds some complication that I haven't been able to justify with a compensating benefit.

jyasskin: I've added a similar example to the wording here.

horo-t: Thank you. Now I understand.

PR #591, opened 2020-08-12
Update loading spec to support subresource substitution

horo-t:
I'd like to update the "Loading Signed Exchanges" spec to support Signed Exchange subresource substitution.
The logic is copied from "Algorithm sketch" section of the explainer of [Signed Exchange subresource substitution](https://github.com/WICG/webpackage/blob/master/explainers/signed-exchange-subresource-subtitution-explainer.md#algorithm-sketch).

***
<a href="https://pr-preview.s3.amazonaws.com/horo-t/webpackage/pull/591.html" title="Last updated on Sep 4, 2020, 12:29 AM UTC (8466393)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/webpackage/591/7d82584...horo-t:8466393.html" title="Last updated on Sep 4, 2020, 12:29 AM UTC (8466393)">Diff</a>

### DIFF
--- a/loading.bs
+++ b/loading.bs
@@ -122,11 +122,11 @@ spec: 7230; urlPrefix: https://tools.ietf.org/html/rfc7230#
         text: token; url: section-3.2.6
 spec: RFC8288; urlPrefix: https://tools.ietf.org/html/rfc8288#
     type: dfn
-        text: Target Attributes list; url: section-2.2
+        text: Target Attribute; url: section-2.2
         text: Link Target; url: section-3.1
         text: Link Context; url: section-3.2
         text: Relation Type; url: section-3.3
-        text: Target Attributes; url: section-3.4
+        text: Serialisation-Defined Target Attribute; url: section-3.4
         text: Parsing a Link Field Value; url: appendix-B.2
     type: http-header
         text: Link; url: section-3
@@ -281,8 +281,9 @@ Rewrite [=request/clone|clone a request=] to run these steps:
 
 ## New response fields ## {#mp-response}
 
-A [=response=] has an associated <dfn for="response">came from a signed
-exchange</dfn> boolean. Unless stated otherwise, it is false.
+A [=response=] has an associated <dfn for="response" lt="came from a signed
+exchange|did not come from a signed exchange">came from a signed exchange</dfn>
+boolean. Unless stated otherwise, it is false.
 
 A [=response=] has an associated <dfn for="response">signed exchange outer
 header list</dfn> (a [=header list=]). Unless stated otherwise it is empty.
@@ -456,20 +457,20 @@ satisfy the `https://cdn.publisher.example/lib.js` `<script>` element.
     dfn-type=dfn>alternate signed exchange</dfn> at the [=Link Target=] of the
     [=alternate signed exchange link=].
 
-* New serialisation-defined [=Target Attributes=] on the <a http-header>Link</a>
+* New [=serialisation-defined Target Attributes=] on the <a http-header>Link</a>
     header.
 
     * The <dfn element-attr for="link">variants</dfn> target attribute has the
         same values as the <a http-header>Variants</a> Header Field.  There MUST
-        NOT be more than one <{link/variants}> attribute in a link-value; occurrences
-        after the first MUST be ignored by parsers. It is serialized to an HTTP
-        link-param by first encoding as if it were a <a http-header>Variants</a>
-        Header Field value, and then if there are characters that don't match
-        the <a grammar>token</a> production, encoding as a <a
-        grammar>quoted-string</a>.
+        NOT be more than one <{link/variants}> attribute in a link-value;
+        occurrences after the first MUST be ignored by parsers. It is serialized
+        to an HTTP link-param by first encoding as if it were a <a
+        http-header>Variants</a> Header Field value, and then if there are
+        characters that don't match the <a grammar>token</a> production,
+        encoding as a <a grammar>quoted-string</a>.
 
-        This parameter is used in <{link/rel/allowed-alt-sxg}> link and
-        [=alternate signed exchange link=] to indicate what representations are
+        This parameter is used in <{link/rel/allowed-alt-sxg}> links and
+        [=alternate signed exchange links=] to indicate what representations are
         available for the resource at the time that the response is produced.
 
     * The <dfn element-attr for="link">variant-key</dfn> target attribute has
@@ -481,8 +482,8 @@ satisfy the `https://cdn.publisher.example/lib.js` `<script>` element.
         characters that don't match the <a grammar>token</a> production,
         encoding as a <a grammar>quoted-string</a>.
 
-        This parameter is used in <{link/rel/allowed-alt-sxg}> link and
-        [=alternate signed exchange link=] to indicate one or more sets of
+        This parameter is used in <{link/rel/allowed-alt-sxg}> links and
+        [=alternate signed exchange links=] to indicate one or more sets of
         available-values that identify the representation of the resource.
 
 <h3 algorithm id="mp-document">New Document fields</h3>
@@ -494,7 +495,7 @@ section, add the following lines:
 <blockquote dfn-for="Document">
 
 The {{Document}} has a map of <dfn dfn>prefetched signed exchanges for
-navigation</dfn>, which is a [=map=] of [=URL=] to [=exchange=], initially
+navigation</dfn>, which is a [=map=] of [=URLs=] to [=exchanges=], initially
 empty.  <span class="note">This might merge with the set of prefetched resources
 when <{link/rel/prefetch}> is specified more completely.</span>
 
@@ -506,44 +507,45 @@ empty.
 
 <h3 algorithm id="mp-navigation-params">New navigation params struct field</h3>
 
-In the [=navigation params=], add the following items:
+In the [=navigation params=] struct, add the following items:
 
 <dl dfn-for="navigation params">
 
 : <dfn>prefetched subresource signed exchanges</dfn>
-:: A set of [=exchanges=], initially empty.
+:: A set of [=exchanges=], empty unless specified otherwise.
 
 </dl>
 
 <h3 algorithm id="mp-link-type-prefetch">Monkeypatch Link type "prefetch"</h3>
 
 Note: To prefetch the [=alternate signed exchanges=] when the UA receives a
-prefetched main resource signed exchange, this section monkeypatchs Link type
-"<{link/rel/prefetch}>". This is getting all the [=alternate signed exchange
-links=] from the outer response, and see if they are allowed by the inner
-response's <{link/rel/allowed-alt-sxg}> links, and if they are match with the
-inner response's <{link/rel/preload}> links.
+prefetched main resource signed exchange, this section monkeypatches Link type
+"<{link/rel/prefetch}>". This algorithm gets all the [=alternate signed exchange
+links=] from the outer response, checks if they are allowed by the inner
+response's <{link/rel/allowed-alt-sxg}> links, and if they are, associates them
+with the inner response's <{link/rel/preload}> links.
 
 Issue(w3c/resource-hints#77): Currently the behavior of recursive prefetch is
-not defined in the specs. This section monkeypatches Link type
-"<{link/rel/prefetch}>" to support only for the [=alternate signed exchanges=]
-subresources. When we will change the HTML spec to support the general recursive
-prefetching, this section must be made align with the change.
+not specified. This section monkeypatches Link type "<{link/rel/prefetch}>" to
+add support only for the [=alternate signed exchanges=] subresources. When we
+change the HTML spec to support general recursive prefetching, this section
+needs to align with that change.
 
-The [=process the linked resource=] given a link element |el|, boolean
-|success|, and [=response=] |response|:
+To [=process the linked resource|process this type of linked resource=]
+("<{link/rel/prefetch}>"), given a link element |el|, boolean |success|, and
+[=response=] |response|:
 
 1. If |success| is true, [=fire an event=] named {{HTMLElement/load}} at |el|.
 1. Otherwise, [=fire an event=] named {{HTMLElement/error}} at |el|.
-1. If the |success| is false, then return.
-1. If the |response|'s [=response/came from a signed exchange=] is false, then
+1. If |success| is false, then return.
+1. If the |response| [=response/did not come from a signed exchange=], then
     return.
-1. Let |clonedExchange| be the result of creating a new [=exchange=] with the
-    first item of |response|'s [=response/URL list=] and the result of
-    [=response/cloning=] |response|.
-1. [=map/Set=] |clonedExchange| to |el|'s [=Node/node document=]'s
-    [=Document/prefetched signed exchanges for navigation=][|clonedExchange|'s
-    [=exchange/request URL=]].
+1. Let |requestUrl| be the first item of |response|'s [=response/URL list=].
+1. Let |clonedExchange| be the result of creating a new [=exchange=] with
+    |requestUrl| and the result of [=response/cloning=] |response|.
+1. Let |prefetchedSignedExchanges| be |el|'s [=Node/node document=]'s
+    [=Document/prefetched signed exchanges for navigation=].
+1. [=map/Set=] |prefetchedSignedExchanges|[|requestUrl|] to |clonedExchange|.
 1. Let |outerLinkHeader| be the result of [=header list/getting=] `` `Link` ``
     from |response|'s [=response/signed exchange outer header list=].
 1. If |outerLinkHeader| is null, then return.
@@ -760,34 +762,34 @@ add the following steps:
                 links. This is intended to prevent the referrer page from
                 encoding a tracking ID into the set of subresources it
                 prefetches.
-            1. [=list/Append=] the [=alternate signed exchange peload info=]
+            1. [=list/Append=] the [=alternate signed exchange preload info=]
                 (|link|, |linkTarget|, |prefetched alternate exchange|) to
                 |preloadLinkItems|.
         1. For each |preloadLinkItem| of |preloadLinkItems|:
             1. Let |corsAttribute| be the [=CORS settings attribute=] of
-                |preloadLinkItem|'s [=alternate signed exchange peload
+                |preloadLinkItem|'s [=alternate signed exchange preload
                 info/link=]'s [=Target Attribute named=] `"crossorigin"`.
             1. Let |request| be the result of [=creating a potential-CORS
                 request=] given |preloadLinkItem|'s [=alternate signed exchange
-                peload info/target=], the empty string, and |corsAttribute|.
+                preload info/target=], the empty string, and |corsAttribute|.
             1. Set |request|'s [=request/synchronous flag=].
             1. Set |request|'s [=request/client=] to |document|.
             1. Set |request|'s [=request/cryptographic nonce metadata=] to
-                |preloadLinkItem|'s [=alternate signed exchange peload
+                |preloadLinkItem|'s [=alternate signed exchange preload
                 info/link=]'s [=Target Attribute named=] `"nonce"`.
             1. Set |request|'s [=request/integrity metadata=] to
-                |preloadLinkItem|'s [=alternate signed exchange peload
+                |preloadLinkItem|'s [=alternate signed exchange preload
                 info/link=]'s [=Target Attribute named=] `"integrity"`.
             1. Set |request|'s [=request/referrer policy=] to be the [=referrer
                 policy attribute=] of |preloadLinkItem|'s [=alternate signed
-                exchange peload info/link=]'s [=Target Attribute named=]
+                exchange preload info/link=]'s [=Target Attribute named=]
                 `"referrerpolicy"`.
             1. Set |request|'s [=request/destination=] to |preloadLinkItem|'s
-                [=alternate signed exchange peload info/link=]'s [=Target
+                [=alternate signed exchange preload info/link=]'s [=Target
                 Attribute named=] `"as"`.
             1. If |canLoadAlternateSxg| is true, then set |request|'s
                 [=request/stashed exchange=] to |preloadLinkItem|'s [=alternate
-                signed exchange peload info/prefetched alternate exchange=].
+                signed exchange preload info/prefetched alternate exchange=].
 
                 Note: When |canLoadAlternateSxg| if false or there is no
                 matching prefetched alternate exchange, the original resource
@@ -917,55 +919,57 @@ An exchange signature is a [=struct=] with the following items:
 
 <h3 dfn-type=dfn export>Allowed signed exchange link info</h3>
 
-An allowed signed exchange link info is a [=struct=] with the following items:
+An allowed signed exchange link info is a [=struct=] with the following items,
+holding a parsed <{link/rel/allowed-alt-sxg}> link:
 
 <dl dfn-for="allowed signed exchange link info">
 : <dfn>target</dfn>
 :: A [=Link Target=] of the link.
 : <dfn>header integrity</dfn>
-:: A [=string=] holding the value of the link's of [=Target Attributes
-    list|Target Attribute=] named `"header-integrity"`.
+:: A [=string=] holding the value of the link's [=Target Attribute=] named
+    `"header-integrity"`.
 : <dfn>variants</dfn>
-:: A [=string=] holding the value of the link's of [=Target Attributes
-    list|Target Attribute=] named `"variants"`, or null.
+:: A [=string=] holding the value of the link's [=Target Attribute=] named
+    `"variants"`, or null.
 : <dfn>variant key</dfn>
-:: A [=string=] holding the value of the link's of [=Target Attributes
-    list|Target Attribute=] named `"varivariant keyants"`, or null.
+:: A [=string=] holding the value of the link's [=Target Attribute=] named
+    `"variant key"`, or null.
 
 </dl>
 
 <h3 dfn-type=dfn export>Alternate signed exchange link info</h3>
 
-An alternate signed exchange link info is a [=struct=] with the following items:
+An alternate signed exchange link info is a [=struct=] with the following items,
+holding a parsed <{link/rel/alternate}> link:
 
 <dl dfn-for="alternate signed exchange link info">
 : <dfn>target</dfn>
-:: A [=Link Target=] of the link.
+:: The [=Link Target=] of the link.
 : <dfn>context</dfn>
-:: A [=string=] holding the value of the link's of [=Link Context=].
+:: A [=string=] holding the value of the link's [=Link Context=].
 : <dfn>variants</dfn>
-:: A [=string=] holding the value of the link's of [=Target Attributes
-    list|Target Attribute=] named `"variants"`, or null.
+:: A [=string=] holding the value of the link's [=Target Attribute=] named
+    `"variants"`, or null.
 : <dfn>variant key</dfn>
-:: A [=string=] holding the value of the link's of [=Target Attributes
-    list|Target Attribute=] named `"varivariant keyants"`, or null.
+:: A [=string=] holding the value of the link's [=Target Attribute=] named
+    `"variant key"`, or null.
 
 </dl>
 
-<h3 dfn-type=dfn export>Alternate signed exchange peload info</h3>
+<h3 dfn-type=dfn export>Alternate signed exchange preload info</h3>
 
-An alternate signed exchange peload info is a [=struct=] with the following items:
+An alternate signed exchange preload info is a [=struct=] with the following items:
 
-<dl dfn-for="alternate signed exchange peload info">
+<dl dfn-for="alternate signed exchange preload info">
 : <dfn>link</dfn>
 :: A link object [=Parsing a Link Field Value|parsed=] from a HTTP <a
     http-header>Link</a> header.
 : <dfn>target</dfn>
-:: A URL to be preloaded. <span class="note">This MAY be different from
-    [=alternate signed exchange peload info/link=]'s [=Link Target=], when
+:: A URL to be preloaded. <span class="note">This can be different from
+    [=alternate signed exchange preload info/link=]'s [=Link Target=], when
     `"imagesrcset"` is used for image preload.</span>
 : <dfn>prefetched alternate exchange</dfn>
-:: A [=exchange=] of the prefetched [=alternate signed exchange=], or null.
+:: The [=exchange=] of the prefetched [=alternate signed exchange=], or null.
 
 </dl>
 
@@ -1805,17 +1809,14 @@ To <dfn for="read buffer">Dump</dfn> a [=read buffer=] |input| to a
 
         </dl>
 
-<h4 algorithm id="dump-value-of-target-attributes">Get value of [=Target
-Attributes list|Target Attributes=]</h4>
+<h3 algorithm id="dump-value-of-target-attributes">Get value of [=Target Attributes=]</h3>
 
 A link header value |link|'s <dfn noexport>Target Attribute named</dfn> |name|
-is the second item of the first tuple of |link|'s [=Target Attributes
-list|Target Attributes=] whose first item matches the string |name|, or null if
-no such tuple is present.
+is the second item of the first tuple of |link|'s [=Target Attributes=] whose
+first item matches the string |name|, or null if no such tuple is present.
 
 
-<h4 algorithm id="get-allowed-signed-exchange-links">Get allowed signed exchange
-link info</h4>
+<h3 algorithm id="get-allowed-signed-exchange-links">Get allowed signed exchange link info</h3>
 
 The result of <dfn>getting allowed signed exchange link info</dfn> from a
 [=list=] |links| is returned by the following steps:
@@ -1835,8 +1836,8 @@ The result of <dfn>getting allowed signed exchange link info</dfn> from a
         |allowedSxgLinks|.
 1. Return |allowedSxgLinks|.
 
-<h4 algorithm id="get-alternate-signed-exchange-links">Get alternate signed
-exchange link info</h4>
+<h3 algorithm id="get-alternate-signed-exchange-links">Get alternate signed
+exchange link info</h3>
 
 The result of <dfn>getting alternate signed exchange link info</dfn> from a
 [=list=] |links| is returned by the following steps:
@@ -1844,9 +1845,8 @@ The result of <dfn>getting alternate signed exchange link info</dfn> from a
 1. Let |alternateSxgLinks| be an empty [=list=].
 1. For each |link| of |links|:
     1. If |link|'s [=Relation Type=] is not 'alternate', then continue.
-    1. If |link|'s [=Target Attributes list|Target Attributes=] doesn't
-        [=list/contain=] (`"type"`,`"application/signed-exchange"`), then
-        continue.
+    1. If |link|'s [=Target Attributes=] doesn't [=list/contain=]
+        (`"type"`,`"application/signed-exchange"`), then continue.
     1. Let |linkTarget| be |link|'s [=Link Target=].
     1. Let |linkContext| be |link|'s [=Link Context=].
     1. Let |linkVariants| be |link|'s [=Target Attribute named=] `"variants"`.
--- a/loading.bs
+++ b/loading.bs
@@ -557,8 +557,8 @@ add the following steps:
     9. Let |allowed alternate subresource signed exchange links| be null.
 
     10. If |sourceBrowsingContext|'s [=active document=]'s
-        [=Document/prefetched signed exchanges for navigation=] has a matching exchange
-        for the |request|'s [=request/url=], then:
+        [=Document/prefetched signed exchanges for navigation=] has a matching
+        exchange for the |request|'s [=request/url=], then:
 
         1. Let |clonedExchange| be the result of creating a new [=exchange=]
             with the matching exchange's [=exchange/request URL=] and the result
@@ -569,7 +569,7 @@ add the following steps:
             [=request/stashed exchange=].
 
         1. Copy <{link/rel/allowed-alt-sxg}> link of the exchange's
-            [=exchange/response=]'s inner response to
+            [=exchange/response=] to
             |allowed alternate subresource signed exchange links|.
 
 
@@ -583,10 +583,10 @@ add the following steps:
 
     19. Copy |sourceBrowsingContext|'s [=active document=]'s
         [=Document/prefetched subresource signed exchanges=]
-        to navigationParams's
+        to |navigationParams|'s
         [=navigation params/prefetched subresource signed exchanges=].
 
-    20. Set navigationParams's
+    20. Set |navigationParams|'s
         [=navigation params/allowed alternate subresource signed exchange links=]
         to |allowed alternate subresource signed exchange links|.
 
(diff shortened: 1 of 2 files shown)


loading.bs
```diff
@@ -331,6 +331,67 @@ Note: Applying the signed exchange's response here has the effect of letting a
...
+<h3 algorithm id="mp-fetching-the-resource-hint-link">
+Monkeypatch Fetching the resource hint link</h3>
+In [Fetching the resource hint
+link](https://w3c.github.io/resource-hints/#fetching-the-resource-hint-link),
+add the following lines:
+
+The user agent SHOULD prefetch the alternate subresources signed exchanges if
+declared in link headers in the outer and the inner response header of the
+signed exchanges by running the following steps:
+    1. When the UA detects a "preload" link HTTP header in a signed exchange
```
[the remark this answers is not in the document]
yoavweiss: `prenavigate` is indeed not currently defined. All we have is a few [past discussions](https://github.com/w3c/resource-hints/issues/82#issuecomment-532072736) :/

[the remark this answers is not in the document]
kinu: (As we also chatted elsewhere) I don't think this part specifically relies on `prenavigate` (while this one itself needs to have spec too), but rather depending on the unclarified behavior processing link headers won't be considered preprocessing the content (https://github.com/w3c/resource-hints/issues/77#issuecomment-360743041).  But yes this part also needs to be reflected in the spec text.

loading.bs
```diff
@@ -331,6 +331,67 @@ Note: Applying the signed exchange's response here has the effect of letting a
...
+Monkeypatch Link type "preload"</h3>
+
+The navigated-to document has a set of preloads for which it uses the
+allowed-alt-sxg link relation to declare that they can be served by signed
+exchanges. The UA either serves all of them from SXGs prefetched by the previous
+page, or none of them.
+So in [processing](https://www.w3.org/TR/preload/#processing) of [Link type
+"preload"](https://www.w3.org/TR/preload/#link-type-preload) add the following
+step:
+    1. For each preload, use the imagesrcset and imagesizes attributes to pick a
```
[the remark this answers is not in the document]
horo-t: While runnning [default fetch and process the linked resource](https://html.spec.whatwg.org/multipage/semantics.html#default-fetch-and-process-the-linked-resource), the [linked resource fetch setup steps for preload](https://html.spec.whatwg.org/multipage/links.html#link-type-preload:linked-resource-fetch-setup-steps) is called from step 10.
>  10. Run the linked resource fetch setup steps, given el and request. If the result is false, then return.

The [linked resource fetch setup steps for preload](https://html.spec.whatwg.org/multipage/links.html#link-type-preload:linked-resource-fetch-setup-steps) calls [select an image source](https://html.spec.whatwg.org/multipage/images.html#select-an-image-source) which calls [update the source set](https://html.spec.whatwg.org/multipage/images.html#update-the-source-set) to update `source set` using the imagesrcset and imagesizes attributes and returns `selected source` from the `source set`. And
> 4. Set request's url to url.

So I think we need the monkypatch after `4. If as is "image", then:` of [the linked resource fetch setup steps for preload](https://html.spec.whatwg.org/multipage/links.html#link-type-preload:linked-resource-fetch-setup-steps).

loading.bs
```diff
@@ -331,6 +331,67 @@ Note: Applying the signed exchange's response here has the effect of letting a
 newer HTTP cache entry override a signed exchange's content, and of not storing
 the signed exchange's response in the HTTP cache.
 
+# Subresource substitution # {#subresource-substitution}
+
+To support [Signed Exchange subresource substitution](https://github.com/WICG/webpackage/blob/master/explainers/signed-exchange-subresource-subtitution-explainer.md),
```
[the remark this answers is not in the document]
horo-t: Updated this section and introduced new terms **allowed-alt-sxg link**, **alternate signed exchange link** and **header integrity value**.

loading.bs
```diff
@@ -331,6 +331,67 @@ Note: Applying the signed exchange's response here has the effect of letting a
...
+this secion describes how web browsers prefetch subresource signed exchanges.
+
+<h3 algorithm id="mp-fetching-the-resource-hint-link">
+Monkeypatch Fetching the resource hint link</h3>
+In [Fetching the resource hint
+link](https://w3c.github.io/resource-hints/#fetching-the-resource-hint-link),
+add the following lines:
+
+The user agent SHOULD prefetch the alternate subresources signed exchanges if
```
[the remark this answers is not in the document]
horo-t: Introduced new term **alternate signed exchange link**.

loading.bs
```diff
@@ -331,6 +331,67 @@ Note: Applying the signed exchange's response here has the effect of letting a
...
+declared in link headers in the outer and the inner response header of the
+signed exchanges by running the following steps:
+    1. When the UA detects a "preload" link HTTP header in a signed exchange
+        inner response, check whether a matching “allowed-alt-sxg” link HTTP
```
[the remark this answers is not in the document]
horo-t: Introduced the terms **allowed-alt-sxg link**.

loading.bs
```diff
@@ -331,6 +331,67 @@ Note: Applying the signed exchange's response here has the effect of letting a
...
+    1. If the resulting signed exchange is valid and matches the allowed-alt-sxg
+        link, attach it to the top-level prefetch.
+
+<h3 algorithm id="mp-navigating-across-documents">
+Monkeypatch Navigating across documents</h3>
+In [navigating across documents](https://html.spec.whatwg.org/multipage/browsing-the-web.html#navigating-across-documents),
+add the following step:
+    1. Copy the signed exchanges that were prefetched above to the target
+        document except for the one that serves the navigation itself.
+- Note that as browsers move toward partitioned HTTP caches, the source
```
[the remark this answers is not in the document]
horo-t: Done.

loading.bs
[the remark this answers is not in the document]
horo-t: Changed this PR to monkeypatch [linked resource fetch setup steps for preload type of linked resource](https://html.spec.whatwg.org/multipage/links.html#link-type-preload:linked-resource-fetch-setup-steps).

loading.bs
```diff
@@ -331,6 +331,67 @@ Note: Applying the signed exchange's response here has the effect of letting a
...
+        was served with a matching “alternate” link HTTP header.
+    1. If the outer signed exchange did identify an alternate version of the
+        subresource, prefetch the subresource signed exchange.
+    1. If the resulting signed exchange is valid and matches the allowed-alt-sxg
+        link, attach it to the top-level prefetch.
+
+<h3 algorithm id="mp-navigating-across-documents">
+Monkeypatch Navigating across documents</h3>
+In [navigating across documents](https://html.spec.whatwg.org/multipage/browsing-the-web.html#navigating-across-documents),
+add the following step:
```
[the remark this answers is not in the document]
horo-t: Done

loading.bs
```diff
@@ -331,6 +331,67 @@ Note: Applying the signed exchange's response here has the effect of letting a
...
+    1. Copy the signed exchanges that were prefetched above to the target
+        document except for the one that serves the navigation itself.
```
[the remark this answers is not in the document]
horo-t: Introduced **prefetched subresource singed exchanges** and **allowed alternate subresource signed exchange links** in [navigation param struct](https://html.spec.whatwg.org/multipage/browsing-the-web.html#navigation-params), and **prefetched singed exchanges for navigation**, **prefetched subresource singed exchanges**, **transferred subresource singed exchanges** and **allowed alternate
subresource signed exchange links** in [document  object](https://html.spec.whatwg.org/multipage/dom.html#the-document-object)

loading.bs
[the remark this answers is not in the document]
horo-t: I think this monkeypatch does describes the behavior of processing the preload header in the inner response of the prefetched signed exchange.

horo-t:
@jyasskin @kinu 
Cold you please review the latest PR?
https://github.com/WICG/webpackage/pull/591/commits/99de0b338b9ae715160bf03217d92558a3ac1945

loading.bs
```diff
@@ -331,6 +338,222 @@ Note: Applying the signed exchange's response here has the effect of letting a
 newer HTTP cache entry override a signed exchange's content, and of not storing
 the signed exchange's response in the HTTP cache.
 
+# Subresource substitution # {#subresource-substitution}
+
+To let the UA know that subresources of a page it's prenavigating to, can
+be retrieved as signed exchanges from the source site, this section introduces
+an extension to the HTTP Link header.
+
+* A <dfn dfn-type=dfn>allowed-alt-sxg link</dfn> is a relationship that is used
```
jyasskin, line 347: Imitate the language used in https://html.spec.whatwg.org/multipage/links.html#link-type-prefetch when defining this new link type. It probably makes sense to describe this as a monkeypatch to HTML so we don't end up with duplication like https://www.w3.org/TR/resource-hints/#prefetch.

loading.bs
```diff
@@ -331,6 +338,222 @@ Note: Applying the signed exchange's response here has the effect of letting a
...
+    in HTTP <a http-header>Link</a> header of inner HTTP response of main
+    resource signed exchange to indicate that the subresource of the
+    [=link target=] can be loaded from different URL as a signed exchange. This
+    link has a
+    [rel](https://html.spec.whatwg.org/multipage/semantics.html#attr-link-rel)
+    parameter contains `allowed-alt-sxg` and a `header-integrity` parameter set
+    to the value of the [=header integrity value=] of the [=alternate signed
+    exchange=] 
+* A <dfn dfn-type=dfn>alternate signed exchange link</dfn> is a relationship
```
jyasskin, line 356: This is not actually a link relationship since it's not a keyword. Imitating https://www.w3.org/TR/resource-hints/#dfn-resource-hint-link, how about:

```suggestion
* An <dfn>alternate signed exchange link</dfn> is a `` `Link` `` header sent with a signed exchange |S|, with the [=alternate=] link type, the `type` parameter set to `application/signed-exchange`, and a [=Link Context=]. The [=Link Context=] MUST be the [=Link Target=] of a [=preload=] `` `Link` `` header inside the signed content of |S|, and the [=alternate signed exchange link=] means that this resource can also be found inside the signed exchange at the [=Link Target=] of the [=alternate signed exchange link=].

    <div class="example" id="example-alternate-signed-exchange-link">
    ```http
    Link: <...>; ...
    ```
  </div>
```

Add the Link Target (https://tools.ietf.org/html/rfc8288#section-3.1) and Link Context (https://tools.ietf.org/html/rfc8288#section-3.2) definitions to the `<pre class='anchors'>` so they appear correctly in the References section.

You may have to fiddle with the bikeshed markdown to get it to link to https://html.spec.whatwg.org/multipage/links.html#rel-alternate and look right. See https://tabatkins.github.io/bikeshed/#autolinking.

This would probably make more sense with a diagram showing where in the inner and outer signed exchanges the headers appear. Try to make one with https://plantuml.com/component-diagram#58bb821cf05c48a2?

loading.bs
```diff
@@ -331,6 +338,222 @@ Note: Applying the signed exchange's response here has the effect of letting a
...
+    signed exchange's headers have to include a `Digest` value that covers
+    the body.)
+
+<h3 algorithm id="mp-document">
+Monkeypatch Document</h3>
+In the end of
+[the Document object](https://html.spec.whatwg.org/multipage/dom.html#the-document-object)
+section, add the following lines:
+
+The <code>Document</code> has a <dfn dfn-type=dfn>prefetched singed exchanges
```
jyasskin, line 389:
```suggestion
The {{Document}} has a set of <dfn for="Document" dfn-type=dfn>prefetched singed exchanges
```

loading.bs
```diff
@@ -331,6 +338,222 @@ Note: Applying the signed exchange's response here has the effect of letting a
...
+for navigation</dfn>, which is a set of [=exchanges=], initially empty.
+
+The document has a <dfn dfn-for="document-obj" dfn-type=dfn>prefetched
```
jyasskin, line 392:
```suggestion
The {{Document}} has a set of <dfn for="Document" dfn-type=dfn>prefetched
```

And similarly for the rest.

loading.bs
horo-t: > You may have to fiddle with the bikeshed markdown to get it to link to https://html.spec.whatwg.org/multipage/links.html#rel-alternate and look right. See https://tabatkins.github.io/bikeshed/#autolinking.

Sorry I may not understand well how bikeshed works.
I created a PR to the HTML spec to export `rel` and `alternate`. https://github.com/whatwg/html/pull/5843
Do I need to wait for the PR to be merged?

loading.bs
```diff
@@ -331,6 +338,222 @@ Note: Applying the signed exchange's response here has the effect of letting a
...
+    parameter. This link has a
+    [rel](https://html.spec.whatwg.org/multipage/semantics.html#attr-link-rel)
+    parameter contains
+    [alternate](https://html.spec.whatwg.org/multipage/links.html#rel-alternate)
+    keyword and a
+    [type](https://html.spec.whatwg.org/multipage/links.html#attr-hyperlink-type)
+    parameter set to the value 'application/signed-exchange' and an [=anchor=]
+    parameter set to the original subresource URL.
+* A <dfn dfn-type=dfn>header integrity value</dfn> is the SHA256 hash value of
+    the |signedHeaders| value from the
```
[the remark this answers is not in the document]
horo-t: Done

loading.bs
horo-t: Done

loading.bs
horo-t: Done

loading.bs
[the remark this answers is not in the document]
horo-t: done

loading.bs
```diff
@@ -331,6 +338,222 @@ Note: Applying the signed exchange's response here has the effect of letting a
 newer HTTP cache entry override a signed exchange's content, and of not storing
 the signed exchange's response in the HTTP cache.
 
+# Subresource substitution # {#subresource-substitution}
+
+To let the UA know that subresources of a page it's prenavigating to, can
+be retrieved as signed exchanges from the source site, this section introduces
+an extension to the HTTP Link header.
```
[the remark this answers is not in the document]
horo-t: Done

loading.bs
```diff
@@ -331,6 +338,224 @@ Note: Applying the signed exchange's response here has the effect of letting a
...
+    link has a `header-integrity` parameter set to the value of the
+    [=exchange/header integrity value=] of the [=alternate signed exchange=].
+    <div class="example" id="example-allowed-alt-sxg-link">
+    ```http
+    Link: <https://cdn.publisher.example/lib.js>;
+            rel="preload";
+            as="script"
+    Link: <https://cdn.publisher.example/lib.js>;
+            rel="allowed-alt-sxg";
+            header-integrity="sha256-XXXXXX
```
jyasskin, line 364:
```suggestion
            header-integrity="sha256-XXXXXX"
```

loading.bs
```diff
@@ -331,6 +338,224 @@ Note: Applying the signed exchange's response here has the effect of letting a
...
+<h3 algorithm id="mp-document">
+Monkeypatch Document</h3>
+In the end of
+[the Document object](https://html.spec.whatwg.org/multipage/dom.html#the-document-object)
+section, add the following lines:
+
+The {{Document}} has a set of <dfn for="Document" dfn-type=dfn>prefetched singed
+exchanges for navigation</dfn>, which is a set of [=exchanges=], initially empty.
+
+The {{Document}} has a <dfn dfn-for="Document" dfn-type=dfn>prefetched
```
jyasskin, line 394:
```suggestion
The {{Document}} has a set of <dfn dfn-for="Document" dfn-type=dfn>prefetched
```

and similarly for the sets below.

horo-t:
I think 1304768f307ba220c8714a35a892c1c62a310a54 is ready for review.
@jyasskin Could you please review this again?
Thank you.

loading.bs
```diff
@@ -331,6 +371,304 @@ Note: Applying the signed exchange's response here has the effect of letting a
...
+    that held the [=alternate signed exchange link=].
+1. Let |L| be the set of <{link/rel/allowed-alt-sxg}> <a http-header>Link</a>
+    HTTP headers on |R| whose [=Link Target=] is the same as the [=alternate
+    signed exchange link=]'s [=Link Context=] and whose <{link/variants}>
+    parameter and <{link/variant-key}> parameter are the same as the [=alternate
+    signed exchange link=]'s parameters if they are present.
+1. If |L| is empty, return false.
+1. Let |preload links| be the set of <{link/rel/allowed-alt-sxg}> <a
+    http-header>Link</a> HTTP headers on |R|.
+1. For each |preload link| in |preload links|:
```
jyasskin, line 462: I would probably use this line to explain why we create link elements and requests that seem unused.

```suggestion
1. For each |preload link| in |preload links|, use the following steps to resolve the link's actual target, incorporating attributes like {<link/imagesrcset>} and <{link/imagesizes}>}:
```

loading.bs
```diff
@@ -331,6 +371,304 @@ Note: Applying the signed exchange's response here has the effect of letting a
...
+signed exchanges. The UA either serves all of them from SXGs prefetched by the
+previous page, or none of them. So in the
+[linked resource fetch setup steps for preload type of linked resource](https://html.spec.whatwg.org/multipage/links.html#link-type-preload:linked-resource-fetch-setup-steps)
+before
+
+> 5. Return true.
+
+add the following steps:
+
+    5. If |el| is for the main resource's HTTP <a http-header>Link</a> header,
```
jyasskin, line 632: I've run out of time to review today, so this comment is just a general direction rather than actually what needs to be fixed, but:

We're going to have to reconcile that the HTML spec pretends that it's dealing with `<link>` elements, but we're doing `Link` headers, and that HTML is trying to handle things incrementally, but we're putting a restriction on the overall set of `Link` headers. It might work to define a predicate on the response that runs over the full set of response headers to say whether to do the resource substitution at all, and then maybe you can avoid checking whether each element came from a header...

horo-t: Added note about the Issue(whatwg/html#4224).

loading.bs
```diff
@@ -331,6 +376,322 @@ Note: Applying the signed exchange's response here has the effect of letting a
...
+
+</dl>
+
+<h3 algorithm id="mp-fetching-the-resource-hint-link">Monkeypatch Fetching the resource hint link</h3>
+
+At the end of the [Fetching the resource hint
+link](https://w3c.github.io/resource-hints/#fetching-the-resource-hint-link)
+section in [[resource-hints]], add the following lines:
+
+The user agent SHOULD run the following steps if the response of a
```
yoavweiss, line 557: I think this should mokeypatch Fetch or HTML (if this is only meant for prenavigates) directly.
So when either processes a response, they need to call an algorithm defined here that would:
* Check if the response is a prefetch one (e.g. by checking its request's initiator) and otherwise bail
* Check if the response came from an SXG. Otherwise bail
* Create an exchange...

loading.bs
```diff
@@ -331,6 +376,322 @@ Note: Applying the signed exchange's response here has the effect of letting a
...
+        http-header>Variants</a> Header Field.
+
+    * The <dfn element-attr for="link">variant-key</dfn> parameter may be
+        present. This parameter is used in <{link/rel/allowed-alt-sxg}> link and
+        [=alternate signed exchange link=] to indicate one or more sets of
+        available-values that identify the representation of the resource.
+        This parameter is using the same format as <a
+        http-header>Variant-Key</a> Header Field.
+
+The [=appropriate time=] to [=fetch and process the linked resource=] for an
```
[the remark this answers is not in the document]
yoavweiss: As such, it should be called from somewhere.

loading.bs
```diff
@@ -331,6 +376,322 @@ Note: Applying the signed exchange's response here has the effect of letting a
...
+before
+
+> 9. While true:
+
+add the following steps:
+
+    9. Let |allowed alternate subresource signed exchange links| be null.
+
+    10. If |sourceBrowsingContext|'s [=active document=]'s
+        [=Document/prefetched signed exchanges for navigation=] has a matching
```
yoavweiss, line 577: "has a matching exchange" is a bit handwavy. Might be better to define "prefetched signed exchanged for navigation" as a [map](https://infra.spec.whatwg.org/#maps)

loading.bs
yoavweiss: Also "is for the main resource" is not really precise, and it's be better to have some boolean that indicates what you mean here.

horo-t:
I changed this PR to monkeypatch `Link type "prefetch"` for recursive prefetching of alternate signed exchange subresources, and monkeypatch `Page load processing model for HTML files` for Link rel=preload HTTP headers processing.

@jyasskin @yoavweiss 
Could you please review the latest patch?

loading.bs
```diff
@@ -116,6 +116,21 @@ spec: RFC7231; urlPrefix: https://tools.ietf.org/html/rfc7231#
...
+    type: dfn
+        text: Link Target; url: section-3.1
+    type: dfn
+        text: Link Context; url: section-3.2
+    type: dfn
+        text: Relation Type; url: section-3.3
+    type: dfn
+        text: Target Attributes; url: section-3.4
+    type: dfn
+        text: Parsing a Link Field Value; url: appendix-B.2
```
jyasskin, line 133: You don't have to repeat `type: dfn` for subsequent definitions.

```suggestion
    type: dfn
        text: Link Target; url: section-3.1
        text: Link Context; url: section-3.2
        text: Relation Type; url: section-3.3
        text: Target Attributes; url: section-3.4
        text: Parsing a Link Field Value; url: appendix-B.2
```

loading.bs
```diff
@@ -331,6 +390,433 @@ Note: Applying the signed exchange's response here has the effect of letting a
...
+    1. If |outerLink|'s [=Relation Type=] is not 'alternate', then continue.
+    1. If |outerLink|'s [=Target Attributes list|Target Attributes=] doesn't
+        [=list/contain=] (`"type"`,`"application/signed-exchange"`), then
+        continue.
+    1. Let |linkTarget| be |outerLink|'s [=Link Target=].
+    1. Let |linkContext| be |outerLink|'s [=Link Context=].
+    1. Let |outerLinkVariants| be the second item of the first tuple of
+        |outerLink|'s [=Target Attributes list|Target Attributes=] whose first
+        item matches the string `"variants"` or the empty string ("") if it is
+        not present.
```
jyasskin, line 544: I would probably define a small algorithm for this bit to make it more readable, since it's used so many times. E.g.

> `A link header value |link|'s <dfn noexport>Target Attribute named</dfn> |name| is the second item of the first tuple of |link|'s [=Target Attributes list|Target Attributes=] whose first item matches the string |name|, or the empty string ("") if no such tuple is present.`

And then use it like
```
Let |outerLinkVariants| be |outerLink|'s [=Target Attribute named=] `"variants"`.
```

loading.bs
```diff
@@ -331,6 +390,433 @@ Note: Applying the signed exchange's response here has the effect of letting a
...
+            |innerLink|'s [=Target Attributes list|Target Attributes=] whose
+            first item matches the string `"imagesrcset"` or the empty string
+            ("") if it is not present.
+        1. Let |imagesizes| be the second item of the first tuple of
+            |innerLink|'s [=Target Attributes list|Target Attributes=] whose
+            first item matches the string `"imagesizes"` or the empty string
+            ("") if it is not present.
+        1. Create a <{link}> element |linkElement| which <{link/href}> attribute
+            is |linkTarget|, and <{link/imagesrcset}> attribute is
+            |imagesrcset|, and <{link/imagesizes}> is |imagesizes|.
```
jyasskin, line 587:
```suggestion
        1. Create a <{link}> element |linkElement| whose <{link/href}> attribute
            is |linkTarget|, <{link/imagesrcset}> attribute is
            |imagesrcset|, and <{link/imagesizes}> attribute is |imagesizes|.
```

loading.bs
```diff
@@ -331,6 +390,433 @@ Note: Applying the signed exchange's response here has the effect of letting a
...
+        1. Let |selected source| and <var ignore=''>selected pixel density</var>
+            be the URL and pixel density that results from [=selecting an image
+            source=] given |linkElement|, respectively.
+        1. If |selected source| is not null, then set |linkTarget| to the result
+            of [=URL parser|parsing=] |selected source|, with a base URL of
+            |response|'s [=response/URL=].
+    1. For each |allowedSxgLink| of |allowedSxgLinks|:
+        1. If the first element isn't the same as |linkTarget|, then continue.
```
jyasskin, line 595:
```suggestion
        1. If |allowedSxgLink|'s linkTarget isn't the same as |linkTarget|, then continue.
```

It might be good to explicitly define the Allowed SXG Link struct so that you could link to its fields in the above line.

loading.bs
```diff
@@ -331,6 +390,433 @@ Note: Applying the signed exchange's response here has the effect of letting a
...
+        1. If |innerLinkVariants| is not the empty string (""), [=header
+            list/set=] `` `Variants` ``/|innerLinkVariants| in
+            |storedExchange|'s [=exchange/response=]'s [=response/header list=].
+        1. If |innerLinkVariantKey| is not the empty string (""), [=header
+            list/set=] `` `Variant-Key` ``/|innerLinkVariantKey| in
+            |storedExchange|'s [=exchange/response=]'s [=response/header list=].
+        1. Let |requestForMatch| be the result of [=creating a potential-CORS
+            request=] given |linkTarget|, the empty string, and [=No CORS=].
+        1. If |requestForMatch| doesn't [=matches the stored exchange=]
+            |storedExchange|, then continue.
```
jyasskin, line 609: We can probably refactor [=matches the stored exchange=] so this algorithm doesn't need to create a placeholder exchange, but no need to do that in this change.

loading.bs
```diff
@@ -331,6 +390,433 @@ Note: Applying the signed exchange's response here has the effect of letting a
...
+                    respectively.
+                1. If |selected source| is not null, then set |linkTarget| to
+                    the result of [=URL parser|parsing=] |selected source|, with
+                    a base URL of |navigationParams|'s [=navigation
+                    params/response=]'s [=response/URL=].
+            1. Let |headerIntegrity| be an empty string.
+            1. For each |allowedSxgLink| of |allowedSxgLinks|:
+                1. If the first element isn't the same as |linkTarget|, then
+                    continue.
+                1. Let |innerLinkVariants| be the second item of
```
jyasskin, line 745: Similarly down here, because [tuples are structs](https://infra.spec.whatwg.org/#tuple), all the items have names, which need to be implied or stated when the tuple is created. Here, the variable names that created the tuple work well as names, so you can just use `|allowedSxgLink|'s innerLinkVariants` or similar wherever it's used, instead of needing steps to pull them out into local variables.

loading.bs
```diff
@@ -331,6 +390,433 @@ Note: Applying the signed exchange's response here has the effect of letting a
...
+                    |storedExchange|, then continue.
+                1. Set |headerIntegrity| to the fourth item of |allowedSxgLink|.
+            1. Let |alternateSXG| be null.
+            1. For each |sxg| of |navigationParams|'s [=navigation
+                params/prefetched subresource signed exchanges=]:
+                1. If |headerIntegrity| is the same as the value of |sxg|'s
+                    [=exchange/response=]'s [=response/header integrity value=]
+                    encoded as a [[CSP]] <a grammar>hash-source</a>, then set
+                    |alternateSXG| to |sxg|.
+            1. If |headerIntegrity| is not an empty string and |alternateSXG| is
```
jyasskin, line 773: It wasn't obvious why it was ok for |headerIntegrity| to be an empty string here. I think it's that when it's empty, there was no allowed signed exchange, and so we don't need it to have been prefetched ... in which case maybe it'll be clearer to put the "If |headerIntegrity| is not an empty string" condition around both the for loop and what's left of this step?

loading.bs
```diff
@@ -331,6 +385,376 @@ Note: Applying the signed exchange's response here has the effect of letting a
...
+Monkeypatch Page load processing model for HTML files</h3>
+In
+[Page load processing model for HTML files](https://html.spec.whatwg.org/multipage/browsing-the-web.html#read-html)
+before
+
+> 2. Create an HTML parser and ...
+
+add the following steps:
+    2. Run the following steps [=in parallel=]:
+        1. Wait until |document|'s [=viewport=] is initialized.
```
yoavweiss, line 646: I don't think that's what you want to do here. You want to wait until the viewport is "known" (in practice, by scanning HTML tokens). Otherwise, IIUC, the viewport is initialized by the parser, which creation you're blocking here...
Might be worthwhile to add a note/TODO saying that this needs to be better defined if/when we'll define the preloadScanner

loading.bs
```diff
@@ -331,6 +385,376 @@ Note: Applying the signed exchange's response here has the effect of letting a
...
+        1. Create a <{link}> element |linkElement| whose <{link/href}> attribute
+            is |linkTarget|, and <{link/imagesrcset}> attribute is
+            |imagesrcset|, and <{link/imagesizes}> attribute is |imagesizes|.
+        1. Let |selected source| and <var ignore=''>selected pixel density</var>
+            be the URL and pixel density that results from [=selecting an image
+            source=] given |linkElement|, respectively.
+        1. If |selected source| is not null, then set |linkTarget| to the result
+            of [=URL parser|parsing=] |selected source|, with a base URL of
+            |response|'s [=response/URL=].
+    1. For each |allowedSxgLink| of |allowedSxgLinks|:
```
yoavweiss, line 549: This is an O(n^2) algorithm. Is this how we implementing this?
Not a blocker, but might be better to redefine this using a map, as a followup.

loading.bs
```diff
@@ -331,6 +385,376 @@ Note: Applying the signed exchange's response here has the effect of letting a
...
+                1. Let |selected source| and <var ignore=''>selected pixel
+                    density</var> be the URL and pixel density that results from
+                    [=selecting an image source=] given |linkElement|,
+                    respectively.
+                1. If |selected source| is not null, then set |linkTarget| to
+                    the result of [=URL parser|parsing=] |selected source|, with
+                    a base URL of |navigationParams|'s [=navigation
+                    params/response=]'s [=response/URL=].
+            1. Let |headerIntegrity| be null.
+            1. For each |allowedSxgLink| of |allowedSxgLinks|:
```
yoavweiss, line 680: Same O(n^2) comment as above

loading.bs
```diff
@@ -331,6 +385,376 @@ Note: Applying the signed exchange's response here has the effect of letting a
...
+                |preloadLinkItem|'s [=alternate signed exchange peload
+                info/link=]'s [=Target Attribute named=] `"integrity"`.
+            1. Set |request|'s [=request/referrer policy=] to be the [=referrer
+                policy attribute=] of |preloadLinkItem|'s [=alternate signed
+                exchange peload info/link=]'s [=Target Attribute named=]
+                `"referrerpolicy"`.
+            1. Set |request|'s [=request/destination=] to |preloadLinkItem|'s
+                [=alternate signed exchange peload info/link=]'s [=Target
+                Attribute named=] `"as"`.
+            1. If |canLoadAlternateSxg| is true, then set |request|'s
```
yoavweiss, line 745: I'm not sure I get this. It seems that the resource is loaded even if `canLoadAlternateSxg` is false.
If that's intentional, a better name could clarify that

loading.bs
```diff
@@ -331,6 +385,376 @@ Note: Applying the signed exchange's response here has the effect of letting a
...
+<dl dfn-for="navigation params">
+
+: <dfn>prefetched subresource signed exchanges</dfn>
+:: A set of [=exchanges=], initially empty.
+
+</dl>
+
+<h3 algorithm id="mp-link-type-prefetch">Monkeypatch Link type "prefetch"</h3>
+
+The [=process the linked resource=] given a link element |el|, boolean
```
yoavweiss, line 501: As the algorithms are rather lengthy, I think it would help readability to add a note here outlining (at a high level) what they are doing and why

loading.bs
```diff
@@ -331,6 +385,376 @@ Note: Applying the signed exchange's response here has the effect of letting a
...
+
+<h3 algorithm id="mp-page-load-processing-model-for-html-fiiles">
+Monkeypatch Page load processing model for HTML files</h3>
+In
+[Page load processing model for HTML files](https://html.spec.whatwg.org/multipage/browsing-the-web.html#read-html)
+before
+
+> 2. Create an HTML parser and ...
+
+add the following steps:
```
yoavweiss, line 644: Same comment here - a note would be helpful

horo-t: done

loading.bs
horo-t: Added Note.

loading.bs
horo-t: Added Note.

loading.bs
horo-t: I will use a map as a followup.

loading.bs
horo-t: Blink is using a map AlternateSignedExchangeResourceInfo.EntryMap.

I will do so as a followup.

loading.bs
```diff
@@ -331,6 +386,391 @@ Note: Applying the signed exchange's response here has the effect of letting a
...
+
+1. If |success| is true, [=fire an event=] named {{HTMLElement/load}} at |el|.
+1. Otherwise, [=fire an event=] named {{HTMLElement/error}} at |el|.
+1. If the |success| is false, then return.
+1. If the |response|'s [=response/came from a signed exchange=] is false, then
+    return.
+1. Let |clonedExchange| be the result of creating a new [=exchange=] with the
+    first item of |response|'s [=response/URL list=] and the result of
+    [=response/cloning=] |response|.
+1. [=map/Set=] |el|'s [=Node/node document=]'s [=Document/prefetched signed
```
yoavweiss, line 523: "Set an entry" would make this clearer without clicking the link

loading.bs
```diff
@@ -331,6 +386,391 @@ Note: Applying the signed exchange's response here has the effect of letting a
...
+: <dfn>prefetched subresource signed exchanges</dfn>
+:: A set of [=exchanges=], initially empty.
+
+</dl>
+
+<h3 algorithm id="mp-link-type-prefetch">Monkeypatch Link type "prefetch"</h3>
+
+Note: To prefetch the [=alternate signed exchanges=] when the UA receives a
+prefetched main resource signed exchange, this section monkeypatchs
+Link type "<{link/rel/prefetch}>".
```
yoavweiss, line 504: It would be good to make the note more descriptive, outlining that this is getting all the alternateLinks from the outer response, see if they're allowed by the inner response, matches preloads to allowed, etc.

loading.bs
```diff
@@ -331,6 +386,391 @@ Note: Applying the signed exchange's response here has the effect of letting a
...
+Note: As browsers move toward partitioned HTTP caches, the source document's
+    cache will likely be separate from the target's cache, so we can't just pass
+    prefetched content through the cache.
+
+
+<h3 algorithm id="mp-page-load-processing-model-for-html-fiiles">
+Monkeypatch Page load processing model for HTML files</h3>
+
+Note: To use the prefetched subresource signed exchanges after the navigation,
+this section monkeypatchs [=Page load processing model for HTML files=].
```
yoavweiss, line 645: Same as above, would be good to outline what the algorithm is doing

loading.bs
```diff
@@ -331,6 +386,391 @@ Note: Applying the signed exchange's response here has the effect of letting a
...
+                exchange peload info/link=]'s [=Target Attribute named=]
+                `"referrerpolicy"`.
+            1. Set |request|'s [=request/destination=] to |preloadLinkItem|'s
+                [=alternate signed exchange peload info/link=]'s [=Target
+                Attribute named=] `"as"`.
+            1. If |canLoadAlternateSxg| is true, then set |request|'s
+                [=request/stashed exchange=] to |preloadLinkItem|'s [=alternate
+                signed exchange peload info/prefetched alternate exchange=].
+
+                Note: When |canLoadAlternateSxg| if false or there is no
```
yoavweiss, line 767: Thanks! That makes things clearer! :)

loading.bs
```diff
@@ -331,6 +386,391 @@ Note: Applying the signed exchange's response here has the effect of letting a
...
+            `` from |navigationParams|'s [=navigation params/response=]'s
+            [=response/header list=].
+        1. If |linkHeader| is null, then return.
+        1. Let |links| be the result of [=Parsing a Link Field Value=] from
+            |linkHeader|.
+        1. Let |allowedSxgLinks| be the result of [=getting allowed signed
+            exchange link info=] from |links|.
+        1. Let |preloadLinkItems| be an empty [=list=].
+        1. Let |canLoadAlternateSxg| be true. 
+        1. For each |link| of |links|:
```
yoavweiss, line 675: Non blocking note: Might be worthwhile to coalesce the common parts of the 2 large fetching algorithms into a single one, called from both places. As is, there's a lot of repetition.

yoavweiss (approved):
LGTM % not-really-blocking comments

I still want @jyasskin to take a look before landing

horo-t (commented):
Thank you!

loading.bs
```diff
@@ -331,6 +392,412 @@ Note: Applying the signed exchange's response here has the effect of letting a
...
+            key=] in |storedExchange|'s [=exchange/response=]'s
+            [=response/header list=].
+        1. Let |requestForMatch| be the result of [=creating a potential-CORS
+            request=] given |linkTarget|, the empty string, and [=No CORS=].
+        1. If |requestForMatch| doesn't [=matches the stored exchange=]
+            |storedExchange|, then continue.
+        1. Let |alternateSxgUrl| be [=alternate signed exchange link
+            info/target=] of the first [=alternate signed exchange link info=]
+            of |alternateLinks| whose [=alternate signed exchange link
+            info/variants=] matches |allowedSxgLink|'s [=allowed signed exchange
```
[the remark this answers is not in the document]
horo-t: Yes. The implementation is checking the equality of strings in  alternate_signed_exchange_resource_info.cc

loading.bs
```diff
@@ -331,6 +392,412 @@ Note: Applying the signed exchange's response here has the effect of letting a
...
+            link info/variants=] and whose [=alternate signed exchange link
+            info/variant key=] matches |allowedSxgLink|'s [=allowed signed
+            exchange link info/variant key=] and whose [=alternate signed
+            exchange link info/context=] matches |linkTarget| or continue if it
```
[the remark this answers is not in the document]
horo-t: Yes. The implementation is checking the equality of URL in alternate_signed_exchange_resource_info.cc.
Sorry I'm not sure what kind of security problems exist here.

loading.bs
```diff
@@ -331,6 +392,417 @@ Note: Applying the signed exchange's response here has the effect of letting a
...
+                    `Variant-Key` ``/|allowedSxgLink|'s [=allowed signed
+                    exchange link info/variant key=] in |storedExchange|'s
+                    [=exchange/response=]'s [=response/header list=].
+                1. Let |requestForMatch| be the result of [=creating a
+                    potential-CORS request=] given |allowedSxgLink|'s [=allowed
+                    signed exchange link info/target=], the empty string, and
+                    [=No CORS=].
+                1. If |requestForMatch| [=doesn't match the stored exchange=]
+                    |storedExchange|, then continue.
+                1. Set |headerIntegrity| to |allowedSxgLink|'s [=allowed signed
```
jyasskin: This looks like it'll use the `header-integrity` of the last matching `allowed-alt-sxg` link if multiple such links match. Does that match your implementation?

jyasskin (approved):
I finally got through the whole thing. Thanks for bearing with me. I'll merge it now, and we can fix up any mistakes I made in another PR.

loading.bs
horo-t: No. Chrome is using the first allowed-alt-sxg link entry.h.

I will update the spec to make it closer to the Chrome's implementation.
(But it may take a few months because of my paternity leave.)

PR #607, opened 2020-12-04
go/signedexchange/README: Remove reference to allow-sxg-certs-without-extension flag

irori:
`chrome://flags/#allow-sxg-certs-without-extension` flag is [removed](https://crbug.com/862003) in Chrome M89.


hayatoito (approved):
LGTM.

PR #608, opened 2021-01-12
Introduce scopes attribute for subresource Web Bundles link.

horo-t:
Currently the explainer of "subresource loading with Web Bundles" is
using a new "scope" attribute of the link element. However the new
attribute should be "scopes" that accept multiple scopes, giving web
developers more flexibility. For example, in this case, some dynamic
contents can be under "https://example.com/dir/api/".

So this PR changes the explainer to use "scopes".

### DIFF
--- a/explainers/subresource-loading.md
+++ b/explainers/subresource-loading.md
@@ -179,19 +179,21 @@ compressed](https://github.com/yoavweiss/url_compression_experiments).
 
 Several other mechanisms are available to give the bundler more flexibility or to compress the resource list.
 
-#### Defining the scope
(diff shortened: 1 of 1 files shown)


hayatoito (approved):
Thanks! I think `scopes` would be an improvement.

littledan:
Do you have any more context on the use cases for `scopes`?

horo-t:
`scopes` attribute is useful when the writer of HTML don't know the all dependent subresources (eg: `import` of JS, @import of CSS, images referred by CSS).

PR #610, opened 2021-01-26
Write an index to the explainers.

jyasskin:
@philipwalton @khempenius, do you think this makes it easy enough to find the explainers?


yoavweiss (approved):
LGTM

philipwalton:
Thanks, this is definitely helpful. One question: what are your thoughts on putting all the content of the explainers.md page (since it's not a lot) into the main README? That might make it even easier to find for folks just driving by.

jyasskin:
@philipwalton Good idea; how's this?

philipwalton:
Looks great, thanks!

PR #611, opened 2021-01-27
dump-signedexchange: add -headerIntegrity flag

twifkak:
When specified, only prints the header-integrity value for the given signed
exchange, as needed for subresource substitution.


twifkak:
@horo-t Can you mark this PR as non-normative (per the ipr check), or do you know who can?

horo-t (approved):
lgtm
Thank you.

PR #617, opened 2021-02-08
Separating the primary URL into a section

### LINKED ISSUE #616: Should the primary URL be included in the bundle format?
Primary URL does not make much sense as a concept for some uses of resource bundles (e.g., for subresource loading). I'd propose that it be a separate, optional section (like `manifest`), not a required field for all usages.

littledan:
Closes #616

### DIFF
--- a/draft-yasskin-wpack-bundled-exchanges.md
+++ b/draft-yasskin-wpack-bundled-exchanges.md
@@ -136,7 +136,6 @@ schema:
 webbundle = [
   magic: h'F0 9F 8C 90 F0 9F 93 A6',
   version: bytes .size 4,
-  primary-url: whatwg-url,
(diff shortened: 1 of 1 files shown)


littledan:
This is just an idea to open for discussion. The idea here is to make the primary URL more thoroughly optional by moving it into a separate section. The main semantic difference here is that the primary URL can no longer server as a fallback (e.g., in case the magic number is wrong)--it's a bit questionable how/when this fallback would be useful, and the identification of this fallback is a bit out of step with other changes to this document which made it less operationally defined.

hayatoito:
We can close this PR, in favor of wpack-wg.
https://github.com/wpack-wg/bundled-responses/pull/3

Issue #612, opened 2021-02-02
Should content negotiation be part of the webpackage format?


littledan:
All of the applications on the Web that I know of will do content negotiation to decide which bundle to send, not within the bundle on the client side. I'm wondering whether the format should include client-side declarative content negotiation.

jyasskin:
Moved to https://github.com/wpack-wg/bundled-responses/issues/2.

PR #618, opened 2021-02-08
Remove content negotiation

littledan:
As proposed in #612

This is just to be a concrete reference point for discussion.


hayatoito:
I think we can close this, in favor of https://github.com/wpack-wg/bundled-responses/pull/7

PR #619, opened 2021-02-09
Remove the manifest and critical sections

littledan:
The idea of this PR is that the "core" Web Bundles specification does not include the manifest or critical sections, but that these sections can be defined in other documents. The motivation is to factor the Web Bundles specification to include just the minimal amount needed to support cases like subresource loading. Subresource loading does not need a manifest. For the critical section, if it is missing, the behavior is likely to be to not require any sections, so it can also be defined and interpreted just where it is needed for more detailed bundle types.

### DIFF
--- a/draft-yasskin-wpack-bundled-exchanges.md
+++ b/draft-yasskin-wpack-bundled-exchanges.md
@@ -18,7 +18,6 @@ author:
     email: jyasskin@chromium.org
 
 normative:
-  appmanifest: W3C.WD-appmanifest-20180523
   CBORbis: I-D.ietf-cbor-7049bis
   CDDL: RFC8610
   FETCH:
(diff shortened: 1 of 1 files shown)


jyasskin:
I'm friendly to the idea of pulling `"manifest"` out to an extension document, like we're doing for `"signatures"`.

I feel like `"critical"` needs to be in the core spec in order to be worth having at all, since its purpose is to be able to get all implementations to reject a bundle with a section they don't understand. It's possible we could replace the `"critical"` section entirely with your idea around defining an HTTP header to negotiate it, but we should make sure that's likely to work before deciding to get rid of the `"critical"` section.

littledan:
Yeah, maybe this should've been two separate PRs. I don't fully understand the full story around the "critical" section, and don't feel strongly that it needs to be removed; "discuss" seems like an appropriate label.

hayatoito:
I think we can close this PR, in favor of https://github.com/wpack-wg/bundled-responses/pull/6

PR #626, opened 2021-02-19
Set a request's destination to "webbundle"

hayatoito:
This is a sort of a monky patch until the upstream issue,
https://github.com/whatwg/fetch/issues/1120, is resolved.


littledan (approved):
Making a new destination seems like a good idea to me, e.g., so that all of the content negotiation headers are sent.

hayatoito:
WPT is here. https://chromium-review.googlesource.com/c/chromium/src/+/2706327

jyasskin (approved):
Thanks!

PR #627, opened 2021-02-22
Limit Subresource Web Bundles only in secure contexts

horo-t:


littledan (approved):

jyasskin (approved):
The canonical definition of "secure context" has moved to HTML (https://html.spec.whatwg.org/multipage/webappapis.html#secure-context), but the CR is still the explainer-level explanation, so no need to change this PR..

hayatoito (approved):
LGTM.

PR #629, opened 2021-02-25
Mentions about custom scheme handler

horo-t:


horo-t:
@jyasskin
Could you please review this PR?

explainers/subresource-loading.md
```diff
@@ -156,6 +156,10 @@ subresources, `a.js`, `b.js`, and `c.png`, from the bundle.
 `urn:uuid:f81d4fae-7dec-11d0-a765-00a0c91e6bf6` is also loaded from
 the bundle, and a subframe is instantiated as an
 [opaque origin](https://html.spec.whatwg.org/multipage/origin.html#concept-origin-opaque) frame.
+Even if a custom scheme handler for `urn` schme was registered by
+[registerProtocolHandler(scheme, url)](https://html.spec.whatwg.org/multipage/system-state.html#dom-navigator-registerprotocolhandler)
+method, the browser should ignore the registered handler and load the resource
```
hayatoito, line 161: Nit; Could we replace "should" with "must"?
(This is an explainer, so this is not an issue, however, "must" may be better for this case, IMO)

hayatoito (approved):
LGTM with a nit.

explainers/subresource-loading.md
horo-t: Done

jyasskin (commented):
This makes me really nervous. `urn:` is treated as a scheme on the web platform layer, but it actually accepts a whole universe of subschemes, each of which needs to be treated differently. IIRC we hit the same thing in CORS handling, where we'd like to define how to treat `urn:uuid:`, but we had to define a uniform behavior across all `urn:`s.

I checked in with @domenic who suggested we define a new top-level scheme for this, e.g. `packageuuid:` or `uuid-in-package:`.

horo-t:
I understand your concerns. We may need to introduce a new scheme, at least before shipping this feature, so as not to violate the existing scheme. However, since the browser changes its behavior only when the `<link rel = webbundle>` element is inserted, I think it's okay to start the Origin Trial with the current behavior. What do you think? @jyasskin @hayatoito

hayatoito:
Hmm. Let me confirm the possible behavior at first, if UA doesn't ignore a custom handler, so that I can understand the problem correctly.

If the user had visited a site at https://example.com/ that made the following call:

```
navigator.registerProtocolHandler('urn', 'soup?url=%s')
```

...and then, much later, visit https://www.example.net/, which has a ```<link rel=webundle href="..bundle" ..>``` element, and
the bundle has `urn:uuid:aaaaa` content, and the page also has `<iframe src="urn:uuid:aaaaa"></iframe>`

, then the UA might make iframe navigate to the following URL:

https://example.com/soup?url=urn:uuid:aaaaa

as if the page, https://www.example.net/, had `<iframe src="https://example.com/soup?url=urn:uuid:aaaaa"></iframe>`.

Is this an expected behavior, according to the current spec?

horo-t:
Yes. But it was unclear whether the browser should load the "https://example.com/soup?url=urn:uuid:aaaaa" from the web bundle or from the server.

Chrome before https://crrev.com/c/2717122 failed to load the "https://example.com/soup?url=urn:uuid:aaaaa" from the web bundle, and failed because the web bundle does not have such content.
After https://crrev.com/c/2717122, Chrome loads "urn:uuid:aaaaa" from the web bundle.

explainers/subresource-loading.md
```diff
@@ -156,6 +156,10 @@ subresources, `a.js`, `b.js`, and `c.png`, from the bundle.
 `urn:uuid:f81d4fae-7dec-11d0-a765-00a0c91e6bf6` is also loaded from
 the bundle, and a subframe is instantiated as an
 [opaque origin](https://html.spec.whatwg.org/multipage/origin.html#concept-origin-opaque) frame.
+Even if a custom scheme handler for `urn` schme was registered by
```
hayatoito, line 159: How about not mentioning "urn" scheme here? I think it is redundant.
Can we simply say, for example, "The browser must ignore any custom registered protocol handler in loading a resource from a bundle"?

hayatoito: I'm not sure whether it is a good idea or not to ignore a custom registered protocol handler in loading a subresource from a bundle. That needs a discussion, however, it seems much *safer* to ignore any custom registered protocol handler until we have a consensus.

horo-t: Done

explainers/subresource-loading.md
```diff
@@ -156,6 +156,8 @@ subresources, `a.js`, `b.js`, and `c.png`, from the bundle.
 `urn:uuid:f81d4fae-7dec-11d0-a765-00a0c91e6bf6` is also loaded from
 the bundle, and a subframe is instantiated as an
 [opaque origin](https://html.spec.whatwg.org/multipage/origin.html#concept-origin-opaque) frame.
+The browser must ignore any [registered custom protocol handler](https://html.spec.whatwg.org/multipage/system-state.html#dom-navigator-registerprotocolhandler) in loading a resource
+from a bundle.
```
jyasskin, line 160: This is much more comfortable; thanks @hayatoito. Custom protocol handlers are always used from https://html.spec.whatwg.org/multipage/browsing-the-web.html#process-a-navigate-url-scheme, which looks like it's only called after filtering out "fetch schemes" (like `https:`). Since a URL in a bundle is fetchable, it makes sense to avoid checking custom protocol handlers. So ...

```suggestion
If a URL is available from an attached bundle, the browser must retrieve it from the bundle, instead of using any [registered custom protocol handler](https://html.spec.whatwg.org/multipage/system-state.html#dom-navigator-registerprotocolhandler) for its scheme.
```

domenic:
I'm still a little lost.

> Is this an expected behavior, according to the current spec?

Yes.

> Yes. But it was unclear whether the browser should load the "https://example.com/soup?url=urn:uuid:aaaaa" from the web bundle or from the server.

Loading it from the web bundle seems reasonable, as long as the web bundle contains a mapping for `https://example.com/soup?url=urn:uuid:aaaaa`.

However, I'm not sure I understand the need for the patch. The patch, with or without Jeffrey's suggestion, seems to imply that registered protocol handlers get bypassed in some way. Why would we bypass them?

To implement loading `https://example.com/soup?url=urn:uuid:aaaaa` from the web bundle, I'd assume that the same fetch interception which makes any navigation to `https://example.com/soup?url=urn:uuid:aaaaa` go through a web bundle, would also work after the registerProtocolHandler. Note how [the spec](https://html.spec.whatwg.org/#dom-navigator-registerprotocolhandler) does the URL translation before ever calling navigate (and I assume it's navigate, or perhaps fetch which is called by navigate, which starts loading things from web bundles.)

jyasskin:
Re https://github.com/WICG/webpackage/pull/629#issuecomment-790254524, I think it's ok to start the origin trial using `urn:uuid:` and then change it to be more specific later. We're trying to get developer feedback on the whole thing, and that piece being wrong isn't likely to invalidate any of the feedback or break guarantees.

Re @domenic's https://github.com/WICG/webpackage/pull/629#issuecomment-790860454, since it seems like everyone's happy with moving away from `urn:uuid:` to a scheme that isn't [safelisted](https://html.spec.whatwg.org/#safelisted-scheme), we don't really need this PR. I think the PR is acceptable because the call sites for [process a navigate URL scheme](https://html.spec.whatwg.org/multipage/browsing-the-web.html#process-a-navigate-url-scheme) always prioritize fetchable URLs over registerProtocolHander-registered handlers, so prioritizing something we know how to fetch (because it's in a bundle) seems reasonable even if it has a registerable scheme. But on the other hand, the data flow look like it'll become weird, so if we don't *need* to do it, we probably shouldn't.

explainers/subresource-loading.md
horo-t: Done

hayatoito:
It seems complex than I expected. I'm not 100% confident how to interpret the current spec. However, this is an explainer, so maybe we don't need a normative one at this point.

My suggestions are:
1. Merge a PR so that we can clarify that UA must load a resource from a bundle, bypassing any custom registered protocol handler.
2. And, file a new issue to discuss moving away from "urn:uuid:".

WDYT?

explainers/subresource-loading.md
```diff
@@ -156,6 +156,9 @@ subresources, `a.js`, `b.js`, and `c.png`, from the bundle.
 `urn:uuid:f81d4fae-7dec-11d0-a765-00a0c91e6bf6` is also loaded from
 the bundle, and a subframe is instantiated as an
 [opaque origin](https://html.spec.whatwg.org/multipage/origin.html#concept-origin-opaque) frame.
+If a URL is available from an attached bundle, the browser must retrieve it from
+the bundle, instead of using any [registered custom protocolhandler](https://html.spec.whatwg.org/multipage/system-state.html#dom-navigator-registerprotocolhandler)
```
jyasskin, line 160:
```suggestion
the bundle, instead of using any [registered custom protocol handler](https://html.spec.whatwg.org/multipage/system-state.html#dom-navigator-registerprotocolhandler)
```

jyasskin (approved):
That sounds right to me.

explainers/subresource-loading.md
horo-t: Done.
Thank you.

PR #631, opened 2021-03-05
Extract function to read exchange prologue only.

twifkak:
Can be helpful for computing header-integrity, which doesn't rely on reading
the payload (because of Digest).


irori (approved):
Thanks!

PR #635, opened 2021-03-12
Add a strawperson proposal: Content Addressable Bundles

hayatoito:
This is a strawperson proposal at very early stage, aiming to load
multiple resources efficiently, using a Content-Addressable Bundle,
which can link to other bundles.


hayatoito:
@jyasskin @irori @littledan 
As I promised, I've written a strawperson proposal. Could you please quick check it?
This is just a starting point. So we don't need to discuss much in this PR, I guess.
We can discuss using our issue trackers or other places, after we merge this.

I'll continue to update, refine, add more details! This is very early draft!

irori (approved):
Merging this as-is and discuss on issue tracker sounds good to me.

hayatoito:
Thanks! Let me merge this. Let me continue to update the proposal there.
Early feedback is always welcome!

explainers/content-addressable-bundles.md
```diff
@@ -0,0 +1,429 @@
...
+
+- Loading many unbundled resources is still slower in 2021. We concluded that
+  [bundling was necessary in 2018](https://v8.dev/features/modules#bundle), and
+  our latest local measurement still suggests that.
+
+- The output of JS bundlers (e.g. webpack) doesn't interact well with the HTTP
+  cache. They are pretty good tools but configuring them to work in an optimal
+  way is tough, and sometimes they're also incompatible with new requirements
+  like
+  [dynamic bundling](https://github.com/azukaru/progressivef-fetching/blob/master/docs/dynamic-bundling/index.md)
```
horo-t, line 48: https://github.com/azukaru/progressive-fetching/blob/master/docs/dynamic-bundling/index.md

horo-t (commented):
Sorry, it is not clear for me what is the benefit of this proposal.
What is the difference between your proposal and just using two `<link rel=webbundle>`s?
Example: 
```html
<link rel=webbundle
      href=https://example.com/app/index.js.abcd.wbn
      resources="index.js foo.js bar.js" >
<link rel=webbundle
      href=https://cdn.example.com/date-util.js.f1f2.wbn
      resources="https://cdn.example.com/date-util.js" >
```

hayatoito:
> Sorry, it is not clear for me what is the benefit of this proposal.
> What is the difference between your proposal and just using two <link rel=webbundle>s?

Good questions! This can be a good candidate for FAQ. I'll add FAQ.

This proposal strongly suggests that a bundle should be **self-contained** as much as possible; Every external dependencies, including their hash values, should be explicitly declared in the bundle for reproducibility.

In the example, `https://example.com/app/index.js.abcd.wbn` depends on the specific version of `date-util.js`, which is inside of  `https://cdn.example.com/date-util.js.f1f2.wbn`. Such a dependency, including its *hash*, should be explicitly declared in the bundle itself.

If the `index.js.abcd.wbn` doesn't declare the dependency to the specific version of `date-util.js`, there is no guarantee that  `index.js` can run without any version incompatible issues. That can beak anytime.

Of course, `index.js` can hard-code the specific version of `data-util.js` such as `https://cdn.example.com/date-util.js@4.12` or something in its source file,:

```
import * as date from "https://cdn.example.com/date-util.js@4.12";
```

However, this proposal suggests that this kind of hard-coding should be avoided, and such a dependency information should be specified outside of source code.

I think this is an established pattern, and there are many similar pattern in other package management system or convention, such as node (package.json), Rust (Cargo.toml), deno (deps.ts).

This proposal also implies that only direct depencencies should be declared in a bundle (or in a page which uses `<link>`-based API). Indirect dependencies don't have to be declared.

In the example case, the main page doesn't depends on `date-util.js` directly, so the page doesn't have to declare it.

explainers/content-addressable-bundles.md
hayatoito: Thanks. Let me fix in another PR!

explainers/content-addressable-bundles.md
```diff
@@ -0,0 +1,429 @@
...
+
+### Example
+
+The page (`https://example.com/app/index.html`):
+
+<!-- prettier-ignore -->
+```html
+<link rel=webbundle
+      href=https://example.com/app/index.js.abcd.wbn
+      resources="index.js foo.js bar.js https://cdn.example.com/date.js.f1f2.wbn" >
```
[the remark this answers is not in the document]
hayatoito: Ops. Thanks! Let me fix it in another PR.

horo-t:
Sorry I don’t understand well.

> In the example case, the main page doesn't depends on date-util.js directly, so the page doesn't have to declare it.

So, the https://cdn.example.com/date-util.js.f1f2.wbn can be removed from the example’s link element’s resources attribute?
If we do so, how can the browser know that the date-util.js can be loaded from the date-util.js.f1f2.wbn?

hayatoito:
> So, the https://cdn.example.com/date-util.js.f1f2.wbn can be removed from the example’s link element’s resources attribute?

That makes sense.  The page can declare only `index.js` in resources in the example case.

e.g.
```html
<link rel=webbundle
      href=https://example.com/app/index.js.abcd.wbn
      resources="index.js" >
```

Let me fix this. Thanks!

hayatoito:
> If we do so, how can the browser know that the date-util.js can be loaded from the date-util.js.f1f2.wbn?

The browser should do a similar step as 6 between 2 and 3 (in *How to Load* section). Let me fix that!

horo-t:
If my understanding is correct, the browser must defer all subresource requests after starting the web bundle fetch until the browser finishes parsing the web bundle.  Is my understanding correct?

hayatoito:
> If my understanding is correct, the browser must defer all subresource requests after starting the web bundle fetch until the browser finishes parsing the web bundle. Is my understanding correct?

A good question. This proposal doesn't have an intention to defer all subresource requests. The browser must defer only subresources requests for resources which are explicitly specified resources attribute (or resources listed in the index section of a bundle).

For example,  if the top page uses `foo.js` directly, without any forward declaration, such as:

`<script type="module" src="foo.js" />`.

That should be probably considered as a configuration error. There is no guarantee from where `foo.js` is loaded because of race condition, as of now. The proposal doesn't define any formal semantics in this case yet.

PR #636, opened 2021-03-16
Update content-addressable-bundles.md

hayatoito:
- Address the comments of PR #625, fixing editorial errors
- Introduce terminorogies: mutable bundles and immutable bundles
- Add some FAQs


hayatoito:
@irori I've updated the explainer a bit. Could you please do quick check and merge this?

cc: @littledan @jyasskin 

Let me work on updating this explainer this week. There are still many TODO items.

irori (approved):

irori:
Merged.

littledan:
I'm a bit concerned that this protocol has too many round-trips in it. The nested bundles are discovered one by one, down the tree. I think it's important to load all bundle chunks ASAP, without multiple round trips, to avoid being a regression vs current loading techniques. We could do this by making a manifest of the bundle chunk tree structure available inline in the HTML on initial load, so the browser can immediately calculate what should be fetched. I previously wrote up a somewhat different version of this idea at https://gist.github.com/littledan/18a1bd6e14e4f0ddb305a2a051ced01e (this gist preserves the idea that the integrity hash and chunk ID are separated, but they could be unified as in your document).

hayatoito:
Thanks for the comment!

A good question! A possible workaround as a user-land solution would be to declare indirect dependent bundles as well as direct dependencies.

In the example case, top page may declare:

```html
<link rel=webbundle
      href=https://example.com/app/index.js.abcd.wbn
      resources="index.js" >
<link rel=webundle href=https://cdn.example.com/date-util.js.f1f2.wbn>
<link rel=webundle href=https://cdn.example.com/common-util.js.3b3b.wbn>
```
Then, this should guarantee that a browser prefetch every bundles in parallel, instead of on demand.

The 2nd and 3rd link elements are optional. They are declared to avoid a round trip.

I'm about to add this to FAQ. Let me reflect this. Thanks!

hayatoito:
> https://gist.github.com/littledan/18a1bd6e14e4f0ddb305a2a051ced01e

Thanks for the link! I am afraid I didn't read this.  Let me read!
Yes, Merkle tree inspired me, and this link also mentioned Merkle tree. It looks I can learn many things from there.

littledan:
I like the idea of making the bundle itself self-contained and having these redundant inline elements as optional to avoid extra round-trips. There are some more fundamental reasons why dynamic bundling could be a win (which Yoav and Shubhie have written about before) but this is a pretty tricky topic and I am working on consolidating that information in one place to be easier to understand.

Issue #638, opened 2021-03-17
Content Addressable Bundles


hayatoito:
Let me file an issue to discuss a proposal: [Resource Loading with Content Addressable Bundles](https://github.com/WICG/webpackage/blob/main/explainers/content-addressable-bundles.md).

This is a strawperson proposal at very early stage, and this can be positioned as "my personal proposal", as of now.

I've put it in `explainer` directory, however, if that is inappropriate and confusing, I'll move it to somewhere. Maybe we can have `proposals` directory, as [WICG/webcomponenets has](https://github.com/WICG/webcomponents/tree/gh-pages/proposals).  I am not sure what is the best practice in this repository. 

I hope we can use this issue to discuss and get early feedback.
I've observed that PRs are not a good place to discuss overall design and ideas because we are likely to lose the discussion history. An issue might be a better place to discuss.

cc: @littledan @jyasskin @yoavweiss

horo-t:
Thank you for filing this issue. As I mentioned at today’s meeting, the goal of this proposal is still unclear for me.

> - This proposal aims to support Code Splitting, as webpack or other bundlers
   already support as a user-land solution. Smaller bundles, if used correctly,
   can have a major impact on load time.

As written in this page (https://webpack.js.org/guides/code-splitting/), the purpose of Code Splitting is to “[have a major impact on load time](https://webpack.js.org/guides/code-splitting/#:~:text=have%20a%20major%20impact%20on%20load%20time)”. It would be helpful if you can write more precisely how this proposal can support this purpose, and why it is better than just using the existing proposal "[Subresource loading with Web Bundles](https://github.com/WICG/webpackage/blob/main/explainers/subresource-loading.md)".


> - Non-opinionated about bundle granularity. There are trade-offs how a site
  composes their resources into bundles in order to balance various factors like
  total bytes transferred, loading latency, or cache granularity. Instead of a
  _all-or-nothing_ bundle, this proposal aims to provide a way to express a
  dependency graph of bundles. The use of bundlers is an established practice in
  Web development. Bundlers, such as webpack, skypack, would know much about
  which resources should be grouped as a bundle, and might want to express their
  intent as a dependency graph of bundles, considering various trade-offs. They
  wouldn't want to lose this information in building an all-or-nothing bundle,
  and a browser wants to know it to improve a loading performance.

I understand the importance of carefully maintaining the dependency graph of modules while developing.
[...]

PR #639, opened 2021-03-17
Update content-addressable-bundles.md

hayatoito:
- Add a link to issue #638
- Add "URL integrity" to Goal
- Add "Resource Batch Preloading" to references


hayatoito:
I've added a link to issue #638 so that a reader can know the position of this proposal, and find easily where they can give feedback.

hayatoito:
There are several commits in this PR. I'll do "squash and merge" so that this can be one commit.

explainers/content-addressable-bundles.md
```diff
@@ -425,9 +450,13 @@ formal procedure, however, there are several possible approaches:
   space-consuming.
 
   For example, given that browser has the following dependency graph of bundles,
-  where the bodies of `bundle-C`, `bundle-D`, `bundle-E` are missing, the
-  browser can start to fetch them again, possibly in parallel, without waiting
-  fetching `bundle-C`.
+  where the bodies of `bundle-C`, `bundle-D`, `bundle-E` are omitted to save the
+  cache disk space, the browser can start to fetch them again, possibly in
+  parallel. Note that a browser can start to fetch `bundle-D` and `bundle-E`
+  immediately without waiting fetching `bundle-C` (and scanning its index
```
horo-t, line 456: My understanding is:
 - Web developers should write multiple webbundle link elements in the HTML to fetch the webbundles in parallel if the loading performance is important.
 - This nested webbundle format should be used when lazy-loading is important, and the loading performance is not important.
 
Is my understanding correct?

If so, I don't understand the motivation of browser developers of introducing a new cache mechanism to keep the dependency information. Could you please explain this?

hayatoito: A good question!

It's fine for a browser to prefetch nested bundles even if a user doesn't request it explicitly.

- If a user requests prefetch via `<link>` elements explicitly,  browser *should* prefetch it.
- If a user doesn't request it via `<link>` elements, a browser doesn't have to prefetch it, but *may* prefetch it, enabling the content to load instantly if and when the user requests it.

hayatoito:
Let  me close this PR since I don't have a plan to work on this soon.

Issue #630, opened 2021-03-04
Subresource loading: Add a section to explain how resource-bundles are related.


hayatoito:
We should spell out how "Subresource loading with Web Bundles" are related to [WICG/resource-bundles](https://github.com/WICG/resource-bundles/) in its [explainer](https://github.com/WICG/webpackage/blob/main/explainers/subresource-loading.md).

The context: https://github.com/w3ctag/design-reviews/issues/616#issuecomment-790395734

jyasskin:
We should also update the [alternate designs](https://github.com/WICG/webpackage/blob/main/explainers/subresource-loading.md#alternate-designs) with some of the ideas from [resource-bundles](https://github.com/WICG/resource-bundles/blob/main/subresource-loading.md#example) and [MT's post](https://lowentropy.net/posts/bundles/#indirection-for-identifiers)

hayatoito:
A good point. Some of alternative designs have not been updated, and I'm feeling a bit outdated.
We should collect alternative designs and summarize them in the explainer.

hayatoito:
[Issue Triage]
Let me close this tentatively.

PR #640, opened 2021-03-23
Update subresource-loading.md #630

hayatoito:
Clean up the Alternate designs section:

- Remove Approximate Membership Query datastructure, which is
  a bit obsolete. Resource Bundles would be a successor.
- Some editorial fixes.


jyasskin:
We should keep the list of alternate designs in the explainer, as shown in https://w3ctag.github.io/explainers#considered-alternatives. When we decide to definitely not pursue a particular design, we should add the reason, rather than removing the design, so that future readers know whether to reopen that discussion.

hayatoito:
Closing.

PR #661, opened 2021-08-03
Switch from Travis-CI to GitHub workflows

### LINKED ISSUE #660: CI is broken
https://github.com/WICG/webpackage/pull/659 is blocked by CI not working.

https://travis-ci.org/github/WICG/webpackage/builds/ says:

> Since June 15th, 2021, the building on travis-ci.org is ceased. Please use travis-ci.com from now on.

I think there are two options:

- [Migrate](https://docs.travis-ci.com/user/migrate/open-source-repository-migration) to travis-ci.com. Probably it needs repository admin's power (@jyasskin)
- Switch from Travis-CI to GitHub workflows. I see some other WICG repositories recently switched to GH workflows ([link](https://github.com/search?o=desc&q=org%3AWICG+travis&s=committer-date&type=Commits)), though I'm not sure whether there's strong reason to do that.

hayatoito:
Fix #660. This is a straightforward migration from Travis-CI to GitHub
workflows.

It seems we could have more sophisticated workflow files, such as [1],
however, this is a good starter to switch to GitHub workflows.

[1] https://github.com/wpack-wg/bundled-responses/tree/main/.github/workflows

Disclaimer: publish.yml is not tested yet because it actually uploads
changes to IETF datatracker.


hayatoito:
I've roughly migrated Travis-CI (`.travis.yml`) to GitHub workflows (`.github/workflows/*.yml`).

@jyasskin 
Do you think it is okay that we switch from Travis-CI to GitHub workflows?

If it is okay:

I've removed `.travis.yml` in this PR, however, travis-ci is still running for this PR.
Is this an expected behavior? I'm not familiar with Travis-CI. If we need a further action to switch, I appreciate your advice!

.github/workflows/build.yml
```diff
@@ -0,0 +1,22 @@
...
+          go-version: '^1.13.1'
+      - run: go test -v ./go/...
+
+  javascript:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node-version: 15.x
```
jyasskin, line 19: Can we use a maintained version here?

```suggestion
          node-version: 16
```

or 

```suggestion
          node-version: lts/*
```

.github/workflows/publish.yml
```diff
@@ -0,0 +1,31 @@
...
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+        with:
+          python-version: 3.9
+      - uses: ruby/setup-ruby@477b21f02be01bcb8030d50f37cfec92bfa615b6
+        with:
+          ruby-version: 2.6
+      - name: Install dependencies
+        run: |
+          pip install bikeshed xml2rfc python-dateutil
```
jyasskin, line 19: I think we can leave Bikeshed out of the publish.yml flow, and instead copy https://github.com/martinthomson/i-d-template/blob/master/template/.github/workflows/publish.yml over literally.

.github/workflows/spec.yml
```diff
@@ -0,0 +1,31 @@
...
+      - name: Install dependencies
+        run: |
+          pip install bikeshed xml2rfc python-dateutil
+          bikeshed update
+          if head -1 -q *.md | grep '^\-\-\-' >/dev/null 2>&1; then gem install --no-doc kramdown-rfc2629; fi
+          if head -1 -q *.md | grep '^%%%' >/dev/null 2>&1; then go get "$mmark_src" && go build "$mmark_src"; fi
+      - name: Make github pages
+        run: |
+          bikeshed --die-on=warning spec loading.bs
+          make
```
jyasskin, line 29: It'll probably be more straightforward if we split the Bikeshed build and gh-pages upload from the kramdown-rfc2629 one, into separate files. The kramdown-rfc2629 build can use https://github.com/martinthomson/i-d-template/blob/master/template/.github/workflows/ghpages.yml, and the Bikeshed build should use https://w3c.github.io/spec-prod/#deploy-to-github-pages.

`--die-on=warning` becomes https://w3c.github.io/spec-prod/#build_fail_on.

.github/workflows/build.yml
```diff
@@ -0,0 +1,22 @@
+name: build
+on: [push, pull_request]
+jobs:
+  go:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-go@v2
+        with:
+          go-version: '^1.13.1'
```
jyasskin, line 10: Is this still the right Go version? It looks like they [released 1.16 in 2021-02](https://golang.org/project).

hayatoito: Thanks! Fixed.
It seems we have to specify a specific version.
We can't specify "latest" or similar general term for Go.

I referred to https://github.com/actions/go-versions/blob/main/versions-manifest.json

.github/workflows/build.yml
hayatoito: Thanks! Fixed. Let's use 'lts/*'
I referred to https://github.com/actions/setup-node

.github/workflows/publish.yml
hayatoito: Thanks. Fixed.
I literally copied the template.

.github/workflows/spec.yml
hayatoito: Thanks. I've split this file into ghpages.yml and spec-prod.yml.

jyasskin (approved):
Great! The spec-prod error is probably real; could you fix it in a separate PR? I'll make you an admin so that you can iterate on this if it breaks.

hayatoito:
Thanks for the review! 

> Great! The spec-prod error is probably real; could you fix it in a separate PR? I'll make you an admin so that you can iterate on this if it breaks.

Sure. It seems ReadableStream is not defined. Let me try to fix in a separate PR.

PR #667, opened 2021-08-25
Revive the signatures section as an extension

horo-t:
This PR revives the signatures section as an extension which was removed from the Web Bundles spec by https://github.com/WICG/webpackage/pull/604/files. 

This extension document doesn't follow the latest changes on the Web Bundles spec now. The content negotiation part was removed by https://github.com/wpack-wg/bundled-responses/pull/7/. So the `variants-value` in the CDDL is nonsense now.

But for the records, this PR just revives the deleted part with a note.

### DIFF
--- a/extensions/signatures-section.md
+++ b/extensions/signatures-section.md
@@ -1,0 +1,83 @@
+# Signatures optional section
+
+This document outlines the format and usage of the "signatures" optional section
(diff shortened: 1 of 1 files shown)


hayatoito (approved):

hayatoito:
I think we can merge this now. We can update this anytime later!

irori (commented):
lgtm

PR #672, opened 2021-09-06
Remove content negotiation support from 'b2' bundle version

myrzakereyms:
As 'content negotiation' is removed from the spec in [this PR](https://github.com/wpack-wg/bundled-responses/pull/7), we need to update the Go encoder/decoder as well. Since the existing bundles of 'b1' version still use content negotiation, we only remove it from the 'b2' version (and newer). 

Issue: https://github.com/wpack-wg/bundled-responses/issues/2

Because the support is partially still there (until 'b1' bundles get phased out), the issue won't be closed just yet.

### DIFF
--- a/go/bundle/decoder.go
+++ b/go/bundle/decoder.go
@@ -159,122 +159,64 @@ func decodeCborHeaders(dec *cbor.Decoder) (http.Header, map[string]string, error
 
 // https://wicg.github.io/webpackage/draft-yasskin-dispatch-bundled-exchanges.html#index-section
 // "To parse the index section, given its sectionContents, the sectionsStart offset, the sectionOffsets CBOR item, and the metadata map to fill in, the parser MUST do the following:" [spec text]
-func parseIndexSection(sectionContents []byte, sectionsStart uint64, sos []sectionOffset, bs []byte) ([]requestEntryWithOffset, error) {
-	// Step 1. "Let index be the result of parsing sectionContents as a CBOR item matching the index rule in the above CDDL (Section 3.4). If index is an error, return nil, an error." [spec text]
-	idxdec := cbor.NewDecoder(bytes.NewBuffer(sectionContents))
-	nidx, err := idxdec.DecodeArrayHeader()
+func parseIndexSection(sectionContents []byte, sectionsStart uint64, sos []sectionOffset) ([]requestEntryWithOffset, error) {
+	// Step 1. "Let index be the result of parsing sectionContents as a CBOR item matching the index rule in the above CDDL (Section 3.5). If index is an error, return an error." [spec text]
+	dec := cbor.NewDecoder(bytes.NewBuffer(sectionContents))
+	numUrls, err := dec.DecodeMapHeader()
 	if err != nil {
-		return nil, fmt.Errorf("bundle.index: Failed to decode \"index\" array header: %v", err)
+		return nil, fmt.Errorf("bundle.index: failed to decode index section map header: %v", err)
 	}
+	// Step 2. "Let requests be an initially-empty map ([INFRA]) from URLs to response descriptions, each of which is either a single location-in-stream value" [spec text]
+	requests := []requestEntryWithOffset{}
 
-	// Step 2. "Check that the responses array has the right number of items:" [spec text]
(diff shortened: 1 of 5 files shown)


go/bundle/encoder.go
```diff
@@ -95,9 +95,12 @@ type indexSection struct {
 	bytes []byte
 }
 
-func (is *indexSection) addExchange(e *Exchange, offset, length int) error {
+func (is *indexSection) addExchange(e *Exchange, offset, length int, ver version.Version) error {
 	variants := normalizeHeaderValues(e.Response.Header[http.CanonicalHeaderKey("variants")])
 	variantKey := normalizeHeaderValues(e.Response.Header[http.CanonicalHeaderKey("variant-key")])
+	if (!ver.SupportsVariants() && (variants != "" || variantKey != "")) {
+		panic("Variants included in the exchange, but the specific bundle version '" + ver + "' does not support it.")
```
irori, line 102: I think `b2` bundle should be "transparent" regarding the `variants` / `variant-key` headers, i.e. these headers are just stored in the bundle without any special treatment.

irori:
Wouldn't this break `Unversioned` format support, where the index section was a CBOR array?

I think it's OK to drop the `Unversioned` format support now, but please create a separate PR if you want to do so.

irori:
Please rebase this branch and resolve the conflicts.

myrzakereyms:
Done

go/bundle/encoder.go
```diff
@@ -165,6 +165,36 @@ func (is *indexSection) Finalize(ver version.Version) error {
...
+		// whatwg-url = tstr
+		// location-in-responses = (offset: uint, length: uint)
+		m := make(map[string][]*indexEntry)
+		for _, e := range is.es {
+			url := e.URL.String()
+			m[url] = append(m[url], e)
+		}
+
+		mes := []*cbor.MapEntryEncoder{}
+		for url, es := range m {
```
irori, line 180: What happens if there are multiple indexEntries for a single URL?

myrzakereyms: We will have a slice of indexEntries per url and (as far as I know) we just pick the first one when finding a matching response to a url.

go/bundle/encoder.go
```diff
@@ -165,6 +165,36 @@ func (is *indexSection) Finalize(ver version.Version) error {
...
+				if err := keyE.EncodeTextString(url); err != nil {
+					panic(err)
+				}
+
+				if err := valueE.EncodeUint(es[0].Offset); err != nil {
+					panic(err)
+				}
+				if err := valueE.EncodeUint(es[0].Length); err != nil {
+					panic(err)
+				}
```
irori, line 191: This generates an invalid CBOR structure, because value of a map is not a single CBOR item.

According to the CDDL:
```
index = {* whatwg-url => [ location-in-responses ] }
location-in-responses = (offset: uint, length: uint)
```
A value of the `index` map is an array of two `uint`s. So you should add
```
valueE.EncodeArrayHeader(2)
```
before encoding the `uint`s.

irori (commented):
Oh, I meant `[spec text]`s *added in this PR* should be removed. I'm fine with removing all of them in this CL, but feel free to move it in a separate PR.

go/bundle/encoder.go
myrzakereyms: You're right, fixed.

go/bundle/decoder.go
```diff
@@ -202,7 +178,67 @@ func parseIndexSection(sectionContents []byte, sectionsStart uint64, sos []secti
 			return nil, fmt.Errorf("bundle.index[%d]: URL contains credentials: %q", i, rawUrl)
 		}
 
-		// Step 4.3. "If the first element of responses is the empty string:" [spec text]
+		numItems, err := dec.DecodeArrayHeader()
+		if err != nil {
+			return nil, fmt.Errorf("bundle.index[%d]: Failed to decode value array header: %v", i, err)
+		}
+		if numItems == 0 {
```
irori, line 185: I think `numItems` must be exactly 2.

irori (approved):
Thanks, LGTM

PR #674, opened 2021-09-07
Remove formal support for Unversioned bundles.

### LINKED ISSUE #673: Remove the support for Unversioned bundles
'Unversioned' bundles are the bundles that use an old format that is not present in the [latest spec](https://github.com/wpack-wg/bundled-responses/blob/main/draft-ietf-wpack-bundled-responses.md)
No one uses these anymore, let's remove the support for it.

myrzakereyms:
Closes #673


go/bundle/version/version.go
```diff
@@ -120,9 +108,9 @@ func (v Version) HasPrimaryURLFieldInHeader() bool {
 }
 
 func (v Version) SupportsVariants() bool {
-	return v != Unversioned
+	return true
```
irori, line 111: `return v == VersionB1`?

irori (commented):
Thanks, looking good

irori (approved):
Thanks, LGTM

PR #677, opened 2021-09-09
Use uuid-in-package: scheme for opaque origin

### LINKED ISSUE #650: Consider using a more specific scheme for opaque origin
Regarding [Support opaque origin iframes](https://github.com/WICG/webpackage/blob/main/explainers/subresource-loading-opaque-origin-iframes.md), we might want to use a new scheme, instead of `urn:uuid`.

@jyasskin I remember you mentioned that we should introduce a new scheme, but I couldn't find. It would be appreciated if you can share the context!

Note this is tracked as a  v1 issue in #648.

irori:
Closes #650.

The `urn:` scheme is included in the [safelisted schemes](https://html.spec.whatwg.org/multipage/system-state.html#safelisted-scheme) of the HTML spec, meaning that web sites can [register a custom protocol handler](https://html.spec.whatwg.org/multipage/system-state.html#custom-handlers) that handles `urn:` scheme.

To avoid the potential for conflict, this introduces a new URL scheme `uuid-in-package:` that can be used in resource URLs in web bundles.

### DIFF
--- a/explainers/subresource-loading-opaque-origin-iframes.md
+++ b/explainers/subresource-loading-opaque-origin-iframes.md
@@ -31,11 +31,6 @@ In addition to the same origin subresource explained in the
 section in the explainer, this extension allows a bundle to include a
 `uuid-in-package:` URL subresource.
 
-Note: Previous version of this document used
-[`urn:uuid:` URLs](https://datatracker.ietf.org/doc/html/rfc4122) for this
-perpose. Chromium's experimental implementation currently supports only
-`urn:uuid:` (as of M95).
-
 ### Opaque origin iframes
 
 If a `<iframe>`'s `src` attribute is a `uuid-in-package:` URL subresource in the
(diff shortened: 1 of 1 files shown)


horo-t:
I think we should have some note about that currently Chromium is supporting `urn:uuid:`.
And after we land changes in Chromium, updates the explainer like "Chrome has supported `urn:uuid:` until M9x".

irori:
> I think we should have some note about that currently Chromium is supporting `urn:uuid:`.

Done.

explainers/subresource-loading-opaque-origin-iframes.md
```diff
@@ -16,16 +16,29 @@ Support the use case of
...
+Where `<UUID>` is a UUID as specified in
+[RFC 4122](https://datatracker.ietf.org/doc/html/rfc4122).
 
 In addition to the same origin subresource explained in the
 [`<link>`-based API](https://github.com/WICG/webpackage/blob/main/explainers/subresource-loading.md#link-based-api)
 section in the explainer, this extension allows a bundle to include a
-[`urn:uuid:`](https://tools.ietf.org/html/rfc4122) URL subresource.
+`uuid-in-package:` URL subresource.
+
+Note: Previous version of this document used
```
hayatoito, line 34: Nit; Can we have "Alternative Considered" section and move these into the section?

We might want to mention the reason why we no longer consider "urn:uuid", as this PR's comment explains.

hayatoito (changes requested):
Thanks! LGTM. Can we have "Alternatives Considered" section and mention "urn:uuid" there?

explainers/subresource-loading-opaque-origin-iframes.md
irori: Done.

hayatoito (approved):
Thanks!

horo-t (approved):
lgtm

Issue #647, opened 2021-04-08
Support a relative URL of a resource in a bundle.


hayatoito:
It might be worth considering to support a relative URL in the bundle format.

I've heard this feature request from several people. I've searched an existing issue, but I couldn't find one so let me file an issue.

Use cases:

Regarding [subresource loading](https://github.com/WICG/webpackage/blob/main/explainers/subresource-loading.md) use case, we use the following syntax in the main doc:

```html
<link rel=webundle href=https://example.com/app/bundle.wbn resources=".....">
```

Because we have *path restriction*, the resources in the bundle *must* start with `https://example.com/app/`, such as:

- https://example.com/app/top.js
- https://example.com/app/util1.js
- https://example.com/app/util2.js
- ...

It seems a bit redundant.  If we support a relative URL, each subresource doesn't have to include a redundant common prefix, and can be:

- ./top.js
- ./util1.js
- ./util2.js
- ...

We have to decide on which `base` a relative URL should be resolved. For subresource loading use cases, it's clear. 
For other use cases, TBD. 

The other benefits are:

- A bundle can be more portable. For example, we can move `bundle.wbn` from https://example.com/app/ to https://example.com/other/, freely, without re-creating a bundle.
- The size of a bundle can be reduced.

hayatoito:
https://github.com/WICG/webpackage/pull/617 can be related. We need to make primary URL optional too.

jyasskin:
Yep, I think we do this by either
1. putting the base URL in a section (possibly one holding a map of several metadata values), or
2. using the bundle's URL as the base.

horo-t:
Hi.

I created a [PR](https://github.com/WICG/webpackage/pull/678) to support the second idea "using the bundle's URL as the base." in @jyasskin 's [comment](https://github.com/WICG/webpackage/issues/647#issuecomment-816005302) in subresource bundle.

I think we can introduce a `base` section which can override the base URL in future. But even so, I think it is reasonable to use the bundle's URL as the base URL which can be overridden by the `base` section.

hayatoito:
Thanks!
Yes, I think we can agree on the "using the bundle's URL as the base".
While I can appreciate the PR, we probably wan to wait for a resolution of https://github.com/wpack-wg/bundled-responses/issues/14.

hayatoito:
Fixed. Now we support a relative URL.

PR #678, opened 2021-09-10
Support relative URLs in bundle in subresource loading

horo-t:
This is intended to support the second idea "using the bundle's URL as the base." in @jyasskin 's [comment](https://github.com/WICG/webpackage/issues/647#issuecomment-816005302) in subresource bundle.

### DIFF
--- a/explainers/subresource-loading.md
+++ b/explainers/subresource-loading.md
@@ -134,6 +134,11 @@ Suppose that the bundle, `subresources.wbn`, includes the following resources:
 - … (omitted)
 ```
 
+A URL of the resource in the bundle can be a [relative
+URL](https://url.spec.whatwg.org/#syntax-url-relative) to the bundle.
+A browser must [parse a URL](https://html.spec.whatwg.org/#parse-a-url)
+using bundle's URL.
+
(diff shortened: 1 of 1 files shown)


horo-t:
@hayatoito @jyasskin What do you think about this change?
I merged https://github.com/wpack-wg/bundled-responses/pull/15 into the web bundle spec. So this change may not be necessary. But I think it is kind to mention relative URLs in this explainer for web developers.

hayatoito:
Thanks. I've checked https://github.com/wpack-wg/bundled-responses/pull/15. It says:

> In the latter case, the referenced URL is relative to the bundle's URL. 

It seems the format doesn't define what is `the bundle's URL` strictly. I think the definition of `the bundle's URL` depends on each use case.

Given that, it would be helpful to mention how a relative URL is resolved on which URL.
In subresource loading cases, it's obvious, but it would be helpful to clarify it in the explainer so that developers can notice a relative URL can be used.

horo-t:
Thanks. If this PR looks good to you, could you please approve this?

hayatoito (approved):

PR #684, opened 2021-09-28
Document how CORS and CORP work for bundled resources

### LINKED ISSUE #644: Define how CORP and COEP should work for subresource loading
- CORP: https://developer.mozilla.org/en-US/docs/Web/HTTP/Cross-Origin_Resource_Policy_(CORP)
- COEP: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cross-Origin-Embedder-Policy

We need to figure out how CORP and COEP should work for Subresource loading with WebBundles.

Chromium side issue is [here](https://bugs.chromium.org/p/chromium/issues/detail?id=1163788).

irori:
Closes #644, and also describes how CORS works for requests to bundled resurces (rather than requests to bundles).

This does not describe COEP specifically, because it is related to WebBundles only through CORP.

### DIFF
--- a/explainers/subresource-loading.md
+++ b/explainers/subresource-loading.md
@@ -16,6 +16,7 @@ a format that allows multiple resources to be bundled, e.g.
   - [The main document](#the-main-document)
 - [Request's mode and credentials mode](#requests-mode-and-credentials-mode)
 - [Request's destination](#requests-destination)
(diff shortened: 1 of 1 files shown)


hayatoito (approved):
Thanks for updating this. That makes sense to me. I think we can merge this now.

PR #690, opened 2021-10-11
Remove manifest section from a b2 bundle

myrzakereyms:
This PR just makes sure there's no manifest section when creating B2 WebBundles. The support still remains for B1 bundles.
Closes #690


irori (approved):
Thanks!

PR #693, opened 2021-10-20
js/bundle: Support Primary URL optional section in b2

irori:
This adds support for the "primary" section, defined in this extension:
https://github.com/WICG/webpackage/blob/main/extensions/primary-section.md

Now primary URL is set by BundleBuilder.setPrimaryURL(), rather than a constructor argument.


irori:
@hayatoito could you take a look?

PR #694, opened 2021-10-20
Support credentials mode in <script>-based API

### LINKED ISSUE #692: Support credential mode in <script>-based API
The current `<script>`-based API doesn't allow to specify a [credential mode](https://fetch.spec.whatwg.org/#concept-request-credentials-mode), while `<link>`-based APIs allowed that by its `crossorigin` attribute.

We want to support it by having `credentials` key.  For example,

```
<script type="webbundle">
{
  source: "subresources.wbn",
  credentials: "omit"
  resources: ["a.js", "b.js", "c.png"],
}
</script>
```

"omit", "same-origin" and "include" can be specified.
Their meanings are defined in the fetch spec.

https://fetch.spec.whatwg.org/#concept-request-credentials-mode
https://fetch.spec.whatwg.org/#dom-requestinit-credentials

hayatoito:
Fix #692.

### DIFF
--- a/explainers/subresource-loading.md
+++ b/explainers/subresource-loading.md
@@ -14,7 +14,7 @@ a format that allows multiple resources to be bundled, e.g.
 - [Example](#example)
   - [The bundle](#the-bundle)
   - [The main document](#the-main-document)
-- [Request's credentials mode](#requests-credentials-mode)
+- [Request's mode and credentials mode](#requests-mode-and-credentials-mode)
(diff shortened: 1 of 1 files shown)


hayatoito:
Good points! It seems I had removed a mention of a request mode. That was my mistake.  That is always cors. Fixed.

I've clarified that an element's `crossorigin` attribute is not used.

cc: @hiroshige-g

PR #703, opened 2021-11-19
dump-certurl: Print SHA-256 hash of SubjectPublicKeyInfo

irori:
This information can be used as the value of `--ignore-certificate-errors-spki-list=` command line flag of Chrome when using a test certificate.


hayatoito (approved):
Rubber stamp LGTM

PR #706, opened 2021-12-16
Add "Bundle Dependencies" section proposal

hayatoito:
Let's merge a "Bundle Dependencies" proposal so that we can develop a proposal in this repository.

Note that the proposal is still a draft. I'm going to file a separate issue so that we can discuss more easily, hoping we can get feedback from various people who might have an interest.


irori:
Can we put this under `extensions/proposals/` directory, to make it clear that this is at very early stage and we are still exploring possible design spaces? (See Jeffrey's suggestion about `content-addressable-bundles.md` in https://github.com/WICG/webpackage/issues/638#issuecomment-806196738)

hayatoito:
Sure. I've moved a proposal to `extensions/proposals`.

irori (approved):

PR #721, opened 2022-03-27
Specify the HTML spec integration

hiroshige-g:
- Fill the HTML monkeypatches section.
- Introduce `web bundle fetch entry` that tracks fetching and reusing fetched bundles. `source` and `credentials` are moved from `bundle rule`, because they are associated with fetched bundles rather than rules.
- Rename `web bundle` to `web bundle registration`, to distinguish it more clearly from `web bundle fetch entry`.
- Adjust existing parts of the spec accordingly.

***
<a href="https://pr-preview.s3.amazonaws.com/hiroshige-g/webpackage/pull/721.html" title="Last updated on Apr 6, 2022, 9:44 PM UTC (a24a068)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/webpackage/721/dffb601...hiroshige-g:a24a068.html" title="Last updated on Apr 6, 2022, 9:44 PM UTC (a24a068)">Diff</a>

### DIFF
--- a/subresource-loading.bs
+++ b/subresource-loading.bs
@@ -7,6 +7,7 @@ Group: WICG
 Repository: WICG/webpackage
 URL: https://wicg.github.io/webpackage/subresource-loading.html
 Editor: Hayato Ito, Google Inc. https://google.com/, hayato@google.com
+Editor: Hiroshige Hayashizaki, Google https://google.com/, hiroshige@google.com
 Abstract: How UAs load subresources from Web Bundles.
 Complain About: accidental-2119 yes, missing-example-ids yes
 Indent: 2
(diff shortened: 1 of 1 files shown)


subresource-loading.bs
```diff
@@ -55,41 +55,291 @@ Note: This specification is under construction. See
...
+1. [=list/Remove=] |registration| from |document|'s [=Document/web bundle
+   registration list=].
+1. [=Queue a microtask=] to perform the following steps:
+
+   1. Let |fetch entry| be |registration|'s [=web bundle registration/fetch
+      entry=].
+   1. If |fetch entry| is [=web bundle fetch entry/used by a registration=] in
+      |document|, then return.
+
+   1. Note: It is possible that |document|'s [=Document/web bundle fetch entry
```
irori, line 331: Does this step only have a note?

subresource-loading.bs
```diff
@@ -55,41 +55,291 @@ Note: This specification is under construction. See
...
+         result=]'s [=web bundle result/registration=]'s [=web bundle
+         registration/fetch entry=] is |r| was removed. This is to ensure that
+         [=web bundle fetch entries=] are not destructed and re-fetched when
+         {{HTMLScriptElement}}s with the same web bundle [=web bundle fetch
+         entry/source=] and [=web bundle fetch entry/credentials=] are removed
+         and added.
+
+      1. Otherwise, then a user agent MAY set |fetch entry| to |r|.
+
+         ISSUE: Should this be "MUST set", "MAY set", or "MUST NOT set"?
```
hayatoito, line 215: Let's omit this "Otherwise, then a user agent..." clause (That's effectively equivalent with "MUST NOT set").
As you noticed, Chromium doesn't do this optimization for this minor case, where I don't think there is a use case.

Let's keep the algorithm simple at the moment unless an optimization is needed in a real use case.

hayatoito (commented):
Thanks for writing this! Overall, looks great!

subresource-loading.bs
hiroshige-g: Sounds good. Removed.

subresource-loading.bs
hiroshige-g: Moved this note to the following step.

hiroshige-g:
Rebased & reflected comments.

@domenic just FYI for scripting part. While doing some new things (e.g. adding a "removed from the document" hook to script elements etc.), I expect this PR doesn't affect the semantics/mechanisms of scripts elements, as this PR doesn't process updated `textContent` after prepare-a-script.

hayatoito (approved):
Thanks for rebasing and cleaning up.
The PR captures our intention well! (eg. Reusing webbundle resources in "remove, then add" case).

For the detail of script parts, especially consistency with other kinds of <script> elements, I would defer to @domenic.

subresource-loading.bs
```diff
@@ -59,41 +60,286 @@ Note: This specification is under construction. See
...
+
+   1. Let |fetch entry| be |registration|'s [=web bundle registration/fetch
+      entry=].
+   1. If |fetch entry| is [=web bundle fetch entry/used by a registration=] in
+      |document|, then return.
+
+   1. [=list/Remove=] |fetch entry| from |document|'s [=Document/web bundle
+      fetch entry list=].
+
+      Note: It is possible that |document|'s [=Document/web bundle fetch entry
```
irori, line 334: Is this still possible? Since the step 5.1.2 of **prepare a web bundle** was removed as per [Hayato's comment](https://github.com/WICG/webpackage/pull/721#discussion_r838384360), a fetch entry cannot be shared by multiple registrations in a document.

hiroshige-g: Yes, possible. Consider the script below:

```
const script1 = document.createElement('script');
// Set type=webbundle and textContent.
const script2 = document.createElement('script');
// Set type=webbundle and textContent with the same source URL.

// A fetch entry is added to the web bundle fetch entry list,
document.head.appendChild(script1);
script1.remove();
// The fetch entry is reused here.
document.head.appendChild(script2);
script2.remove();
```

Here we have two pending microtasks that attempt to remove the fetch entry from the web bundle fetch entry list.
The first microtask will actually remove the fetch entry from the list, and when the second microtask is executed the fetch entry is no longer in the list.

hiroshige-g: This depends on how `used by a registration` is defined.
In the current PR, a fetch entry is not considered `used` even if it is referenced from a registration pending removal.
Alternatively we might be able to define it more like reference counting or garbage collecting (i.e. remove the fetch entry from the list when the fetch entry is about to be destructed). In that case, this corner case would disappear. (I'm a little less sure how to spec in this way though)
So far I don't see any specific reasons to prefer each of these ways: while there are very slight observable behavior diferrences, I don't expect the differences have any real impacts for ordinal use cases.

irori: Understood. Thanks for the explanation!

I agree that the difference in behavior in this corner case does not have any impact for ordinal use cases. Current definition of `used by a registration` looks good to me.

irori (approved):

hayatoito:
I think it's fine to merge this at this point. Thanks!
For remaining issues, we can address later.

PR #725, opened 2022-03-30
[subresource loading] Call fetch (Fix #723)

### LINKED ISSUE #723: [subresource fetch spec] Should bundles loaded with main fetch instead of HTTP-network-or-cache fetch?
The current draft directly uses [HTTP-network-or-cache fetch](https://fetch.spec.whatwg.org/#concept-http-network-or-cache-fetch), but probably this bypasses many security checks.
Should we use [main fetch](https://fetch.spec.whatwg.org/#concept-main-fetch) instead, or are there any context/intention for this?

hayatoito:
Calling HTTP-network-or-cache fetch is wrong. We should call fetch.


subresource-loading.bs
```diff
@@ -211,8 +211,7 @@ bundle=] |web bundle| and [=fetch params=] |fetch params|:
    version of `1`, but this specification tracks what’s actually implemented in
    browsers, which still uses draft versions.
 
-1. Let |response| be the result of running <a spec="fetch">HTTP-network-or-cache
-   fetch</a> given |fetch params|.
+1. Let |response| be the result of running [=fetch=] given |request|.
```
irori, line 214: [Fetch](https://fetch.spec.whatwg.org/#concept-fetch) returns not a response but a [fetch controller](https://fetch.spec.whatwg.org/#fetch-controller). I guess we need to pass processResponse* algorithm(s) to fetch.
See https://fetch.spec.whatwg.org/#fetch-elsewhere

hayatoito: Nice catch! Done.

hayatoito:
Thanks for the review. Now we are passing processResponse algorithm.

subresource-loading.bs
```diff
@@ -211,12 +213,17 @@ bundle=] |web bundle| and [=fetch params=] |fetch params|:
...
    Web bundle is never fetched from other web bundles.
 
+## Process web bundle response ## {#process-web-bundle-response}
+
+To <dfn id="concept-process-web-bundle-response">process web bundle
+response</dfn> given [=web bundle=] |web bundle| and [=response=] |response|:
+
 1. If |response|'s [=response/status=] is 200,
 
    1. Set |web bundle|'s [=web bundle/fetched bundle=] to the result of parsing
```
irori, line 229: Note that `processResponse` is called when response's header list has been received; response body bytes are not fully avaiable yet.

This step passes response's body to the bundle parser (that matches our implementation which supports streaming parsing), but you can only set the parsing result to `fetched bundle` asynchronously.

hayatoito: A good point. I'm not sure what is the best way to mention that, but I've added a note to mention that UA might parse the bundle by [incrementally read](https://fetch.spec.whatwg.org/#body-incrementally-read) a body.

hayatoito:
Thanks for the review.  Note that I've rebased the branch for the recent change on main branch.

subresource-loading.bs
irori: Could you state that parser runs asynchronously? Something like:

> When the parse algorithm asynchronously completes, set web bundle's fetched bundle to the result...

hayatoito: Sure. Clarified that we set items when parse algorithm asynchronously completes. UA can optimize here, but the spec doesn't have to mention the details.

subresource-loading.bs
```diff
@@ -211,12 +215,17 @@ bundle=] |web bundle| and [=fetch params=] |fetch params|:
    version of `1`, but this specification tracks what’s actually implemented in
    browsers, which still uses draft versions.
 
-1. Let |response| be the result of running <a spec="fetch">HTTP-network-or-cache
-   fetch</a> given |fetch params|.
+1. [=Fetch=] |request| with [=processResponse=] algorithm set to [=process web
+   bundle response=] which is partially applied with |web bundle|.
 
-   Note: Chromium's current implementation doesn't allow a *nested bundle*. A
+   Note: Chromium's current implementation doesn't allow a _nested bundle_. A
```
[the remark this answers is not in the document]
hayatoito: Ah, again... Since the formatter automatically does this, I just removed a bold here.

hayatoito:
Thanks for the review! Addressed.

irori (approved):
Thanks, lgtm.

PR #741, opened 2022-05-13
Switch from `go get` to `go install`

reillyeon:
According to https://golang.org/doc/go-get-install-deprecation the `go get` command is no longer supported outside of a module and `go install` should be used to build and install a command.


irori (approved):
Thanks!

PR #743, opened 2022-05-16
[subresource loading] Skip service workers (Fix #676)

### LINKED ISSUE #676: Better integration of Service Worker and Subresource Web Bundle loading
The [explainer](https://github.com/WICG/webpackage/blob/main/explainers/subresource-loading.md) of "Subresource loading with Web Bundles" doesn't address the expected behavior of the service worker fetch interception.
And in the current Chromium implementation, if the service worker of the page intercepts the web bundle request, the subsequent subresource requests can’t be served from the web bundle.

As a short term solution, this CL https://crrev.com/c/3145076 sets the SkipServiceWorker flag of the both requests. So the inner subresource can be served from the web bundle.

As a long term solution, we need to figure out the ideal design of the service worker integration. And we need to introduce an option argument {skip_service_worker: false (default is true)}.

Chromium side issue: https://crbug.com/1240424

hayatoito:
Fix #676. Skip service workers.

***
<a href="https://pr-preview.s3.amazonaws.com/hayatoito/webpackage/pull/743.html" title="Last updated on May 16, 2022, 6:20 AM UTC (6786dc2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/webpackage/743/5dfe4a2...hayatoito:6786dc2.html" title="Last updated on May 16, 2022, 6:20 AM UTC (6786dc2)">Diff</a>

### DIFF
--- a/subresource-loading.bs
+++ b/subresource-loading.bs
@@ -349,6 +349,20 @@ from the document</a>, user agents must run the following algorithm:
 
 # Fetch monkeypatches # {#fetch-monkeypatches}
 
+## Monkeypatch fetch ## {#monkeypatch-fetch}
+
+In <a spec="fetch">fetch</a>, before
+
+> 2. Let |taskDestination| be null.
+
+add the following step:
+
+2. If the result of [=find a matching web bundle registration=] given |request|
+   is null, set |request|'s [=request/service-workers mode=] to "`none`".
+
+Note: This means that no service workers will get events for a subresource
+loading from a webbundle.
+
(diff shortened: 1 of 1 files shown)


irori (approved):
This change LGTM, but can we leave #676 open? This looks like the "short term solution" described in #676 and I'm not sure if skipping ServiceWorker is ideal.

hayatoito:
Yeah, it's a short term solution. I think we can close #676 now and re-open that anytime when we need to re-visit. Let me add a comment to #676 to clarify that.

PR #746, opened 2022-05-23
js/bundle: Support relative URLs

### LINKED ISSUE #739: Improve relative URLs support in Go/JavaScript bundle tools
Although relative URLs are now explicitly allowed in WebBundles (#647), our tools do not have good support for them. Currently the only way to create a bundle with relative URL is to use a HAR file that contains relative URLs.

irori:
This fixes #739.


hayatoito (approved):
LGTM

PR #758, opened 2022-06-01
Move CBOR code from signedexchange to internal

sonkkeli:
The CBOR library will be needed by the integrity-block signer as well.

There are also some white-space linter fixes. I am using this: https://marketplace.visualstudio.com/items?itemName=golang.Go


irori (approved):
lgtm

PR #759, opened 2022-06-01
Move certs.go from signedexchange to internal

sonkkeli:
I will in the following CLs update ParsePrivateKey and parsePrivateKeyBlock functions so that they can also support our use case with ed25519.


irori (approved):
lgtm

PR #767, opened 2022-06-13
Add SHA-512 hashing in chunks

sonkkeli:
This branch might need to rebase after the other PR is merged.


go/bundle/cmd/sign-bundle/integrityblock.go
```diff
@@ -48,9 +48,7 @@ func obtainIntegrityBlock(bundleFile *os.File) (*integrityblock.IntegrityBlock,
 		return nil, 0, err
 	}
 
-	// Unlike web bundle length, integrity block length cannot be more than max int.
-	integrityBlockLen := int(fileStats.Size() - webBundleLen)
-
+	integrityBlockLen := fileStats.Size() - webBundleLen
```
irori, line 51: This can be negative if the WebBundle's last 8 bytes are corrupted. Should we have a guard against that?

go/integrityblock/hash.go
```diff
@@ -0,0 +1,52 @@
...
+		return hash512ChunkSizeForTesting
+	}
+	return hash512ChunkSize
+}
+func SetHash512ChunkSizeForTesting(val int64) {
+	hash512ChunkSizeForTesting = val
+}
+
+// ComputeWebBundleSha512 computes the SHA-512 hash over the given web bundle file with the given chunk size. chunkSize is for improved testability.
+func ComputeWebBundleSha512(bundleFile *os.File, offset int64) ([]byte, error) {
```
irori, line 25: I think this should work:
```go
func ComputeWebBundleSha512(bundleFile *os.File, offset int64) ([]byte, error) {
	bundleFile.Seek(offset, io.SeekStart)
	h := sha512.New()
	_, err := io.Copy(h, bundleFile)
	if err != nil {
		return nil, err
	}
	return h.Sum(nil), nil
}
```
The `io.Copy()` will do chunked read/write under the hood.

go/bundle/cmd/sign-bundle/integrityblock.go
sonkkeli: Added!

go/integrityblock/hash.go
```diff
@@ -0,0 +1,22 @@
+package integrityblock
+
+import (
+	"crypto/sha512"
+	"io"
+	"os"
+)
+
+// ComputeWebBundleSha512 computes the SHA-512 hash over the given web bundle file.
+func ComputeWebBundleSha512(bundleFile *os.File, offset int64) ([]byte, error) {
```
irori, line 10:
```suggestion
func ComputeWebBundleSha512(bundleFile *io.ReadSeeker, offset int64) ([]byte, error) {
```
Then you can use non-file types that implement `io.ReadSeeker` (and still you can pass `*io.File` to this function). For example, `bytes.Reader` may be used in tests like this:

```go
bundleBytes = []byte{...}
got, err := ComputeWebBundleSha512(bytes.NewReader(bundleBytes), 0)
```

go/integrityblock/hash.go
```diff
@@ -0,0 +1,22 @@
+package integrityblock
```
irori, line 1: Now this file contains only one pretty simple function. Can we merge this into integrityblock.go?

sonkkeli:
PTAL :)

irori (approved):
lgtm

PR #771, opened 2022-06-15
js/bundle: Switch to cborg CBOR library

irori:
This would make the `wbn` library easier to be used in browsers, because the `cborg` library uses `Uint8Array`s rather than Node.js `Buffer`s.

This also removes some code to workaround the quirks of the `cbor` library.


hayatoito (approved):
LGTM.
Just in curious, is there any remaining Node.js specific APIs being used?
Is this ready for the browser environment?

PR #773, opened 2022-06-15
Add the new signature to the integrity block

sonkkeli:
This shall be merged after the signature generation


go/integrityblock/integrityblock.go
```diff
@@ -152,6 +152,16 @@ func ObtainIntegrityBlock(bundleFile *os.File) (*IntegrityBlock, int64, error) {
 	return integrityBlock, integrityBlockLen, nil
 }
 
+func AddNewSignatureToIntegrityBlock(integrityBlock *IntegrityBlock, signatureAttributes map[string][]byte, signature []byte) *IntegrityBlock {
```
irori, line 155: Looking at this function signature I initially thought it would return a new `IntegrityBlock`, but actually it mutates the given `IntegrityBlock`.

How about making it a method of `IntegrityBlock`:

```go
func (ib *IntegrityBlock) AddNewSignature(signatureAttributes map[string][]byte, signature []byte) {
   ...
}
```

PR #774, opened 2022-06-15
Write the integrity block and the web bundle bytes to a new signed web bundle file

sonkkeli:
This shall be merged after the signature generation and the integrity signature addition PRs have both been merged


irori (approved):
lgtm

PR #775, opened 2022-06-16
js/bundle: Remove Node-specific library dependency from encoder

irori:
Now the `wbn` module depends only on `cborg`.


hayatoito (approved):
LGTM.

PR #776, opened 2022-06-21
js/bundle: Migrate to ES Modules

irori:
Before this, TypeScript source files under `src/` were converted to CommonJS module format by `tsc`. Now we use `"module":"esnext"` in `tsconfig.json`, meaning that generated `.js` files are in ES Module format.

For backward compatibility, a CommonJS module is also built (using esbuild), so that `require('wbn')` will still work.


hayatoito (approved):
Looks nice! LGTM.

PR #778, opened 2022-06-22
Add deterministicy check for unsigned integer CBOR bytes

sonkkeli:


go/internal/cbor/deterministic.go
```diff
@@ -0,0 +1,121 @@
...
+	if err != nil {
+		return -1, err
+	}
+
+	_, err = unsignedIntegerCompareValueToLimit(input[0 : lengthInBytes+1])
+	if err != nil {
+		return -1, err
+	}
+
+	return lengthInBytes, nil
```
irori, line 92: Shouldn't be `lengthInBytes + 1`?

go/internal/cbor/types.go
```diff
@@ -13,3 +13,22 @@ const (
 	TypeTag         = 0xc0
 	TypeOther       = 0xe0
 )
+
+// getMajorType returns the first 3 bits of the first byte representing cbor's major type.
+// https://www.rfc-editor.org/rfc/rfc8949.html#name-major-types
+func getMajorType(b byte) Type {
+	switch (b & 0xff) >> 5 {
```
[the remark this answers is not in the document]
sonkkeli: Isn't that then untyped ikf i'd just return that? I would like to have the type for better readability later in code. But I changed b & 0b11100000 for the switch case!

go/internal/cbor/deterministic.go
```diff
@@ -0,0 +1,121 @@
...
+	"math/big"
+)
+
+// Deterministic function loops through the given byte array containing CBOR encoded
+// byte sequences and checks that they follow the deterministic principles described here:
+// https://www.rfc-editor.org/rfc/rfc8949.html#name-deterministically-encoded-c.
+func Deterministic(input []byte) bool {
+	index := 0
+
+	for index < len(input) {
```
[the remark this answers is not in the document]
sonkkeli: Yes this is intentional, you can pass there any CBOR sequence that fits to byte array and it will tell you its deterministicy.

go/internal/cbor/addinfo.go
```diff
@@ -0,0 +1,94 @@
...
+	case AdditionalInfoOneByte:
+		return 1, nil
+	case AdditionalInfoTwoBytes:
+		return 2, nil
+	case AdditionalInfoFourBytes:
+		return 4, nil
+	case AdditionalInfoEightBytes:
+		return 8, nil
+	default:
+		return -1, fmt.Errorf("getLengthByAdditionalInfo() should never be called with: %v", ainfo)
```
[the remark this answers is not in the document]
sonkkeli: How do I test these panics? I was trying with this https://stackoverflow.com/questions/31595791/how-to-test-panics but I didn't get them working.

go/internal/cbor/deterministic.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+// getUnsignedIntegerValue converts the byte array containing the CBOR encoded unsigned integer into an actual value.
+func getUnsignedIntegerValue(input []byte) uint64 {
+	if convertToAdditionalInfo(input[0]) == AdditionalInfoDirect {
+		// Get the value from the last 5 bits of the byte.
+		return uint64(getAdditionalInfoDirectValue(input[0]))
+	}
+
+	return new(big.Int).SetBytes(input[1:]).Uint64()
```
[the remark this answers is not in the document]
sonkkeli: With [encoding/binary](https://pkg.go.dev/encoding/binary) it assumes that the byte array is always of length 8 which is not the case for us so it gives runtime error: index out of range [7] with length 1 [recovered]. Not sure if there's another way to do this?

go/internal/cbor/deterministic.go
```diff
@@ -0,0 +1,121 @@
...
+	if err != nil {
+		return 0, err
+	}
+
+	value := getUnsignedIntegerValue(input[0:])
+
+	if value < limit {
+		return 0, fmt.Errorf("When following CBOR deterministic principles, %v should not be represented with %v bytes.", value, len(input)-1)
+	}
+	return value, nil
```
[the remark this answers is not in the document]
sonkkeli: These functions and their splitting may seem a bit odd because there are some functionality still coming up. I could remove this since it's not needed right now, but it will be needed when the map and array come to the picture. (Because we want to check that the number representing the amount of items on the array/map is also deterministic)

go/internal/cbor/deterministic.go
sonkkeli: The +1 is in the main loop :)

go/internal/cbor/deterministic.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+// getUnsignedIntegerValue converts the byte array containing the CBOR encoded unsigned integer into an actual value.
+func getUnsignedIntegerValue(input []byte) uint64 {
+	if convertToAdditionalInfo(input[0]) == AdditionalInfoDirect {
```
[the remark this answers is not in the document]
sonkkeli: The reason why getUnsignedIntegerValue is separate is because in the tests I want to compare that my on-purpose non-deterministically encoded CBOR still has the same value as the original bytes.

I merged unsignedIntegerCompareValueToLimit to unsignedIntegerDeterministic :)

go/internal/cbor/addinfo.go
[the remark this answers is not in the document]
sonkkeli: Nice thanks! :)

go/internal/cbor/types.go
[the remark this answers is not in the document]
irori: You can cast it to `Type`.
```go
    return Type(b & 0b11100000)
```
This is safe because TypeXXX covers all the 8 possible values of `b & 0b11100000`.

go/internal/cbor/deterministic.go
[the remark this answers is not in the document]
irori: Acknowledged.

go/internal/cbor/deterministic.go
```diff
@@ -0,0 +1,107 @@
...
+import (
+	"errors"
+	"fmt"
+	"math/big"
+)
+
+// Deterministic function loops through the given byte array containing CBOR encoded
+// byte sequences and checks that they follow the deterministic principles described here:
+// https://www.rfc-editor.org/rfc/rfc8949.html#name-deterministically-encoded-c.
+func Deterministic(input []byte) (bool, error) {
```
irori, line 12: The `bool` return value is redundant, because `err == nil` means success.

go/internal/cbor/deterministic.go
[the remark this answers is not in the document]
sonkkeli: I think it's already there: "byte array containing CBOR encoded byte sequences"

go/internal/cbor/deterministic.go
sonkkeli: It's not length -1, it is the right length and then the mainloop is where it keeps track of the CBOR sequences. So if there's like e.g. 2 unsigned integers behind each other of length 2 and 3, the length of the 1st one is calculated and deterministicRec returns 2, then the main function Deterministic notices that there's still something left on the byte array and it increases the index on the main loop to the first byte of the second CBOR item (of len 3) in the sequence and then the deterministicRec returns 3 for that. Then on the main loop it notices that index is no longer less than length of input and the function exits. I am actually not sure if it makes any difference to move the +1 inside deterministicRec for when arrays and maps come, maybe not? I have to code the rest and test

irori: I'm not sure if I'm following you. If the first `deterministicRec` call returns 2, the second `deterministicRec` is called with `input[3:]`. But the second unsigned integer is located at `index[2:5]`, right?

```
 [0] [1] [2] [3] [4]
+---+---+---+---+---+
| first |  second   |
+---+---+---+---+---+
```

go/internal/cbor/addinfo.go
```diff
@@ -0,0 +1,81 @@
...
+// item is computed from the argument); in major type 2 and 3, it gives the length of the string
+// data in bytes that follow; and in major types 4 and 5, it is used to determine the number of data
+// items enclosed.
+
+// https://www.rfc-editor.org/rfc/rfc8949.html#name-specification-of-the-cbor-e
+
+type AdditionalInfo int
+
+const (
+	// The integers match with the length in bytes apart from reserved and indefinite.
```
irori, line 19: I think this is confusing. I'd use simple sequential enum values (using `iota`), and use a `switch` statement in `getAdditionalInfoLength()`.

go/internal/cbor/addinfo.go
```diff
@@ -0,0 +1,81 @@
...
+	case 0x1F:
+		return AdditionalInfoIndefinite
+	default:
+		return AdditionalInfoDirect
+	}
+}
+
+// getAdditionalInfoDirectValue is used for AdditionalInfoDirect to read the value following the additional info.
+func getAdditionalInfoDirectValue(b byte) byte {
+	return b & 0x1F //=0b00011111
```
irori, line 50:
```suggestion
	return b & 0b00011111
```

go/internal/cbor/deterministic.go
```diff
@@ -0,0 +1,107 @@
...
+
+	// TODO(sonkkeli): Remove debug print.
+	fmt.Printf("cborItemLengthInBytes=%v\n", cborItemLengthInBytes)
+
+	return cborItemLengthInBytes, nil
+
+}
+
+// unsignedIntegerDeterministic calculates the value of the unsigned integer to ensure that
+// the right amount of bytes is used in the CBOR encoding and returns the length of the unsigned integer.
```
irori, line 73: When a function returns multiple non-error values, please document them all.

go/internal/cbor/deterministic_test.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+
+	// Also deterministically encoded unsigned integers following each other return true.
+	concatenatedTestCases := multiappend(uint64TestCasesAsBytes)
+	err := Deterministic(concatenatedTestCases)
+	if err != nil {
+		t.Error("Deterministically encoded unsigned integers should not return false for deterministicy.")
+	}
+}
+
```
irori, line 49: Can you add a test case where the input to `Deterministic` is a concatenation of two unsigned integers?

sonkkeli: There is a test where there is a concatenation of all the unsigned integers used in the example. I think this should cover that even better? See TestUintDeterministic second part. I could separate this to it's own test though if that would be clearer?

go/internal/cbor/deterministic.go
```diff
@@ -0,0 +1,121 @@
...
+	}
+
+	// TODO(sonkkeli): Remove debug print.
+	fmt.Printf("cborItemLengthInBytes=%v\n", cborItemLengthInBytes)
+
+	return cborItemLengthInBytes + 1, nil
+}
+
+// unsignedIntegerDeterministic calculates the value of the unsigned integer to ensure that
+// the right amount of bytes is used in the CBOR encoding and returns the length of the unsigned integer.
```
irori, line 72:
```suggestion
// the right amount of bytes is used in the CBOR encoding.
```
(The return values are explained in the next sentence)

go/internal/cbor/deterministic_test.go
```diff
@@ -0,0 +1,174 @@
...
+// TestNotDeterministic tests that if the value of the unsigned integer is kept the same but there are
+// additional non-necessary empty bytes added in-front of the actual value's bytes, it is no longer
+// considered as deterministically encoded CBOR.
+func TestUintNotDeterministic(t *testing.T) {
+	// The additional information values for representing the number with 1, 2, 4 or 8 bytes.
+	firstBytes := [][]byte{{0x18}, {0x19}, {0x1A}, {0x1B}}
```
irori, line 55: It's strange that `firstByte` below is of type `[]byte`. Why not make it
```go
[]byte{0x18, 0x19, 0x1A, 0x1B}
```
?

go/internal/cbor/deterministic.go
```diff
@@ -0,0 +1,121 @@
...
+		index += length
+	}
+	return nil
+}
+
+// deterministicRec is a recursively called helper function to check the deterministicy of a bytes array.
+// TODO(sonkkeli): Recursively called parts will be the maps and arrays, which contain other types, meaning
+// unsigned integers, byte strings, text strings.
+func deterministicRec(input []byte) (int, error) {
+	cborItemLengthInBytes := 0
```
irori, line 32: I think it's easier to read if each `switch`-branch returns a value directly, instead of using this temporary variable. (You can still inject a debug print in the caller side.)

go/internal/cbor/deterministic_test.go
sonkkeli: It was like that cos it was appended to other byte arrays but it doesnt really matter, changed it to be a 1D list now :)

sonkkeli:
PTAL :)

irori (approved):
lgtm

PR #783, opened 2022-06-30
Add deterministicy check for map type CBOR bytes

sonkkeli:


go/internal/cbor/deterministic_test.go
```diff
@@ -188,6 +188,54 @@ func TestByteStringIsNotDeterministic(t *testing.T) {
 	})
 }
 
+func TestMapIsDeterministic(t *testing.T) {
+	hello := multiappend([]byte{0b01100101}, []byte("hello"))
```
irori, line 192: Binary-encoded CBOR headers like `[]byte{0b01100101}` are hard to read. Can you create a helper function that constructs a CBOR header, for readability?

irori: Doing it in a separate CL sounds good.

irori (approved):

PR #785, opened 2022-07-01
Create a helper method for converting majortype and additionalinfo value into a CBOR header

sonkkeli:


irori (approved):
Thanks for doing this!

PR #788, opened 2022-07-07
Disable trying to use dump-bundle with integrity-block

sonkkeli:
peletskyi@ 's opinion was that this approach would be better than trusting on the web bundle length vs file size. Let me know if you disagree :)

### DIFF
--- a/go/bundle/cmd/dump-bundle/main.go
+++ b/go/bundle/cmd/dump-bundle/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"errors"
 	"flag"
 	"fmt"
 	"log"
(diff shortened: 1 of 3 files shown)


go/bundle/cmd/dump-bundle/main.go
```diff
@@ -91,7 +93,30 @@ func isTextType(mimeType string) bool {
 	return strings.HasPrefix(m, "text/") || m == "application/javascript"
 }
 
+func checkDoesNotHaveIntegrityBlock(path string) error {
```
irori, line 96: Can you put this check into `ReadBundleFromFile()` so that the bundle file doesn't have to be opened twice?

irori (commented):
The approach lgtm.

sonkkeli:
PTAL :)

irori (approved):
lgtm

PR #791, opened 2022-07-26
Use corsAttributeState of Anonymous in alternate SignedExchange prefetch

### LINKED ISSUE #790: Alternate SignedExchange prefetch and CORS
This issue is about [Signed Exchange subresource substitution](https://github.com/WICG/webpackage/blob/main/explainers/signed-exchange-subresource-substitution.md).

In the current SignedExchange loading spec, step 19.6.8 of [Monkeypatch Link type "prefetch"](https://wicg.github.io/webpackage/loading.html#mp-link-type-prefetch) says that sxgRequest (the request used to fetch alternate SXG from network) is made with a corsAttributeState of No CORS. However,

1. This doesn't work well with ORB. See https://github.com/annevk/orb/issues/32
2. In principle, new feature shouldn't use no-cors.

So, we want to provide a way to make alternate SXG prefetch requests in `cors` mode.

One possible way is to use the [`crossorigin` attribute](https://html.spec.whatwg.org/multipage/semantics.html#link-options-crossorigin) of the outer `Link: <>;rel=alternate` header. For example, the following `Link` header would specify that `lib.js.sxg` should be prefetched in `cors` mode and `same-origin` credential mode:
```
Link: <https://feed.example/sxg.publisher.example/lib.js.sxg>;
        rel="alternate";
        type="application/signed-exchange;v=b3";
        anchor="https://cdn.publisher.example/lib.js";
        crossorigin="anonymous"
```

However, `crossorigin="anonymous"` would be the only reasonable value here, because

* No-cors (`crossorigin` attribute is not specified) will be blocked once ORB is fully enabled.
* We don't want to allow "use-credentials", to [prevent user tracking](https://github.com/WICG/webpackage/blob/main/explainers/signed-exchange-subresource-substitution.md#:~:text=To%20avoid%20leaking%20user%2Dspecific%20data%20in%20the%20distributor%20of%20signed%20exchange%2C%20the%20prefetch%20request%20must%20not%20contain%20credentials.).

So, I propose that alternate SXG prefetch always uses request mode "cors" and credentials mode "same-origin", regardless of the `crossorigin` attribute of `Link: <>;rel=alternate`.

irori:
This fixes #790.

This changes the corsAttributeState used to create a request for alternate SignedExchange prefetch from No CORS to Anonymous. See #790 for the motivation of this change.

For consistency this also changes the corsAttributeState for `requestForMatch`, but it doesn't make any behavioral changes because the [matches the stored exchange](https://wicg.github.io/webpackage/loading.html#matches-the-stored-exchange) algorithm does not refer the request's mode and credentials mode.

### DIFF
--- a/loading.bs
+++ b/loading.bs
@@ -178,7 +178,6 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
         text: fetch and process the linked resource; url: semantics.html#fetch-and-process-the-linked-resource
         text: linked resource fetch setup steps; url: semantics.html#linked-resource-fetch-setup-steps
         text: navigation params; url: browsing-the-web.html#navigation-params
-        text: No CORS; url: urls-and-fetching.html#attr-crossorigin-none
         text: process the linked resource; url: semantics.html#process-the-linked-resource
         text: response; for: navigation params; url: browsing-the-web.html#navigation-params-response
         text: selecting an image source; url: images.html#select-an-image-source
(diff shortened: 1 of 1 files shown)


irori:
@horo-t Would you take a look?

irori:
@horo-t Ping?

horo-t (approved):
lgtm

Sorry for the delay.

irori:
Thanks!

PR #796, opened 2022-09-27
Fix go/signedexchange/certurl test in Go 1.19

irori:
It was failing because Go 1.19 formats `[]byte{0x7f}` to `"\x7f"`, while Go <=1.18 formats it to `"\u007f"`.

Now `TestParsePEM()` does byte-by-byte comparison with the expected data and use `CborBinaryToReadableString()` only to format the error message.


hayatoito (approved):
RS-LGTM.

PR #799, opened 2022-10-28
Signature attributes should always be freshly generated and never read from the stack

sonkkeli:
This fixes a bug which I had on generating the payload for signing in case there would already be existing signatures on the signature stack.


cmfcmf (approved):
Looks good now 👍

PR #800, opened 2022-10-31
Isolate creating signature and adding it to integrity block into its own function and add better tests with real ed25519 keys

sonkkeli:


go/integrityblock/integrityblock.go
```diff
@@ -228,3 +228,35 @@ func WebBundleHasIntegrityBlock(bundleFile io.ReadSeeker) (bool, error) {
 
 	return bytes.Compare(IntegrityBlockMagic, possibleMagic) == 0, nil
 }
+
+// GetIntegrityBlockWithNewSignature contains the main logic for generating the new signature and
+// prepending the integrity block's signature stack with a new integrity signature object.
+func GetIntegrityBlockWithNewSignature(integrityBlock *IntegrityBlock, ed25519privKey ed25519.PrivateKey, webBundleHash []byte, signatureAttributes map[string][]byte) (*IntegrityBlock, error) {
```
irori, line 234: The `IntegrityBlock` return value is redundant, because this function always returns the `IntegrityBlock` in the first parameter.

Since this function mutates the given `IntegrityBlock`, I'd make it a method of `IntegrityBlock` type. The function signature would be:
```Go
func (integrityBlock *IntegrityBlock) AddNewSignature(ed25519privKey ed25519.PrivateKey, webBundleHash []byte, signatureAttributes map[string][]byte) error
```

go/integrityblock/integrityblock_test.go
```diff
@@ -193,3 +195,120 @@ func TestSignedWebBundleHasIntegrityBlockestAnother(t *testing.T) {
...
+	got, err := testhelper.CborBinaryToReadableString(integrityBlockBytes)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if got != want {
+		t.Errorf("integrityblock: got: %s\nwant: %s", got, want)
+	}
+}
+
+func BytesToCborAndToReadableStringHelper(bts []byte) (string, error) {
```
irori, line 300: Let's not export this function.
```suggestion
func bytesToCborAndToReadableStringHelper(bts []byte) (string, error) {
```

irori:
Also, please try to follow the [50/72 rule](https://www.midori-global.com/blog/2018/04/02/git-50-72-rule) when writing a commit message.

sonkkeli:
I wasnt able to update the commit msg here to be multi-line (i tried git commit --amend on the cmd + git push -f but it didnt override this), so I close this and create a new one!

PR #802, opened 2022-11-02
Encapsulate logic for signing to improve testability

sonkkeli:
Isolate creating signature and adding it into integrity block to its own function and add better tests with real ed25519 keys.


irori (approved):
lgtm

PR #808, opened 2022-11-07
Use constants for versions and separate cli.ts logic

sonkkeli:
No functional changes, just refactoring a little and using constants for the versions


js/bundle/src/cli.ts
```diff
@@ -37,55 +38,82 @@ export function addFilesRecursively(
...
+}
+
+export function main() {
+  const options = readOptions();
+  optionsValid(options);
+
+  const version =
+    options.formatVersion === undefined
+      ? DEFAULT_VERSION
+      : options.formatVersion;
```
cmfcmf, line 102: If support for Node.js 12 is dropped at some point, then this could be rewritten as `const version = options.formatVersion ?? DEFAULT_VERSION` (`??` was introduced in Node.js 14).

js/bundle/src/cli.ts
```diff
@@ -37,55 +38,82 @@ export function addFilesRecursively(
...
+
+  const builder = new BundleBuilder(version);
+
+  if (version === B2 && options.primaryURL) {
+    builder.setPrimaryURL(options.primaryURL);
+  }
+
+  if (version === B1) {
+    builder.setPrimaryURL(options.primaryURL || options.baseURL);
```
cmfcmf, line 111: I feel like a `switch(version)` statement could be a bit more readable here. Or something like `if (version === B2) { if(options.primaryURL) {} } else if (version === B1) { ... }`

js/bundle/src/cli.ts
```diff
@@ -37,55 +38,82 @@ export function addFilesRecursively(
...
     .option(
       '-m, --manifestURL <URL>',
-      'manifest URL (only valid with format version "b1")'
+      'manifest URL (only valid with format version "' + B1 + '")'
     )
     .option('-o, --output <file>', 'webbundle output file', 'out.wbn')
     .parse(process.argv);
+}
 
+function optionsValid(options: any) {
```
cmfcmf, line 68: I think it would be a bit nicer if 
- this function was called `validateOptions`, and
- this function returned the error messages or `null` / `undefined`, and then the main function would check if the error message is not `null` / `undefined`, log it to the console, and exit.

js/bundle/src/encoder.ts
```diff
@@ -23,8 +24,8 @@ export class BundleBuilder {
   private currentResponsesOffset = 0;
   private compatAdapter: CompatAdapter;
 
-  constructor(formatVersion: string = 'b2') {
-    if (formatVersion != 'b1' && formatVersion != 'b2') {
+  constructor(formatVersion: string = DEFAULT_VERSION) {
```
cmfcmf, line 27: WDYT about making the type of `formatVersion` in this file `'b1' | 'b2'` and moving the following code into `main`:

```ts
if (!APPROVED_VERSIONS.includes(formatVersion)) {
  throw new Error(`Invalid webbundle format version`);
}
```

js/bundle/src/constants.ts
```diff
@@ -0,0 +1,4 @@
+export const B1 = 'b1';
+export const B2 = 'b2';
```
cmfcmf, line 2: I am not sure about the value of the `B1` and `B2` constants - I think these could just as well remain strings.

js/bundle/src/constants.ts
```diff
@@ -0,0 +1,4 @@
...
+export const DEFAULT_VERSION = B2;
+export const APPROVED_VERSIONS = [B1, B2];
```
cmfcmf, line 4:
```suggestion
export const APPROVED_VERSIONS = [B1, B2] as const;
```

Marking this with `as const` can lead to better type inference, since the type is currently `string[]`, but with `as const` it would literally be `['b1', 'b2']`.

js/bundle/src/cli.ts
```diff
@@ -37,55 +38,82 @@ export function addFilesRecursively(
...
       '-f, --formatVersion <formatVersion>',
-      'webbundle format version, possible values are "b1" and "b2" (default: "b2")',
-      'b2'
+      'webbundle format version, possible values are ' +
+        APPROVED_VERSIONS.map((v) => `"${v}"`).join(' and ') +
+        ' (default: "' +
+        DEFAULT_VERSION +
+        '")',
+      DEFAULT_VERSION
     )
```
cmfcmf, line 53: Have you looked into using the `.choices` function here? [From the commander docs](https://www.npmjs.com/package/commander):

```ts
  .addOption(new Option('-d, --drink <size>', 'drink size').choices(['small', 'medium', 'large']))
```

js/bundle/src/cli.ts
sonkkeli: Yeah, since that's not the case yet, I didnt use that yet

js/bundle/src/encoder.ts
sonkkeli: I think this is here because this constructor is used from 3 different places (outside of cli.ts), like the 2 other plugins as well. So it needs to be rechecked.

js/bundle/src/constants.ts
sonkkeli: I think the value is if someone decides to refactor them, they cant do it without changing all the places and breaking the code. With strings it's possible to leave something unchanged and not notice it since the code would still compile

js/bundle/src/constants.ts
sonkkeli: Done

js/bundle/src/cli.ts
sonkkeli: I feel like this is a functional change so might look into this a bit later if i have time

js/bundle/src/encoder.ts
cmfcmf: Ah, I didn't realize that. 👍

js/bundle/src/constants.ts
cmfcmf: > With strings it's possible to leave something unchanged and not notice it since the code would still compile

My goal would be to make things typesafe-enough that it would no longer compile if that happens. But I do understand that this might be out of scope of this PR.

js/bundle/src/encoder.ts
```diff
@@ -23,8 +30,8 @@ export class BundleBuilder {
   private currentResponsesOffset = 0;
   private compatAdapter: CompatAdapter;
 
-  constructor(formatVersion: string = 'b2') {
-    if (formatVersion != 'b1' && formatVersion != 'b2') {
+  constructor(formatVersion: FormatVersion = DEFAULT_VERSION) {
```
cmfcmf, line 33: nit: Also use `FormatVersion` throughout the rest of this file, e.g. in `createCompatAdapter` and in the `formatVersion` field of the `CompatAdatper` implementations.

js/bundle/src/constants.ts
```diff
@@ -0,0 +1,5 @@
+export const B1 = 'b1';
+export const B2 = 'b2';
+export const DEFAULT_VERSION = B2;
+export const APPROVED_VERSIONS = [B1, B2] as const;
+export type FormatVersion = 'b1' | 'b2';
```
cmfcmf, line 5: nit: Use `export type FormatVersion = typeof APPROVED_VERSIONS[number];`

cmfcmf (approved):
LGTM, thanks Sonja :)

js/bundle/src/cli.ts
irori: Just a note; since this file is transpiled to JS, we should be able to use new syntaxes as long as we set appropriate `target` in `tsconfig.json`.

js/bundle/src/cli.ts
```diff
@@ -37,55 +44,91 @@ export function addFilesRecursively(
...
     .option(
       '-m, --manifestURL <URL>',
-      'manifest URL (only valid with format version "b1")'
+      'manifest URL (only valid with format version "' + B1 + '")'
     )
     .option('-o, --output <file>', 'webbundle output file', 'out.wbn')
     .parse(process.argv);
+}
 
+function validateOptions(options: any) {
```
irori, line 74: Let's explicitly write the return type if it's not obvious from the function name.
```suggestion
function validateOptions(options: any): string | null {
```

js/bundle/src/cli.ts
```diff
@@ -37,55 +44,91 @@ export function addFilesRecursively(
...
-    if (options.manifestURL) {
-      builder.setManifestURL(options.manifestURL);
-    }
-    addFilesRecursively(builder, options.baseURL || '', options.dir);
-    fs.writeFileSync(options.output, builder.createBundle());
-  } else {
-    console.error('error: invalid format version (must be "b1" or "b2").');
+  if (
+    options.formatVersion !== undefined &&
+    !(APPROVED_VERSIONS as ReadonlyArray<string>).includes(
```
irori, line 81: Why is this cast necessary?

js/bundle/src/decoder.ts
```diff
@@ -31,7 +32,10 @@ export class Bundle {
   constructor(buffer: Uint8Array) {
     const wbn = asArray(cborg.decode(buffer, { useMaps: true }));
 
-    let peekVersion = bytestringToString(wbn[1]).replace(/\0+$/, '');
+    let peekVersion = bytestringToString(wbn[1]).replace(
+      /\0+$/,
+      ''
+    ) as FormatVersion;
```
irori, line 38: This cast breaks type safety, because the `FormatVersion` type is `"b1" | "b2"` but this value can be arbitrary string.

We should do either:

- Let `FormatVersion` be just `string` (losing some type checks), or
- Validate the value before casting to `FormatVersion`.

js/bundle/src/decoder.ts
```diff
@@ -50,7 +54,10 @@ export class Bundle {
     if (bytestringToString(magic) !== '🌐📦') {
       throw new Error('Wrong magic');
     }
-    this.version = bytestringToString(version).replace(/\0+$/, ''); // Strip off the '\0' paddings.
+    this.version = bytestringToString(version).replace(
+      /\0+$/,
+      ''
+    ) as FormatVersion; // Strip off the '\0' paddings.
```
irori, line 60: Same here.

js/bundle/src/cli.ts
sonkkeli: Oh you are right, it's actually not needed. :)

js/bundle/src/decoder.ts
sonkkeli: I added a type predicate to confirm the types!

sonkkeli:
PTAL :)

PR #809, opened 2022-11-09
Create wbn-sign npm package and IntegrityBlockSigner class skeleton

sonkkeli:
Follow-up commit to set up the tests: https://github.com/WICG/webpackage/commit/249d8263b09cf9149358aa1162c3a9cdcaa04541 
Will create a PR when this one has landed :)

### DIFF
--- a/js/sign/src/cli.ts
+++ b/js/sign/src/cli.ts
@@ -1,0 +1,35 @@
+import commander from 'commander';
+import { IntegrityBlockSigner, parseStringKey } from './integrityblock.js';
+import * as fs from 'fs';
+import crypto, { KeyObject } from 'crypto';
+
+function readOptions() {
+  return commander
+    .requiredOption(
+      '-i, --input <file>',
+      'input web bundle to be signed (required)'
+    )
+    .requiredOption(
+      '-k, --privateKey <file>',
+      'path to ed25519 private key (required)'
+    )
+    .option(
+      '-o, --output <file>',
+      'signed web bundle output file',
+      'signed.wbn'
+    )
+    .parse(process.argv);
+}
+
+export function main() {
+  const options = readOptions();
+  const webBundle = fs.readFileSync(options.input);
+
+  const signer = new IntegrityBlockSigner(webBundle, {
+    key: parseStringKey(fs.readFileSync(options.privateKey, 'utf-8')),
+  });
+  const integrityBlock = signer.sign();
+
+  // TODO: Prepend integrity block.
+  // fs.writeFileSync(options.output, webBundle);
+}
--- a/js/sign/src/integrityblock.ts
+++ b/js/sign/src/integrityblock.ts
@@ -1,0 +1,31 @@
+import crypto, { KeyObject } from 'crypto';
+
+// A helper function which can be used to parse string formatted keys to
+// KeyObjects.
+export function parseStringKey(unparsedKey: string): KeyObject {
+  return crypto.createPrivateKey({
+    key: unparsedKey,
+  });
+}
+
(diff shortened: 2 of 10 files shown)


js/sign/src/integrityblock.ts
```diff
@@ -0,0 +1,30 @@
...
+// KeyObjects.
+export function parseStringKey(unparsedKey: string): KeyObject {
+  return crypto.createPrivateKey({
+    key: unparsedKey,
+  });
+}
+
+export class IntegrityBlockSigner {
+  private key: KeyObject;
+  private webBundle: Uint8Array = new Uint8Array();
```
irori, line 13:
```suggestion
  private webBundle: Uint8Array;
```
because it's initialized in the constructor.

Alternatively, you can declare these fields in the constructor signature like this:
```typescript
  constructor(private webBundle: Uint8Array, private key: KeyObject) {
```

js/sign/src/integrityblock.ts
```diff
@@ -0,0 +1,30 @@
...
+
+  constructor(webBundle: Uint8Array, key: KeyObject) {
+    if (!key) {
+      throw new Error('Key is required.');
```
irori, line 17: Is this possible? The type of `key` is not null-able.

js/sign/src/cli.ts
```diff
@@ -0,0 +1,49 @@
...
+      'signed.wbn'
+    )
+    .parse(process.argv);
+}
+
+function readKeyFile(keyPath: string): KeyObject {
+  const file = fs.readFileSync(keyPath);
+  const key = crypto.createPrivateKey({
+    key: file,
+  });
```
irori, line 28: Why not use the `parseStringKey` helper function?

js/sign/src/cli.ts
```diff
@@ -0,0 +1,49 @@
...
+  return key;
+}
+
+export function main() {
+  const options = readOptions();
+  if (!options.privateKey || !options.input) {
```
irori, line 34: This is already checked by `commander`.

js/sign/src/cli.ts
```diff
@@ -0,0 +1,49 @@
+import commander from 'commander';
+import { IntegrityBlockSigner } from './integrityblock.js';
+import * as fs from 'fs';
+import crypto, { KeyObject } from 'crypto';
+
+function readOptions() {
+  return commander
+    .requiredOption(
+      '-i, --input <file>',
+      'input unsigned web bundle (required)'
```
irori, line 10: Does input have to be unsigned? It may be useful if one could specify a signed bundle and `wbn-sign` would push a signature to its signature stack.

js/sign/bin/sign.js
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env node
```
irori, line 1: Let's rename this file to `wbn-sign.js`, to align with the command name.

irori (commented):
`js/sign/out.wbn` probably should not be added to the repository.

js/sign/src/cli.ts
sonkkeli: Yes in the case of the CLI but not in the case of the plugins.

js/sign/src/cli.ts
sonkkeli: I think in the first place we only sign unsigned bundles but it doesnt harm to have it more general on this so I updated the text to be  'input web bundle to be signed (required)' :)

sonkkeli:
PTAL :)

irori (approved):
lgtm

PR #813, opened 2022-11-10
Check web bundle length and create empty integrity block

sonkkeli:


js/sign/src/integrityblock.ts
```diff
@@ -8,13 +13,28 @@ export function parseStringKey(unparsedKey: string): KeyObject {
...
+};
+
 type IntegrityBlockSignerOptions = {
   key: KeyObject;
 };
 
 export class IntegrityBlockSigner {
   private key: KeyObject;
   private webBundle: Uint8Array;
+  private integrityBlock: IntegrityBlock;
```
irori, line 37: `IntegrityBlock | undefined` ? (since it's not initialized in the constructor)

sonkkeli: Done, PTAL :)

irori (approved):

PR #814, opened 2022-11-10
Deterministic check skeleton

sonkkeli:
This is not really doing anything cool yet since all the sub-functions need to be implemented to actually calculate things, but sending out a PR to get the skeleton done so that I have something to start building up on. I will also add some tests once there is something to test and once the test skeleton PR has been merged.

### DIFF
--- a/js/sign/src/deterministic.ts
+++ b/js/sign/src/deterministic.ts
@@ -1,0 +1,55 @@
+export enum MajorType {
+  PosInt = 0,
+  NegInt = 1,
+  ByteString = 2,
+  Text = 3,
+  Array = 4,
+  Map = 5,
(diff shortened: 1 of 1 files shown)


js/sign/src/deterministic.ts
```diff
@@ -0,0 +1,55 @@
+export enum MajorType {
+  TypePosInt = 0,
```
irori, line 2: Can we remove the `Type` prefix from the values? `MajorType.TypeX` looks somewhat redundant.

sonkkeli: Done, PTAL :)

irori (approved):

PR #826, opened 2022-11-17
Check the determisticity of the items during signing

sonkkeli:
Check that the CBOR encoded attributes and integrity block, both empty and signed, follow the deterministic principles.


cmfcmf (approved):
LGTM
Question: Why does `this.integrityBlock` need to be an instance variable?

PR #828, opened 2022-11-17
Write the signed web bundle bytes to the output file

sonkkeli:


js/sign/src/cli.ts
```diff
@@ -30,6 +29,8 @@ export function main() {
   });
   const integrityBlock = signer.sign();
 
-  // TODO: Prepend integrity block.
-  // fs.writeFileSync(options.output, webBundle);
+  fs.writeFileSync(
+    options.output,
+    new Uint8Array([...integrityBlock, ...webBundle])
```
irori, line 34: This can be inefficient if `webBundle` is huge, because `[...integrityBlock, ...webBundle]` creates a big intermediate array of numbers.

You may want to:

- Use `Buffer.concat` to concatenate `integrityBlock` and `webBundle` (still creates a big `Buffer`, but no intermediate array), or
- Write `integrityBlock` and `webBundle` separately, with two `fs.writeSync()` calls.

sonkkeli: I chose the second option :)

irori (approved):

PR #832, opened 2022-11-21
Update README.md with JavaScript snippets

sonkkeli:


js/sign/README.md
```diff
@@ -1,18 +1,85 @@
...
+);
+
+const integrityBlock = new wbnSign.IntegrityBlockSigner(webBundle, {
+  key: privateKey,
+}).sign();
+
+var fd = fs.openSync('signed-webbundle.swbn', 'w');
+fs.writeSync(fd, integrityBlock);
+fs.writeSync(fd, webBundle);
+fs.closeSync(fd);
```
irori, line 43: This unnecessarily exposes the fact that a signed bundle is the concatenation of an integrity block and an unsigned bundle. Also, this wouldn't work when `webBundle` already had an integrity block (`IntegrityBlockSigner` currently does not support that case though).

It would be nice if `IntegrityBlockSigner` provided a function that returns the entire signed bundle as a single `Uint8Array` or `Buffer`.

js/sign/README.md
```diff
@@ -1,18 +1,85 @@
...
+## Usage
+
+Please be aware that the APIs are not stable yet and are subject to change at
+any time.
+
+Signing a web bundle file:
+
+```javascript
+import * as fs from 'fs';
+import * as wbnSign from 'wbnSign';
```
irori, line 27: `from 'wbn-sign'`?

js/sign/README.md
```diff
@@ -1,18 +1,85 @@
...
+This package also includes a CLI tool `wbn-sign` which lets you sign a web
+bundle easily without having to write any additional JavaScript. 
+
+Example command:
+```bash
 wbn-sign \
--i ./wbncli/webbundle.wbn \
+-i ./path/to/webbundle.wbn \
+-o ./path/to/signed-webbundle.swbn \
 -k ~/path/to/ed25519key.pem
```
irori, line 78: Why is only this relative to `~/`?

irori (approved):

PR #833, opened 2022-11-21
Calculate web bundle ID from private ed25519 key

sonkkeli:
Add a helper function to calculate the web bundle's base32 encoded app ID. Also print it out after siging is complete.

### DIFF
--- a/go/bundle/cmd/sign-bundle/integrityblock.go
+++ b/go/bundle/cmd/sign-bundle/integrityblock.go
@@ -4,6 +4,7 @@ import (
 	"crypto"
 	"crypto/ed25519"
 	"errors"
+	"fmt"
 	"io"
 	"os"
 
@@ -69,6 +70,9 @@ func SignWithIntegrityBlock(privKey crypto.PrivateKey) error {
 	if err != nil {
 		return err
 	}
+
+	webBundleId := integrityblock.GetWebBundleId(ed25519privKey)
+	fmt.Println("Web Bundle ID: " + webBundleId)
 
 	signedBundleFile, err := os.Create(*flagOutput)
 	if err != nil {
--- a/go/integrityblock/integrityblock.go
+++ b/go/integrityblock/integrityblock.go
@@ -4,12 +4,14 @@ import (
 	"bytes"
 	"crypto/ed25519"
 	"crypto/sha512"
+	"encoding/base32"
 	"encoding/binary"
 	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
 	"os"
+	"strings"
 
 	"github.com/WICG/webpackage/go/internal/cbor"
 )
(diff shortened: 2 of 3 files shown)


sonkkeli:
I also did [a PR about a very simple CLI tool to run this](https://github.com/sonkkeli/webpackage/commit/16858b446082e0efa487c88f91c3b09345bd5dbb).

I'm not sure if such tool is needed but I feel like printing the web bundle ID after signing the bundle will be "too late" if the web bundle is not using relative URLs, because then the ID would have already be needed for the baseURL of the web bundle. WDYT? Ping also @peletskyi @cmfcmf

peletskyi:
Yes, that sounds like a very useful thing. But I'd add a flag to `sign-bundle` tool that would just print Web Bundle ID instead of building a dedicated executable.

sonkkeli:
@peletskyi that would be too late also, because the web bundle is generated with `gen-bundle` tool before using the `sign-bundle` tool. To me it feels a bit random to pass the private key to the `gen-bundle` tool because it's not otherwise needed by the that tool, but if it's okay for @irori I can do that as well.

peletskyi:
I mean that you can add another meaning to existing flag (say `-privateKey`) or introduce a new flag ( `-i`, it is up to you what to choose) that would just take a private key and no path to Web Bundle:
`sign-bundle -privateKey private-key.pem`
The result would be the Web Bundle ID:
`abcdaaddd....adfadaic`

This command will do the same as `dump-bundle-id -privateKey privkey.pem`

sonkkeli:
Ah gotcha what you meant now.

I think it would be best to do what @cmfcmf suggested and refactor the sign-bundle CLI tool to use sub commands. 

In practise this would mean that instead of the current way:
- `sign-bundle -signType integrityblock -privateKey ./path/key.pem ...all other args` and
- `sign-bundle -signType signaturessection -privateKey ./path/key.pem ...all other args`
- (Following this pattern would be: `sign-bundle -dumpId -privateKey ./path/key.pem`)

The tool would be run like this:
- `sign-bundle integrity-block -privateKey ./path/key.pem ...all other args`
- `sign-bundle signatures-section -privateKey ./path/key.pem ...all other args`
- `sign-bundle dump-id -privateKey ./path/key.pem`



I'm wondering if this would cause any breaking changes to someone using the tool though? Do you @irori know if someone would be affected by this?

peletskyi:
Sounds good

sonkkeli:
I just created a draft [PR](https://github.com/WICG/webpackage/compare/main...sonkkeli:webpackage:sub-commands-for-sign-bundle-golang?expand=1) for it. @irori LMK if I can convert it into an actual one.

Also see the switch case, I now defaulted it to run the signed exchanges one (to prevent any potential breaking changes if those are possible), but this can lead into someone writing non-sense like `sign-bundle helloworld -privateKey ./path/key.pem ...all other args` and it would still run the `signatures-section` one.  So LMK if this defaulting is needed or not. :)

irori:
Using sub commands SGTM.

Don't worry about making breaking changes, AFAIK there are no existing users of the signatures section.

irori (approved):

PR #852, opened 2023-04-06
JS CLI: Add support for using password-encrypted PKCS#8 keys

sonkkeli:


js/sign/src/cli.ts
```diff
@@ -24,10 +26,22 @@ function readOptions() {
...
+
+  return parsePemKey(privateKeyFile, await readPassphrase());
+}
+
+export async function main() {
   const options = readOptions();
   const webBundle = fs.readFileSync(options.input);
-  const parsedPrivateKey = parsePemKey(
+  const parsedPrivateKey = await parseMaybeEncryptedKey(
     fs.readFileSync(options.privateKey, 'utf-8')
```
cmfcmf, line 45: Could we change this to read the key as a Buffer (by removing the encoding parameter)? I should've noticed that earlier; treating the private key as raw bytes is better than trying to utf-8 decode it.

js/sign/src/cli.ts
```diff
@@ -24,10 +26,22 @@ function readOptions() {
     .parse(process.argv);
 }
 
-export function main() {
+async function parseMaybeEncryptedKey(
+  privateKeyFile: string
```
cmfcmf, line 30: Could we change this to `privateKeyBytes: Buffer`?

js/sign/src/integrityblock.ts
```diff
@@ -7,12 +7,34 @@ import {
...
+      rl.close();
+    })
+  );
+}
 
 // A helper function which can be used to parse string formatted keys to
 // KeyObjects.
-export function parsePemKey(unparsedKey: string): KeyObject {
+export function parsePemKey(
+  unparsedKey: string,
```
cmfcmf, line 32: Could we change this to `keyBytes: Buffer`?

js/sign/src/integrityblock.ts
```diff
@@ -7,12 +7,34 @@ import {
...
+// password-decrypted private key.
+export async function readPassphrase(): Promise<string> {
+  const rl = readline.createInterface({
+    input: process.stdin,
+    output: process.stdout,
+  });
+
+  return new Promise((resolve) =>
+    rl.question('Passphrase for the key: ', (passphrase) => {
+      // TODO(sonkkeli): Hide the password when being written.
```
cmfcmf, line 22: How about using https://www.npmjs.com/package/read as part of this PR to hide the password?

js/sign/src/cli.ts
```diff
@@ -24,10 +26,22 @@ function readOptions() {
...
+): Promise<KeyObject> {
+  try {
+    return parsePemKey(privateKeyFile);
+  } catch (e) {
+    console.warn("This might be an encrypted private key, let's try again.");
+  }
+
+  return parsePemKey(privateKeyFile, await readPassphrase());
```
cmfcmf, line 38: Could you use `e.code` in the catch block to check if it is indeed just missing the passphrase, and otherwise re-throw the error?

Pseudocode:
```
try {
  read();
} catch (e) {
  if (e.code !== 'ERR_MISSING_PASSPHRASE') {
    throw e;
  }
  read(await passphrase());
}
```

https://nodejs.org/api/errors.html#err_missing_passphrase

sonkkeli: As discussed offline the e.code is inconsistent between node versions so keeping as is for now

sonkkeli:
PTAL

cmfcmf (approved):
Awesome stuff :)

PR #855, opened 2023-04-12
wbn-sign: Improve readability by splitting bloated files

sonkkeli:
No functional changes, just refactoring. Main change is moving WebBundleId and some helper functions into their own files and moving the IntegrityBlockSigner into a directory called `signers` where I can then later put all the other signers too.


js/sign/src/wbn-sign.ts
```diff
@@ -1,8 +1,6 @@
...
   IntegrityBlock,
-  WebBundleId,
-  parsePemKey,
-  readPassphrase,
-  getRawPublicKey,
-} from './integrityblock.js';
+  IntegrityBlockSigner,
+} from './signers/integrityblock.js';
+export { WebBundleId } from './web-bundle-id.js';
+export * from './utils/utils.js';
```
cmfcmf, line 6: I feel like exporting all utils, even future ones, is a bit bold, since all of them will become part of the public API of this package. I would only export the minimum amount that is necessary, and probably even try to reduce the amount of utils that are exported. 
For now, this means explicitly listing all the things you want to export instead of using `*`.

sonkkeli: We will get rid of the utils exports after the interfaces are introduced so I dont think it's a problem, but will change this to export them by their names.

cmfcmf (approved):

PR #856, opened 2023-04-21
Use public key instead of private key for GetWebBundleId()

sonkkeli:
This will be needed later on when the private key is no longer certain to be available for the tools.


cmfcmf (approved):
LGTM

PR #857, opened 2023-04-21
Add type for SignatureAttributesMap for readability and code consistency

sonkkeli:


go/integrityblock/integrityblock_test.go
```diff
@@ -44,7 +44,7 @@ func TestAddNewSignatureToIntegrityBlock(t *testing.T) {
 		t.Errorf("integrityBlock.CborBytes. err: %v", err)
 	}
 
-	want := `["🖋📦" "1b\x00\x00" [[map["ed25519PublicKey":"publickey"] "signature"]]]`
+	want := `["🖋📦" "1b\x00\x00" [[map["` + Ed25519publicKeyAttributeName + `":"publickey"] "signature"]]]`
```
cmfcmf, line 47: nit: I would argue that in these `want` lines of tests, you actually do want to use the literal name of the attribute, and not the constant. I think I'd want the tests to fail if you accidentally or deliberately change `Ed25519publicKeyAttributeName` in the future.

cmfcmf (approved):
LGTM % nit in tests

go/integrityblock/integrityblock_test.go
sonkkeli: Done thanks!

PR #862, opened 2023-05-02
wbn & wbn-sign: Drop `.js` from imports

sonkkeli:


js/bundle/tsconfig.json
```diff
@@ -1,15 +1,18 @@
...
+    "moduleResolution": "node",
+    "esModuleInterop": true,
+    "strict": true,
+    "declaration": true,
+    "newLine": "lf",
+    "outDir": "lib",
+    "noImplicitReturns": true,
+    "skipLibCheck": true,
+    "allowJs": true,
+    "noEmit": true,
```
cmfcmf, line 14: I don't think we want `noEmit` here, since otherwise the `build:esm` script will no longer create `.js` files.

cmfcmf (approved):
LGTM % tsconfig comments

PR #863, opened 2023-05-03
wbn-sign: Introduce ISigningStrategy interface

sonkkeli:
Also introduce NodeCryptoSigningStrategy which will replace the currently existing way to sign.


js/sign/README.md
```diff
@@ -96,8 +104,11 @@ rm ed25519key.pem
 
 ## Release Notes
 
-### v0.0.2
+### v0.1.0
 
+- BREAKING CHANGE: Introducing the support for using different types of signing
```
cmfcmf, line 109: Is it really a breaking change if using `{ key: privateKey }` is still possible?

cmfcmf: Ah, I guess the fact that `IntegrityBlockSigner::sign` now returns a Promise is a breaking change.

js/sign/src/signers/node-crypto-signering-strategy.ts
```diff
@@ -0,0 +1,24 @@
+import crypto, { KeyObject } from 'crypto';
+import { isValidEd25519Key } from '../utils/utils.js';
+import { ISigningStrategy } from './signing-strategy-interface.js';
+
+// Class to be used when signing with parsed `crypto.KeyObject` private key
+// provided directly as `SignerOptions.key`.
```
cmfcmf, line 6: nit: `SignerOptions.key` is mentioned nowhere else. I would write something like "...provided directly in the contructor".

js/sign/src/utils/utils.ts
```diff
@@ -34,3 +34,20 @@ export function getRawPublicKey(publicKey: crypto.KeyObject) {
     publicKey.export({ type: 'spki', format: 'der' }).slice(-32)
   );
 }
+
+export function isValidEd25519Key(
```
cmfcmf, line 38: nit: Rename to `checkIsValidEd25519Key` and add a comment to the method saying something along the lines of "will throw if the key is not a valid Ed2559 key of the specified type".

js/sign/README.md
cmfcmf: Ah, it looks like "Option 1" above is just outdated?

js/sign/src/signers/integrity-block-signer.ts
```diff
@@ -134,17 +130,11 @@ export class IntegrityBlockSigner {
     return new Uint8Array(buffer);
   }
 
-  signAndVerify(
+  verifySignature(
     dataToBeSigned: Uint8Array,
```
cmfcmf, line 134: nit: Rename to just `data`

js/sign/src/signers/integrity-block-signer.ts
```diff
@@ -16,28 +17,22 @@ type IntegritySignature = {
...
+  private signingStrategy: ISigningStrategy;
 
-  constructor(webBundle: Uint8Array, opts: IntegrityBlockSignerOptions) {
-    if (opts.key.asymmetricKeyType !== 'ed25519') {
-      throw new Error(
-        `IntegrityBlockSigner: Only ed25519 keys are currently supported. Your key's type is ${opts.key.asymmetricKeyType}.`
-      );
-    }
-    this.key = opts.key;
+  constructor(webBundle: Uint8Array, signingStrategy: ISigningStrategy) {
```
cmfcmf, line 24: nit: Use `private readonly` for both parameters and remove the private members explicitly declared above and the assignments in the constructor body.

js/sign/README.md
```diff
@@ -33,10 +33,18 @@ const privateKey = wbnSign.parsePemKey(
...
   key: privateKey,
 }).sign();
 
+// Option 2: To specify the signing strategy. Note that one can also create and
+// pass their own SigningStrategy class implementing ISigningStrategy.
+const { signedWebBundle } = new wbnSign.IntegrityBlockSigner(
+  webBundle,
+  new wbnSign.NodeCryptoSigningStrategy(privateKey)
+).sign();
+
```
cmfcmf, line 47: Could you add an example here with a custom signing strategy? E.g.

```js
const { signedWebBundle } = new wbnSign.IntegrityBlockSigner(
  webBundle,
  class {
    async sign(data: Uint8Array): Promise<Uint8Array> { /* ... */ }
    async getPublicKey(): Promise<KeyObject> { /* ... */ }
  }
).sign();
```

cmfcmf (approved):
LGTM % some documentation and minor naming suggestions

js/sign/src/signers/integrity-block-signer.ts
sonkkeli: I will do this in a separate PR as this naming convention is used in several places over the code and it's not really a change part of this PR.

js/sign/README.md
sonkkeli: You are right, looks like i forgot to add the awaits for both of them :P Fixed now.

PR #864, opened 2023-05-08
Rename dataToBeSigned to be just data when use case is obvious

sonkkeli:


cmfcmf (approved):
Awesome, sorry for the confusion!

PR #873, opened 2023-05-19
go: Split SignWithIntegrityBlock to be usable without flags

sonkkeli:
This is a step towards being able to use the tooling with one's own signing strategy without being forced to use it as a CMD tool with flags. With this change one can create their own wrapper Golang tool with their own ISigningStrategy implementation and pass their input and output files without being forced to use flags.

The tool does not support providing the bytes directly as that would then limit the size of the web bundle into the max size of []byte array.


irori (approved):
LGTM % comments

irori (approved):
LGTM % comments

Issue #875, opened 2023-07-17
Cannot install bundle tool "matched no packages"


Snugug:
When attempting to install the [bundle tool](https://github.com/WICG/webpackage/tree/main/go/bundle) using the provided command `go install github.com/WICG/webpackage/go/bundle/cmd/...@latest`, I get the following error:

```bash
$: go install github.com/WICG/webpackage/go/bundle/cmd/...@latest
go: warning: "github.com/WICG/webpackage/go/bundle/cmd/...@latest" matched no packages
```

`go version` provides the following:

```
$: go version
go version go1.15.15 linux/amd64
```

Snugug:
Yup, it looks like `go install` was added in  1.18. Upgrading to that fixed the problem for me. Maybe the min Go version for that command should be included in the README for people (like me) who aren't familiar with Go but may need to use the bundler

sonkkeli:
I created a [PR](https://github.com/WICG/webpackage/pull/878) now to mention the min version so closing this issue. If someone wants to use older Golang, then they shall figure out the right commands themselves.

sonkkeli:
I created a https://github.com/WICG/webpackage/pull/878 now to mention the min version so closed this issue.

PR #878, opened 2023-07-18
Update mentions of go install with the required Golang version

sonkkeli:
To address concerns raised here: https://github.com/WICG/webpackage/issues/875


cmfcmf (approved):
LGTM, thanks!

PR #888, opened 2024-04-24
ECDSA P-256 SHA-256 implementation

GrapeGreen:
This PR implements support for the ECDSA P-256 SHA-256 signing scheme for wbn-sign.


js/sign/src/signers/node-crypto-signing-strategy.ts
```diff
@@ -1,12 +1,12 @@
...
 
 // Class to be used when signing with parsed `crypto.KeyObject` private key
 // provided directly in the constructor.
 export class NodeCryptoSigningStrategy implements ISigningStrategy {
   constructor(private readonly privateKey: KeyObject) {
-    checkIsValidEd25519Key('private', privateKey);
+    checkIsValidKey('private', privateKey);
   }
 
   async sign(data: Uint8Array): Promise<Uint8Array> {
```
GrapeGreen, line 12: The main question here -- in your opinion, does it make sense to state the algorithm explicitly for clarity in `.sign()` and `.verify()` or omit it for simplicity?

[crypto](https://nodejs.org/api/crypto.html#cryptosignalgorithm-data-key-callback) documentation states that missing algorithm will be inferred based on the key type; I've validated that assumption and it turns out that for ECDSA P-256 keys the hashing algorithm chosen is indeed SHA-256.

js/sign/package.json
```diff
@@ -43,15 +44,15 @@
...
+    "@types/node": "^17.0.0",
     "esbuild": "^0.14.47",
     "jasmine": "^4.2.1",
     "mock-stdin": "^1.0.0",
     "prettier": "2.8.0",
     "typescript": "^4.7.3"
   },
   "engines": {
-    "node": ">= 14.0.0",
+    "node": ">= 17.0.0",
```
sonkkeli, line 55: Was there a reason why the upgrade of the node version was needed? The bundler plugins are also assuming >=14 node version so increasing it should be made in parallel with the other versions to make sure nothing breaks.

Also if necessary to change, the readme and release notes should also mention this.

js/sign/src/signers/node-crypto-signing-strategy.ts
sonkkeli: I don't have a strong opinion here. Maybe if kept like this, could add a function comment on where it's inferred from

js/sign/src/utils/constants.ts
```diff
@@ -1,4 +1,6 @@
 export const ED25519_PK_SIGNATURE_ATTRIBUTE_NAME = 'ed25519PublicKey';
+export const ECDSA_P256_SHA256_PK_SIGNATURE_ATTRIBUTE_NAME =
```
sonkkeli, line 2: Would it make sense to combine these into a map and to be used with the SignatureType enum values as keys?

js/sign/src/utils/utils.ts
```diff
@@ -31,15 +36,63 @@ export function parsePemKey(
...
+  }
+  return SignatureType.EcdsaP256SHA256;
+}
+
+export function getPublicKeyAttributeName(key: crypto.KeyObject) {
+  switch (getSignatureType(key)) {
+    case SignatureType.Ed25519:
+      return ED25519_PK_SIGNATURE_ATTRIBUTE_NAME;
+    case SignatureType.EcdsaP256SHA256:
+      return ECDSA_P256_SHA256_PK_SIGNATURE_ATTRIBUTE_NAME;
```
sonkkeli, line 68: With the enum+map combo i believe you could even do something like [this](https://stackoverflow.com/a/39372911), but maybe it's not so cool with just 2 supported key types. Are we expecting to have more key types supported?

js/sign/src/web-bundle-id.ts
```diff
@@ -1,33 +1,49 @@
...
+  SignatureType,
+} from './utils/utils.js';
 
 // Web Bundle ID is a base32-encoded (without padding) ed25519 public key
 // transformed to lowercase. More information:
 // https://github.com/WICG/isolated-web-apps/blob/main/Scheme.md#signed-web-bundle-ids
 export class WebBundleId {
   // https://github.com/WICG/isolated-web-apps/blob/main/Scheme.md#suffix
-  private readonly appIdSuffix = [0x00, 0x01, 0x02];
+  private readonly kEd25519PublicKeyTypeSuffix = [0x00, 0x01, 0x02];
```
sonkkeli, line 15: Same as the other consts

sonkkeli (commented):
Some tests are failing still on the github workflow, could you fix those as well

js/sign/src/signers/node-crypto-signing-strategy.ts
cmfcmf: I don't have a strong opinion but would be fine to keep it as is.

js/sign/src/utils/utils.ts
```diff
@@ -31,15 +36,63 @@ export function parsePemKey(
...
 
-export function getRawPublicKey(publicKey: crypto.KeyObject) {
-  // Currently this is the only way for us to get the raw 32 bytes of the public key.
-  return new Uint8Array(
-    publicKey.export({ type: 'spki', format: 'der' }).slice(-32)
+export function isAsymmetricKeyTypeSupported(key: crypto.KeyObject) {
+  return (
+    key.asymmetricKeyType === 'ed25519' ||
+    (key.asymmetricKeyType === 'ec' &&
+      key.asymmetricKeyDetails!.namedCurve === 'secp256k1')
```
cmfcmf, line 43: nit (same behavior, but without the risk to ever attempt to access `namedCurve` on `undefined`): 
```suggestion
      key.asymmetricKeyDetails?.namedCurve === 'secp256k1')
```

js/sign/src/utils/utils.ts
```diff
@@ -31,15 +36,63 @@ export function parsePemKey(
...
+
+export function getSignatureType(key: crypto.KeyObject) {
+  assert(
+    isAsymmetricKeyTypeSupported(key),
+    'Expected either "Ed25519" or "ECDSA P-256" key.'
   );
+  if (key.asymmetricKeyType === 'ed25519') {
+    return SignatureType.Ed25519;
+  }
+  return SignatureType.EcdsaP256SHA256;
```
cmfcmf, line 60: Using a `switch` with a `default` case would feel a bit safer here, IMHO.

js/sign/tests/integrity-block-signer_test.js
```diff
@@ -43,16 +68,27 @@ describe('Integrity Block Signer', () => {
...
+      );
+      expect(() =>
+        initSignerWithTestWebBundleAndKeys(keypairValid.privateKey)
+      ).not.toThrowError();
+    }
 
     for (const invalidKey of [
       { keyType: 'rsa', options: { modulusLength: 2048 } },
       { keyType: 'dsa', options: { modulusLength: 1024, divisorLength: 224 } },
-      { keyType: 'ec', options: { namedCurve: 'secp256k1' } },
```
cmfcmf, line 55: nit: Maybe keep this but change the `namedCurve` to `sect239k1`?

js/sign/tests/integrity-block-signer_test.js
```diff
@@ -91,63 +127,92 @@ describe('Integrity Block Signer', () => {
...
-      new wbnSign.IntegrityBlock().toCBOR(),
-      cborg.encode({
-        [constants.ED25519_PK_SIGNATURE_ATTRIBUTE_NAME]: rawPubKey,
-      })
-    );
+  [
+    crypto.generateKeyPairSync('ed25519'),
+    crypto.generateKeyPairSync('ec', { namedCurve: 'secp256k1' }),
+  ].forEach((keypair) => {
+    fit(`generates the dataToBeSigned correctly with ${createTestSuffix(
```
cmfcmf, line 134: Remove `f`

js/sign/package.json
GrapeGreen: [KeyObject.asymmetricKeyDetails](https://nodejs.org/api/crypto.html#keyobjectasymmetrickeydetails) requires at least `15.7.0`. I've figured that it makes little sense to bump the version to something that old (since the current version is `18.19.0`), so I've ended up picking `17.0.0` as something in the middle. If you're fine with this one, I'll go ahead and update the docs too.

js/sign/src/utils/utils.ts
GrapeGreen: A switch on what exactly? 🤔 

- `SignatureType.Ed25519 <=> key.asymmetricKeyType === 'ed25519'`
- `SignatureType.EcdsaP256SHA256 <=> key.asymmetricKeyType === 'ec' &&
      key.asymmetricKeyDetails?.namedCurve === 'secp256k1'`

js/sign/package.json
sonkkeli: Yeah fine for me as long as you upgrade the version also in the other repo :)

js/sign/src/signers/node-crypto-signing-strategy.ts
GrapeGreen: Added a comment then.

js/sign/tests/integrity-block-signer_test.js
GrapeGreen: Done!

js/sign/src/utils/constants.ts
GrapeGreen: Done!

js/sign/src/utils/utils.ts
GrapeGreen: Done!

js/sign/src/web-bundle-id.ts
GrapeGreen: Done!

js/sign/src/utils/utils.ts
GrapeGreen: Done!

js/sign/tests/integrity-block-signer_test.js
GrapeGreen: Done!

js/sign/package.json
GrapeGreen: Done!

js/sign/src/utils/utils.ts
sonkkeli: I would think @cmfcmf means something like this

`
switch 
`

sonkkeli:
LGTM but please wait for @cmfcmf's lgtm too :) 

(Sidenote: the currently failing builds are not your fault and they always fail so you can ignore those)

sonkkeli (approved):
LGTM but please wait for @cmfcmf's lgtm too :)

(Sidenote: the currently failing builds are not your fault and they always fail so you can ignore those)

js/sign/tests/integrity-block-signer_test.js
```diff
@@ -91,63 +128,94 @@ describe('Integrity Block Signer', () => {
...
-      cborg.encode(sigAttr)
-    );
+  [
+    crypto.generateKeyPairSync('ed25519'),
+    crypto.generateKeyPairSync('ec', { namedCurve: 'secp256k1' }),
+  ].forEach((keypair) => {
+    it(`generates a valid signature with ${createTestSuffix(
+      keypair.publicKey
+    )}.`, async () => {
+      const keypair = crypto.generateKeyPairSync('ed25519');
```
cmfcmf, line 186: typo

js/sign/src/utils/utils.ts
cmfcmf: I thought of something like

```js
switch (key.asymmetricKeyType) {
  case 'ed25519':
    return SignatureType.Ed25519;
  case 'ec':
    if (key.asymmetricKeyDetails?.namedCurve === 'secp256k1') {
      return SignatureType.EcdsaP256SHA256;
    }
    assert(false);
  default:
    assert(false);
```

I think ideally you'd refactor the functions to this:

```
function maybeGetSignatureType(key: crypto.KeyObject): SignatureType | null;
export function getSignatureType(key: crypto.KeyObject): SignatureType {
  const s = maybeGetSignatureType(key);
  assert(s);
  return s;
}
export function isAsymmetricKeyTypeSupported(key: crypto.KeyObject): bool {
  return maybeGetSignatureType(key) !== null;
}
```

cmfcmf (approved):
LGTM % nits

js/sign/src/utils/utils.ts
GrapeGreen: Done!

js/sign/tests/integrity-block-signer_test.js
GrapeGreen: Done!

js/sign/README.md
sonkkeli, line 1: since you pumbed the version, please update the release notes in the end of this file

cmfcmf (approved):
LGTM

PR #890, opened 2024-07-08
Introduce integrity block v2

GrapeGreen:
This PR changes the format of the integrity block from `[magic, version, signatures]` to `[magic, version, ib_attributes, signatures]`, where `ib_attributes` is a map with a single attribute called `webBundleId`; this field's purpose is to simplify key rotation while having a guarantee that the ID of the web bundle is always stable regardless of the signing keys specified.

The default behavior of the signing tool stays the same, i.e. the following invocation
`wbn-sign --input {bundle} -k {key} -o {output}`
will continue generating the usual `v1` bundles.

With `v2`, on the contrary, it's now possible to have the bundle signed by multiple signatures in parallel and explicitly call out the exact `web-bundle-id`:
`wbn-sign --version v2 --input {bundle} -k {key1} {key2} {key3} -o {output} --web-bundle-id {id}`

Chrome supports both `v1` & `v2` bundles as of M128.


js/sign/src/signers/integrity-block-signer.ts
```diff
@@ -18,41 +22,55 @@ type IntegritySignature = {
 };
 
 export class IntegrityBlockSigner {
+  // `webBundleId` is ignored if `is_v2` is false.
   constructor(
+    private readonly is_v2: boolean,
```
sonkkeli, line 27: Would it make sense to instead use an enum + switches and store the version in case more versions will be added in the future? Then the boolean wouldnt work anymore.

js/sign/src/signers/integrity-block-signer.ts
```diff
@@ -18,41 +22,55 @@ type IntegritySignature = {
...
-      this.calcWebBundleHash(),
-      ibCbor,
-      attrCbor
-    );
+      const signature = await signingStrategy.sign(dataToBeSigned);
+      this.verifySignature(dataToBeSigned, signature, publicKey);
 
-    const signature = await this.signingStrategy.sign(dataToBeSigned);
-    this.verifySignature(dataToBeSigned, signature, publicKey);
+      // The signatures are calculated independently, and thus not appended to the integrity block here to not affect subsequent calculations.
```
sonkkeli, line 64: nit: line break

js/sign/src/signers/integrity-block-signer.ts
GrapeGreen: Not really -- my intention is to eliminate everything related to v1 in a month :)

sonkkeli (approved):

PR #894, opened 2024-08-12
Add example way to implement ISigningStrategy using external KMS

zgroza:
also, ran linter on this file


zgroza:
Example created as a separate repository: https://github.com/chromeos/wbn-sign-gcp-kms

PR #913, opened 2026-03-17
Add/remove/replace/info commands added to wbn-sign CLI

rferens:
Summary
  This change refactors the core signing logic into a new high-level SignedWebBundle API, introduces support for multiple signatures

  Key Changes

   * New SignedWebBundle API: Introduced a central class in src/core/signed-web-bundle.ts that provides a simplified, high-level interface for creating, loading, and modifying signed web bundles. It supports:
       * Signing an unsigned bundle with one or more strategies.
       * Adding or removing signatures from an existing signed bundle.
       * Retrieving the Web Bundle ID (App ID).
       * Loading signed bundles directly from bytes.
   * Enhanced Integrity Block Management:
       * Moved IntegrityBlock to a dedicated core module and added the ability to parse existing integrity blocks from CBOR.
       * Refactored IntegrityBlockSigner to be internal, deprecating several methods in favor of the new SignedWebBundle and utility functions.

   * Infrastructure & Utilities:
       * Added new utility functions for identifying pure vs. signed bundles and verifying signatures.
       * Updated tsconfig.json to strip internal declarations from the generated types.

PR #914, opened 2026-03-21
New cli commands

rferens:
This PR expands the CLI with new commands for managing integrity blocks.
  * CLI Improvements:
       * Updated the wbn-sign CLI with a new command-based interface (sign, add-signature, remove-signature, replace-signature, and info), see README for more info.
       * Added colored console output for better readability using the colors package.

This is a continuation of https://github.com/WICG/webpackage/pull/913, which was split into API(original PR) and CLI(this PR) parts.

### DIFF
--- a/js/sign/README.md
+++ b/js/sign/README.md
@@ -124,80 +124,47 @@ const webBundleIdWithIWAOrigin = new wbnSign.WebBundleId(
 
 ## CLI
 
-This package also includes 2 CLI tools
+This package also includes 2 CLI tools:
 
-- `wbn-sign` which lets you sign a web bundle easily without having to write any
-  additional JavaScript.
-- `wbn-dump-id` which can be used to calculate the Web Bundle ID corresponding
-  to your signing key.
+- `wbn-sign`: A comprehensive tool for signing bundles and managing signatures.
+- `wbn-dump-id`: A simple utility to calculate the Web Bundle ID for a given key.
 
-### Running wbn-sign 
-#### New usage form (>= 0.2.7)
+### Running wbn-sign
 
-The base usage from is: `wbn-sign [command] [options] <arguments...>`
+The base usage is: `wbn-sign [command] [options] <arguments...>`
 
-Currenly supported commands are:
-```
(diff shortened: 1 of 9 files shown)


js/sign/README.md
```diff
@@ -218,31 +185,45 @@ The environment variable set like this, can then be used in other scripts, for
...
 
 ```
 openssl genpkey -algorithm Ed25519 -out ed25519key.pem
 ```
 
-For better security, one should prefer using passphrase-encrypted ed25519
-private keys. To encrypt an unencrypted private key, run:
+**Note**: Although both are key types are secure, 
+we recommend using ed25519 as it is cosidered slightly more secure than ecdsa-p256.
+The ecdsa algorithm, in opposite to deterimistic ed25519, is relying on strength of pseudo-random generator.
```
zgroza: Nit: I think it would explain the point better like this

```suggestion
**Note**: We recommend using Ed25519, as it is considered more secure than ECDSA P-256. Unlike ECDSA, which relies on a pseudo-random number generator and is vulnerable to entropy-related flaws, Ed25519 is deterministic and remains secure even if the system's random number generator is compromised.
```

js/sign/package.json
```diff
@@ -44,6 +44,7 @@
   "dependencies": {
     "base32-encode": "^2.0.0",
     "cborg": "^4.2.14",
+    "colors": "^1.4.0",
```
zgroza: This package is historically... problematic: https://www.sonatype.com/blog/npm-libraries-colors-and-faker-sabotaged-in-protest-by-their-maintainer-what-to-do-now

Can You maybe revert to the custom approach used before?

js/sign/README.md
rferens: Thanks, I really like your formulation. It is applied.

js/sign/package.json
rferens: O! I didn't know this. Thanks for catching that! I'm reverting to the previous approach

zgroza (approved):
LGTM

